PR #54, opened 2013-02-17
use package.json package.json

oroce:
Use package.json instead of bundled modules.

It's useful for paas deployments, and for every native modules (like bson, mongodb, etc).

I've had a conversation with @osoner and @gorkem-cetin.
We were talking about updating geoip-lite's country and city data. But as I can see, it is bundled with the module and there's a pull request for auto-updating it: bluesmoon/node-geoip#22

Anyway I created an updater script: https://gist.github.com/oroce/4972816


ar2rsawseen:
Hello, thank you for submitting pull request. In new core that was developed in paralel (since version 15.xx now) all dependencies are installed upon counlty installation for the top folder.

Thank you for collaboration

oroce:
Haha wow this was a long time ago

PR #86, opened 2014-01-30
Change absolute urls to relative

izdravkov:
So now, countly can be used in reverse proxy configurations as:
http(s)://server_address/countly


ar2rsawseen:
Hello, using your idea we made version 15.xx relative url compatible, unfortunately that was done on a separate version (in the new plugins core), thus could not merge it.

thank you for submitting pull request

PR #215, opened 2015-12-09
Allow zero value in the badge parameter

ianko:
We should be able to send a badge equals zero to clear the iOS notification center and badge count.

The if statement was bypassing the parameter, since it was testing a falsey value.


iartem:
Thanks for your contribution. You missed a place to add a fix, so I made a fix in my own commit: https://github.com/Countly/countly-server/commit/b718d5f675414f86ec96dc9e93b3abb6d22429ee

PR #241, opened 2016-03-09
Does not work in Fedora 22

igorklopov:
Seems that Fedora nodesource rpm has no "nodejs>node" symlink


igorklopov:
Sorry, found it: https://github.com/Countly/countly-server/blob/master/bin/countly.install_rhel.sh#L23
However seems that everything works with `node` except my fixed file.

ar2rsawseen:
I thought that default was nodejs and node was a legacy call.
So is the fix still valid after your finding of symlink?

igorklopov:
I think we should better apply the fix. At least to follow single consideration - to use only 'node' calls in this file :)

PR #242, opened 2016-03-09
Does not work in sudo-less instances

igorklopov:
Just one more minor tweak i found.
If it is not important, just discard :)


gorkem-cetin:
Well sudo is a critical system component and without it, many of the apps out there may not work properly. Therefore it's expected that sudo is always present in even a barebone Linux instance.

ar2rsawseen:
If countly.install.sh is itself supposed to run with root permission, does it need more sudo statements inside it?

PR #266, opened 2016-06-15
add a endpoint "/ping" for health checking

diegomarangoni:
Add a endpoint for usage within a load balancer to check api availability.


ar2rsawseen:
Hey, thank you for the great idea.
I will also modify it to respond default Succes text for consistency with other responses and also check connection to database in those ping requests

PR #273, opened 2016-07-24
China map support by echarts

jianglinchun:
Modified countly.city.js & countly.location.js for map support by echarts.
countly.city.js has variables _geoCoord which list all cities of china has not been compelte yet. And this variables should be localized. City name for echarts(2.7) options must be english. So _countryEnMap variables was added and could be optimized.

`CountlyHelpers.loadJS("province/javascripts/baidu/echarts-all.js", drawByECharts);`
The path of echarts-all.js should be replace to the right place.

Please contact me jianglinchun@sohu.com if you have any problems.


ar2rsawseen:
@jianglinchun quick question, frontend/express/public/javascripts/api.js seems to be compatibility lib for google api, right? but if we disable google api and use echarts only, then we don't need it right?

Also is it possible to display only China map using echarts?

jianglinchun:
@ar2rsawseen The answer is yes as I had metioned in wechat.
But I'm wondering if you switch all of you geo-graphics to echarts may take a lot of time. And I had never tested all the geo-graphics for other countries by echarts.

gorkem-cetin:
Since this PR conflicts with underlying core files, it should be converted to a plugin in order to be fully usable in Countly. Closing now & feel free to reopen in the future once it is converted to a plugin.

PR #316, opened 2017-07-19
Added apply button and changed visuals for push plugin date-time-picker.

melihkorkmaz:
Added apply button and changed visuals for push plugin date-time-picker.

### DIFF
--- a/plugins/push/frontend/public/javascripts/component.datepicker.js
+++ b/plugins/push/frontend/public/javascripts/component.datepicker.js
@@ -68,7 +68,7 @@ window.component('datepicker', function(datepicker) {
 			m('.picker', [
 				m('.comp-datepicker-ui-picker', {config: datepicker.config(ctrl)}),
 				m('.comp-datepicker-time', [
-					m('span.comp-datepicker-time-label', 'Pick Time: '),
+					m('span.comp-datepicker-time-label', t('datepicker.pick-time') + ': '),
 					m('input[type=number][min=0][max=23]', {value: ctrl.hours(), oninput: m.withAttr('value', ctrl.setHours)}),
 					m('span.comp-datepicker-time-spacer', ':'),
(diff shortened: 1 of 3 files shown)


ar2rsawseen:
Want @iartem to check it, when you are available :)

plugins/push/frontend/public/stylesheets/main.css
```diff
@@ -987,7 +1015,7 @@ table.d-table .nothing { opacity: 0.3; }
 
 .comp-push input[type="number"]::-webkit-inner-spin-button,
 .comp-push input[type="number"]::-webkit-outer-spin-button {
-    -webkit-appearance:none;
```
iartem: This is rather wrong, we have number inputs not only in date picker

melihkorkmaz: Ok, I searched for other input-number related to .com-push class and I didn't find. 
But you're right. I can add a specific class for those inputs.

plugins/push/frontend/public/javascripts/component.datepicker.js
```diff
@@ -82,17 +68,22 @@ window.component('datepicker', function(datepicker) {
 			m('.picker', [
 				m('.comp-datepicker-ui-picker', {config: datepicker.config(ctrl)}),
 				m('.comp-datepicker-time', [
-					m('span.comp-datepicker-time-label', 'Time'),
```
iartem: Let's localise this

plugins/push/frontend/public/stylesheets/main.css
melihkorkmaz: Ok to be honest, I couldn't find a solution to return back input-number to browser defaults. I asked on stack-overflow, but nobody has answered it yet. https://stackoverflow.com/questions/45230816/browser-default-value-for-input-number-arrows

melihkorkmaz: did you have time to check functionality? I couldn't find an answer for the problem right above. So I can cancel to adding a spinner to input.

PR #373, opened 2017-12-18
Times of days plugin

melihkorkmaz:
Times of day plugin for session and events visualization.

### DIFF
--- a/plugins/times-of-day/api/api.js
+++ b/plugins/times-of-day/api/api.js
@@ -36,7 +36,7 @@ var plugin = {},
 		var update = {};
 		var options = {};
 
-		if (hasSession && params.qstring.hour && params.qstring.dow) {
+		if (hasSession && params.qstring.hour !== undefined && params.qstring.dow !== undefined) {
 			criteria = {
 				"_id": "tod_" + appId + ":[CLY]_session"
 			};
@@ -73,7 +73,10 @@ var plugin = {},
 
 				var dow = events[i].dow || params.qstring.dow;
 				var hour = events[i].hour || params.qstring.hour;
-
+				
+				if(dow === undefined || hour === undefined)
+					continue;
+				
 				var incData = {};
 				incData[dow + "." + hour + ".count"] = 1;
 				var setData = {};
--- a/plugins/times-of-day/api/api.js
+++ b/plugins/times-of-day/api/api.js
@@ -1,0 +1,187 @@
+var plugin = {},
+	common = require('../../../api/utils/common.js'),
+	plugins = require('../../pluginManager.js'),
+	async = require('async');
+
+(function (plugin) {
+	plugins.register("/i", function (ob) {
+		var params = ob.params;
+
+		if (!params.qstring.begin_session && !params.qstring.events) {
+			return false;
+		}
+
+
+		var appId = params.app_id;
+		var hour = params.qstring.hour;
+		var dow = params.qstring.dow;
+		var events = [];
+
+		if (!appId ||
+			!(parseInt(hour) >= 0 && parseInt(hour) <= 23) ||
+			!(parseInt(dow) >= 0 && parseInt(dow) <= 6)) {
+			return false;
+		}
+
+		var hasSession = false;
+		var hasEvents = false;
+
+		if (params.qstring.begin_session) {
+			hasSession = true;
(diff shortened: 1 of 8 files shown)


plugins/times-of-day/api/api.js
```diff
@@ -0,0 +1,187 @@
...
+
+		var queryArray = [];
+		var query = {};
+		var criteria = {};
+		var update = {};
+		var options = {};
+
+		if (hasSession) {
+			criteria = {
+				"_id": "tod_" + appId + ":Sessions"
```
ar2rsawseen: Can we use [CLY]_session here for consistency with other places where session is used?

plugins/times-of-day/api/api.js
```diff
@@ -0,0 +1,187 @@
...
+			};
+
+			var incData = {};
+			incData[dow + "." + hour + ".count"] = 1;
+			var setData = {};
+			setData["_id"] = "tod_" + appId + ":Sessions";
```
ar2rsawseen: Same here

plugins/times-of-day/api/api.js
```diff
@@ -0,0 +1,187 @@
...
+		}
+
+		if (hasEvents) {
+			for (var i = 0; i < events.length; i++) {
+				criteria = {
+					"_id": "tod_" + appId + ":" + events[i].key
+				};
+
+				var incData = {};
+				incData[dow + "." + hour + ".count"] = 1;
```
ar2rsawseen: Actually each event can have their own dow and hour params, and we should use them and fallback to request level dow and hour only if they are not available in the event level

plugins/times-of-day/api/api.js
```diff
@@ -0,0 +1,187 @@
...
+		});
+
+		async.parallel(parallelTasksArray, function (err, result) {
+			if (err) {
+				console.log("Error while updating times of day: ", err.message);
+			}
+		})
+
+		function insertUpdateTimesOfDay(query, callback) {
+			common.db.collection('timesofday').update(query.criteria, query.update, query.options, function (err, result) {
```
ar2rsawseen: May be as we are constantly updating documents in the loop, it would be better to use bulk processing. In this case all queries are sent to db server and are run there and then only result is returned, consuming only 1 connection, instead of opening new one on each update

```
common.db.onOpened(function(){
    var bulk = common.db._native.collection("timesofday").initializeUnorderedBulkOp();
    for(var i = 0; i < queue.length; i++){
        bulk.find(queue[i].criteria).upsert().updateOne(queue[i].update);
    }
    bulk.execute(function(err, updateResult) {
        if(err){
            //there was an error
        }
        //all done
    });
});
```

plugins/times-of-day/frontend/public/javascripts/countly.models.js
```diff
@@ -0,0 +1,446 @@
...
+            url: countlyCommon.API_URL + "/o",
+            data: {
+                "api_key": countlyGlobal.member.api_key,
+                "app_id": countlyCommon.ACTIVE_APP_ID,
+                "method": "get_events"
+            },
+            success: function (json) {
+                _eventsList = json || {};
+            }
+        });
```
ar2rsawseen: Can we reuse the method from countlyEvents model for this?

plugins/times-of-day/frontend/public/javascripts/countly.views.js
```diff
@@ -0,0 +1,126 @@
...
+$(document).ready(function () {
+    var menu = '<a href="#/analytics/times-of-day" class="item" ">' +
+        '<div class="logo fa fa-plugin" style="background-image:none; font-size:24px; text-align:center; width:35px; margin-left:14px; line-height:42px;"></div>' +
+        '<div class="text" data-localize="times-of-day.plugin-title"></div>' +
+        '</a>';
+
+    $('#default-type #analytics-submenu').append(menu);
+    $('#mobile-type #analytics-submenu').append(menu);
+    $('#web-type #analytics-submenu').append(menu);
+    $('#desktop-type #analytics-submenu').append(menu);
```
ar2rsawseen: Assuming this is relevant for all app types, can we add it to all sub trees, or else if we add new app type, it won't be there automatically

PR #375, opened 2017-12-18
Better wording for configurations. Title search fixed for config page.

melihkorkmaz:


plugins/plugins/frontend/public/localization/plugins.properties
```diff
@@ -82,25 +82,25 @@ configs.logs.warn = Warning Level
 configs.logs.error = Error Level
 configs.logs.default = Default Level
 
-configs.help.frontend-production = If enabled, serves minified, combined files
+configs.help.frontend-production = If enabled, serves minified, combined files. (what is the cause of disabling or enabling this? Any potential harms?)
```
ar2rsawseen: I am sure, this is a comment, rather than a change :)

gorkem-cetin: Right, that was my comment for you in fact.. There is a need for a better explanation if you think it helps?

melihkorkmaz: :) Sorry for that. I'll be more carefull. I just copied it :(

ar2rsawseen: Initial load of dashboard should be faster, due to smaller files and smaller file amount, but when developing a plugin, you need to regenerate them to see changes

ar2rsawseen: No probs, thats what reviews are for :)

PR #387, opened 2018-01-05
Some issues have been fixed.

melihkorkmaz:


melihkorkmaz:
- Added common.initTimeObj. (line: 39 and 76)
- Used property "a" instead of "app_id". (line 52 and 100)
- QueryArray changed to Query object to set multiple update at once. Added test for it.
- Used "drop" rather than remove for (cleardata, reset app etc.)
- app_key removed from model.js
- Testing order has been changed to empty data check > add data > session validate > event validate > reset app > empty data check.

PR #401, opened 2018-01-29
[frontend] add Webfont lib for font loading FOUT issue

waiterZen:


frontend/express/views/dashboard.html
```diff
@@ -71,7 +71,8 @@
         <script language="javascript" type="text/javascript" src="<%- cdn %>javascripts/utils/jquery.hoverIntent.minified.js"></script>
         <script language="javascript" type="text/javascript" src="<%- cdn %>javascripts/utils/tooltipster/tooltipster.bundle.min.js"></script>
         <script language="javascript" type="text/javascript" src="<%- cdn %>javascripts/utils/highlight/highlight.pack.js"></script>
-        <script language="javascript" type="text/javascript" src="<%- cdn %>javascripts/utils/dropzone.js"></script>
+		<script language="javascript" type="text/javascript" src="<%- cdn %>javascripts/utils/dropzone.js"></script>
+		<script language="javascript" type="text/javascript" src="<%- cdn %>javascripts/utils/webfont.js"></script>
```
ar2rsawseen: You should also add new webfont.js to Gruntfile.js so it would be added when generating production files:
https://github.com/Countly/countly-server/blob/master/Gruntfile.js#L71

waiterZen: fixed

PR #403, opened 2018-01-31
fix percent of platformVersion

vagromghost:
fix percent of platformVersion issue


frontend/express/public/javascripts/countly/countly.helpers.js
```diff
@@ -1690,10 +1690,12 @@
                     }
                     if(shouldDelete)
                         delete oSVersionData.chartData[i];
+                        delete platformVersionTotal[i];
```
ar2rsawseen: Since `shouldDelete` is not enclosed in {} brackets, it only includes `delete oSVersionData.chartData[i];` and thus `delete platformVersionTotal[i];` is executed for every segment deleting them

PR #422, opened 2018-02-07
[views]add params

waiterZen:


plugins/views/api/api.js
```diff
@@ -397,8 +397,12 @@ var plugin = {},
         return new Promise(function(resolve, reject){
             var params = ob.params;
             if (params.qstring.events && params.qstring.events.length && Array.isArray(params.qstring.events)) {
+                if(!params.views){
+                    params.views = [];
+                }
                 params.qstring.events = params.qstring.events.filter(function(currEvent){
                     if (currEvent.key == "[CLY]_view"){
+                        params.views.push(currEvent);
```
ar2rsawseen: I think this should be put on line 415 where we check if it has visits, so you won't get duplicate views that only update durations

PR #434, opened 2018-02-20
docker building

thatsk:
Building The docker container we can use squash option
docker build --squash -t countly/countly-server


iartem:
Hmm.. So you took packaging branch and want us to merge it into master in this pull request? Or I miss something?

PR #474, opened 2018-04-04
Task manager schema update

waiterZen:


api/utils/taskmanager.js
```diff
@@ -150,6 +156,8 @@ var request = require("request");
         update.view = options.view || "";
         update.request = JSON.stringify(options.request || {});
         update.app_id = options.app_id || "";
+        update.creator = options.creator;
+        update.global = options.global;
```
ar2rsawseen: Maybe we need default values for these, so existing tasks, that do not have or provide this value should be global or private?
I think we need default true on global property, and also a constraint, if creator is not provided, then global is also set as true, or else, no one will be able to access it.

Additionally, it changes the logic of checking if task is already running, so for example, if drill query is running for too long and switches to long task. Then if same user or another user tries the same query, we check if such long task already exists and if yes, we do not perform that task again, but let user wait on the same shared result.

In this changes with global and private, when checking if tasks exists, we also need to check if user has access to it, so if task is global or user is creator, only then we confirm that tasks exists.

PR #509, opened 2018-05-28
Added metric changes setting to configs.

melihkorkmaz:


ar2rsawseen:
I think it would also be beneficial not read metric_changes in case they are disabled:
https://github.com/Countly/countly-server/blob/53cb3e1cb73237f4279e4b9b1b4e52800d6f898e/api/parts/data/fetch.js#L735-L862

PR #519, opened 2018-06-07
injectable widgets, headless browser for dashboard report, injectable email reports

prikshittekta:


api/parts/mgmt/ip.js
```diff
@@ -0,0 +1,47 @@
+var ip = {},
```
ar2rsawseen: Since it is a new utility module, please comment it in JS Doc format, so others can learn about it and know how to use it. Here is an example:
https://github.com/Countly/countly-server/blob/master/api/utils/authorizer.js

api/utils/render.js
```diff
@@ -0,0 +1,97 @@
+var puppeteer = require('puppeteer');
```
ar2rsawseen: Same here, need JSDoc comments. You can then try generating documentation using `countly docs generate` and view them under yourdomain.com/docs/

frontend/express/app.js
```diff
@@ -1431,6 +1434,62 @@ app.post(countlyConfig.path+'/graphnotes/delete', function (req, res, next) {
...
+    var pathUrl = req.url.replace(countlyConfig.path, "");
+    var urlParts = url.parse(pathUrl, true);
+    var fullPath = urlParts.pathname;
+
+    authorize.verify_return({db:countlyDb, token:token, req_path:fullPath, callback:function(valid){
+        if(!valid){
+            return res.redirect(countlyConfig.path+'/login?message=login.result');
+        }
+
+        req.session.uid = valid;
```
ar2rsawseen: Maybe we should assign all session parameters here, like this: https://github.com/Countly/countly-server/blob/master/frontend/express/app.js#L970-L973
As in, what if some view is only rendered for Global admins, and we dont have global admin flag in this session

frontend/express/app.js
```diff
@@ -1431,6 +1434,62 @@ app.post(countlyConfig.path+'/graphnotes/delete', function (req, res, next) {
...
+});
+
+app.get(countlyConfig.path+'/login/token/:token', function(req, res){
+    var token = req.params.token;
+    var pathUrl = req.url.replace(countlyConfig.path, "");
+    var urlParts = url.parse(pathUrl, true);
+    var fullPath = urlParts.pathname;
+
+    authorize.verify_return({db:countlyDb, token:token, req_path:fullPath, callback:function(valid){
+        if(!valid){
```
ar2rsawseen: Since it is form of authorization, it is a security concern and should be logged to System logs plugin. So basically this place should dispatch events `tokenLoginSuccess` and `tokenLoginFailed` like this (just use other name, not apiKey, but tokenLogin, etc):
https://github.com/Countly/countly-server/blob/master/frontend/express/app.js#L1042

And systemlogs plugin should listen to it and record attempts, like this:
https://github.com/Countly/countly-server/blob/master/plugins/systemlogs/frontend/app.js#L59-L65

And of course it needs to be localized too:
https://github.com/Countly/countly-server/blob/master/plugins/systemlogs/frontend/public/localization/systemlogs.properties#L33-L34

frontend/express/public/javascripts/countly/countly.template.js
```diff
@@ -2646,6 +2648,24 @@ var AppRouter = Backbone.Router.extend({
...
+    },
+    addReportsCallbacks: function(widget, options){
+        if(!this.reportCallbacks[widget]){
+            this.reportCallbacks[widget] = {};
+        }
+        Object.assign(this.reportCallbacks[widget], options);
+    },
+    getReportsCallbacks: function(){
+        return this.reportCallbacks;
+    },
```
ar2rsawseen: I actually thought these methods would either be added by plugin or be in plugin's view as static methods. I don't think it makes sense to add them to core. So the easiest way to fix this if plugins themselves add them to `app` object. And then other plugins that use it, firstly need to check if method exists `if(app.addReportsCallbacks)` then add callback. etc

plugins/reports/api/api.js
```diff
@@ -337,6 +380,15 @@ var logpath = path.resolve(__dirname, '../../../log/countly-api.log');
...
     }
+
+    function validateUserForEmails(params){
+        if (!params.member.global_admin){
+            common.returnMessage(params, 401, 'User does not have right to access this information');
+            return false;
+        }else{
+            return true;
+        }
+    }
```
ar2rsawseen: Can you explain this validation? So only users with global admin access are able to create none core email reports?

plugins/reports/frontend/public/javascripts/countly.views.js
```diff
@@ -767,6 +767,22 @@ window.ReportingView = countlyView.extend({
...
+app.addReportsCallbacks = function(plugin, options){
+    if(!this.reportCallbacks){
+        this.reportCallbacks = {};
+    }
+
+    if(!this.reportCallbacks[plugin]){
+        this.reportCallbacks[plugin] = {};
+    }
+
+    Object.assign(this.reportCallbacks[plugin], options);
```
ar2rsawseen: Unfortunately Object.assign has no IE support at all, we want to keep compatible with at least latest IE versions too

frontend/express/public/javascripts/countly/countly.template.js
```diff
@@ -2657,15 +2656,6 @@ var AppRouter = Backbone.Router.extend({
     getWidgetCallbacks: function(){
```
ar2rsawseen: Can we extract Widgets for dahsboard plugin too in the same way?

frontend/express/app.js
```diff
@@ -1431,6 +1434,75 @@ app.post(countlyConfig.path+'/graphnotes/delete', function (req, res, next) {
...
+            return res.redirect(countlyConfig.path+'/login?message=login.result');
+        }
+
+        countlyDb.collection('members').findOne({"_id":countlyDb.ObjectID(valid)}, function (err, member) {
+            if(err || !member){
+                plugins.callMethod("tokenLoginFailed", {req:req, res:res, data: {token: token, token_owner: valid}});
+                return res.redirect(countlyConfig.path+'/login?message=login.result');
+            }
+
+            req.session.uid = member["_id"];
```
ar2rsawseen: Also due to it being authentication, we should regenerate sesison id to prevent session hijacking, like this:
https://github.com/Countly/countly-server/blob/master/frontend/express/app.js#L968

PR #521, opened 2018-06-13
[SERVER-853] Countly Version check for upgrades

Cookiezaurs:


bin/scripts/checking_versions.js
```diff
@@ -0,0 +1,87 @@
...
+{
+    console.log("could not load current version. Using: "+versions[versions.length-1]);
+    current_version = versions[versions.length-1];
+}
+
+if(marked_version=="")
+{
+    console.log("could not load marked version. Using: "+versions[0]);
+    marked_version = versions[0];
+}
```
ar2rsawseen: If one of these checks fail, it is better whole script to fail rather than make assumptions of versions

bin/scripts/checking_versions.js
```diff
@@ -0,0 +1,87 @@
...
+
+if(current_version==marked_version)
+{
+    console.log("up to date");
```
ar2rsawseen: As this script will be used by bash script which will get and process response, it is better to leave console output limited only to the result array. Maybe with some additional debug flag, it could output more console logs

bin/scripts/checking_versions.js
```diff
@@ -0,0 +1,87 @@
...
+{
+    while(versions[from]<=marked_version && from<versions.length){ from++;}
+    while(versions[til]>current_version && til>=0){ til--;}
+
+    if(til==-1 || from ==versions.length)
+        console.log("version range not found");
+    else
+        {
+        versions = versions.slice(from,til);
+        console.log(versions.join(","));
```
ar2rsawseen: output should be in a way that bash script could easily interpret as an array of values and iterate through result. 
or empty if there was an error

PR #522, opened 2018-06-13
server side table export update

waiterZen:


frontend/express/public/javascripts/countly/countly.template.js
```diff
@@ -2282,9 +2282,9 @@ var AppRouter = Backbone.Router.extend({
                         });
                         exportDrop.on("open",function(){ 
                             if(exportAPIData) {
-                                $(".server-export .countly-drop-content").empty().append(CountlyHelpers.export(oSettings._iRecordsDisplay, exportAPIData, null, true).removeClass("dialog"));
+                                $(".server-export .countly-drop-content").empty().append(CountlyHelpers.export(oSettings._iRecordsDisplay, app[exportView].getExportAPI(oSettings.sTableId), null, true).removeClass("dialog"));
                             } else if(exportQueryData) {
-                                $(".server-export .countly-drop-content").empty().append(CountlyHelpers.export(oSettings._iRecordsDisplay, exportQueryData).removeClass("dialog"));
+                                $(".server-export .countly-drop-content").empty().append(CountlyHelpers.export(oSettings._iRecordsDisplay, app[exportView].getExportQuery(oSettings.sTableId)).removeClass("dialog"));
```
ar2rsawseen: What would be the difference between these two? Either stored in variable or direct method call?

plugins/compliance-hub/frontend/public/javascripts/countly.views.js
```diff
@@ -13,6 +13,55 @@ window.ConsentManagementView = countlyView.extend({
...
+            var apiQueryData = {
+                api_key: countlyGlobal.member.api_key,
+                app_id: countlyCommon.ACTIVE_APP_ID,
+                path: requestPath,
+                method: "GET",
+                filename:"Compliance_export_or_purge_history_on_" + moment().format("DD-MMM-YYYY"),
+                prop: ['aaData']
+            };
+            return apiQueryData;
+        }
```
ar2rsawseen: This table was actually added by systemlogs, would make sense that export was handled by systemlogs as well

frontend/express/public/javascripts/countly/countly.template.js
waiterZen: Because this function triggered by user click download icon to show download menu.
before that time, user may change the filter condition for talble, So we had better reload the newest query string for download.

plugins/compliance-hub/frontend/public/javascripts/countly.views.js
waiterZen: But the table defined the same file here, so I think we should keep the download query related function with table definition. How do you think?

ar2rsawseen: No, table is defined in systemlogs plugin:
https://github.com/Countly/countly-server/blob/master/plugins/systemlogs/frontend/public/javascripts/countly.views.js#L349

PR #529, opened 2018-06-19
Web feedback review changes

frknbasaran:
- device_id_change, app_user_delete, app_user_data_export cases hooks
- app_life-cycle actions, on_create, on_delete, clear_specific_period, clear_all_data, app_reset hooks
- feedback collection uid, ts indexes
- external google font files moved into core
- app_id requirements changed to app_key on api
- console outputs removed
- localizations completed
- api validations
- arrow functions removed from client-side


plugins/star-rating/api/api.js
```diff
@@ -10,30 +10,373 @@ var plugin = {},
...
+                                 *  the original event format like:
+                                 *  { key: '[CLY]_star_rating', count:1, sum:1, segmentation:{ platform:"iOS", version:"3.2", rating:2}
+                                 *  this function will add a field call "platform_version_rate" in segmentation.
+                                 */
+                                currEvent.segmentation['platform_version_rate'] = currEvent.segmentation['platform'] + "**" + currEvent.segmentation['app_version'] + "**" + currEvent.segmentation['rating'] + "**";
+                                // remove data from core data
+                                // for GDPR compatibility
+                            });
+                    });
+                    return true;
```
ar2rsawseen: So you only return true if event key **[CLY]_star_rating**
Meaning all other events are filtered out and never make it to be processed :D

plugins/star-rating/api/api.js
```diff
@@ -10,30 +10,373 @@ var plugin = {},
...
+                            if (plugins.getConfig("api").safe) {
+                                common.returnMessage(params, 400, 'App does not exist');
+                            }
+                            return false;
+                        }
+                        
+                        var collectionName = 'feedback' + app["_id"];
+                        
+                        common.db.collection(collectionName)
+                            .insert({
```
ar2rsawseen: Here we insert feedback without even checking if feedback was provided at all? I mean if fields like comment is not even provided or is empty, do we need to insert it at all?

plugins/star-rating/api/api.js
```diff
@@ -10,30 +10,373 @@ var plugin = {},
...
+            query["widget_id"] = params.qstring.widget_id;
+        }
+        if (params.qstring.rating) {
+            query["rating"] = params.qstring.rating;
+        }
+        if (params.qstring.device_id) {
+            query["device_id"] = params.qstring.device_id;
+        }
+
+        common.db.collection(collectionName)
```
ar2rsawseen: Here we need to validate if user has right to read for this app

plugins/star-rating/api/api.js
```diff
@@ -10,30 +10,373 @@ var plugin = {},
...
+    /*
+    * @apiName: GetWidgetsData
+    * @apiDescription: Get feedback widgets with or without filters
+    * @apiParam: 'app_id', app_id of related application
+    * @apiParam: 'is_active', is_active option for widgets
+    */
+    plugins.register('/o/web-feedback/widgets',  function(ob) {
+        var params = ob.params;
+        common.db.collection('apps')
+            .findOne({'key': params.qstring.app_key}, (err, app) => {
```
ar2rsawseen: Here we should use app_id and validation for user read too

plugins/star-rating/api/api.js
```diff
@@ -10,30 +10,373 @@ var plugin = {},
...
+                        }
+                    })
+            })
+        return true;
+    });
+
+    /*
+    * @apiName: GetOneWidget
+    */
+    plugins.register('/o/web-feedback/widget', function(ob) {
```
ar2rsawseen: I assume this one is for fetching form for SDK, so we use app_key and no validation, correct?
So it means only person, who has widget_id can fetch the data.

plugins/star-rating/api/api.js
```diff
@@ -110,14 +452,87 @@ var plugin = {},
...
             );
             return true;
         }
         return false;
     });
+    
+    plugins.register("/i/apps/create", function(ob){
+        var params = ob.params;
+        var appId = ob.appId;
+        common.db.collection('web_feedback_widgets_' + appId).insert({_id:"meta"},function(){});
```
ar2rsawseen: We have meta document for widgets? What does it hold?

plugins/star-rating/api/api.js
```diff
@@ -110,14 +452,87 @@ var plugin = {},
...
+        common.db.collection('feedback' + appId).ensureIndex({"ts":1, "uid":1}, {unique:true}, function(){});
```
ar2rsawseen: Here I think we need separate indexes and not compound, cause we are either fetch by ts or by uid (whem merging, deleting, etc) not by both.
Also definitely not needed unique constraint

plugins/star-rating/api/api.js
```diff
@@ -110,14 +452,87 @@ var plugin = {},
...
+    });
+
+    plugins.register("/i/apps/delete", function(ob){
+        var appId = ob.appId;
+        common.db.collection('web_feedback_widgets_' + appId).drop(function() {});
+        common.db.collection('feedback' + appId).drop(function() {});
+        if(common.drillDb)
+            common.drillDb.collection("drill_events" + crypto.createHash('sha1').update("[CLY]_star_rating" + appId).digest('hex')).drop(function() {});
```
ar2rsawseen: We delete drill event, but may be we should also delete core event from countly.events{hash} collection?
Or is it deleted somewhere else?

plugins/star-rating/api/api.js
```diff
@@ -110,14 +452,87 @@ var plugin = {},
...
+    });
+    
+    plugins.register("/i/apps/clear", function(ob){
+        var appId = ob.appId;
+        var ids = ob.ids;
+        common.db.collection('web_feedback_widgets_'+appId).remove({$and:[{'_id': {$regex: appId + ".*"}}, {'_id': {$nin:ids}}]},function(){});
```
ar2rsawseen: This completele does not look correct. For period deletion, we should not delete widgets at all.

plugins/star-rating/api/api.js
```diff
@@ -110,14 +452,87 @@ var plugin = {},
...
+        common.db.collection('feedback' + appId).remove({ts:{$lt:ob.moment.unix()}}, function() {});
+        if(common.drillDb)
+            common.drillDb.collection("drill_events" + crypto.createHash('sha1').update("[CLY]_star_rating" + appId).digest('hex')).remove({ts:{$lt:ob.moment.valueOf()}}, function() {});
+    });
+
+    plugins.register("/i/apps/clear_all", function(ob){
+        var appId = ob.appId;
+        common.db.collection('web_feedback_widgets_' + appId).drop(function() {});
```
ar2rsawseen: No need to delete widgets on clear, only on reset

plugins/star-rating/frontend/public/javascripts/countly.models.js
```diff
@@ -87,6 +89,162 @@
...
+        })
+    }
+
+    starRatingPlugin.createFeedbackWidget = function(feedbackWidget, callback) {
+        return $.ajax({
+            type: "GET",
+            url: countlyCommon.API_URL + "/i/web-feedback/widgets/create",
+            data: {
+                api_key: countlyGlobal['member'].api_key,
+                app_key: countlyGlobal['apps'][store.get("countly_active_app")].key,
```
ar2rsawseen: no need to pass app_key when not needed, let's not expose it that much, same in all other places

plugins/star-rating/frontend/public/javascripts/countly.views.js
```diff
@@ -337,18 +399,16 @@ window.starView = countlyView.extend({
...
-
         }
+        // there is no localization for these strings for now
+        var rating_strings = ["Very dissatisfied","Somewhat dissatisfied","Neither satisfied nor dissatisfied","Somewhat satisfied","Very satisfied"];
         var typeName = '<a style="font-size: 1px; display:none;">' + starName + '</a>';
         if (times && times > 0) {
-            while (times--) {
-                result += '<i class="fa fa-star star-icon" aria-hidden="true"></i>';
-            }
+            result += '<span><img class="little-feedback-icon" src="/star-rating/images/star-rating/'+(times-1)+'_color.svg"></span><span class="star-rating-icon-title">' + rating_strings[times-1] + '</span>';
```
ar2rsawseen: Let's not use absolute paths

plugins/star-rating/frontend/public/javascripts/countly.views.js
```diff
@@ -502,18 +562,370 @@ window.starView = countlyView.extend({
...
+                CountlyHelpers.refreshTable(self.widgetTable, json);    
+            })
+        } else {
+            var columnsDefine = [
+                {
+                    "mData": function (row) {
+                        return row.popup_header_text;   
+                    } ,
+                    sType: "string",
+                    "sTitle": "NAME"
```
ar2rsawseen: columns should be localized, try reusing what we have in core dashboard.properties and if not there, then only add new localization strings

plugins/star-rating/frontend/public/javascripts/countly.views.js
```diff
@@ -502,18 +562,370 @@ window.starView = countlyView.extend({
...
+                },
+                {
+                    "sTitle": "STATUS",
+                    "mData": function (row) {
+                        if (!(countlyGlobal.member.admin_of && (countlyGlobal.member.admin_of.indexOf(countlyCommon.ACTIVE_APP_ID) !== -1)) && !(countlyGlobal.member.global_admin)) {
+                            return (row.is_active) ? 'Active' : 'Deactive';
```
ar2rsawseen: Localize

plugins/star-rating/frontend/public/javascripts/countly.views.js
```diff
@@ -502,18 +562,370 @@ window.starView = countlyView.extend({
...
+                        } else {
+                            var input = '<div class="on-off-switch">';
+                            if (row.is_active == 'true') {
+                                input += '<input type="checkbox" id="widget-status-'+row._id+'"" class="on-off-switch-checkbox" checked>';
+                            } else {
+                                input += '<input type="checkbox" id="widget-status-'+row._id+'"" class="on-off-switch-checkbox">';
+                            }
+                            input += '<label class="on-off-switch-label  widget-edit-status" data-id="'+ row._id + '" for="plugin-' + row._id + '"></label>';
+                            input += '<span class="text">' + 'Enable' + '</span>';
```
ar2rsawseen: Localize

plugins/star-rating/frontend/public/stylesheets/countly-feedback.css
```diff
@@ -0,0 +1,525 @@
...
+         box-sizing: border-box;
+         width: 90%;
+         margin-left: 5%;
+         padding-bottom: 25px;
+         padding-top: 25px;
+         border-radius: 5px;
+         display: none;
+    }
+     #question-area {
+        font-family: 'Lato', sans-serif;
```
ar2rsawseen: I see font files being add and fonts being user, but I don't see any font-face definitions for the font files, like we have here, as example:
https://github.com/Countly/countly-server/blob/master/frontend/express/public/stylesheets/main.css#L13

plugins/star-rating/frontend/public/stylesheets/countly-feedback.css
```diff
@@ -0,0 +1,525 @@
...
+}
+ .dbviewer-filter-status {
+     float: right;
+     display: none;
+     font-size: 13px;
+     margin-top: 5px;
+     margin-right: 5px;
+}
+/* On mouse-over, add a grey background color */
+ .countly-feedback-checkbox-container:hover input ~ .dbviewer-custom-checkbox-checkmark {
```
ar2rsawseen: What does it have to do with dbviewer?

api/lib/countly.common.js
```diff
@@ -1551,6 +1551,26 @@ var countlyCommon = {},
...
         return countlyCommon.periodObj;
     }
+
+    /**
+    * Async for each function
+    * @param {array} array - array to loop
+    * @param {function} fn - callback function
+    * @param {function} atEnd - optional function to will work at the end
+    */
+    countlyCommon.asyncForeach = function(array, fn, atEnd) {
```
ar2rsawseen: Why won't you just use async.js?
http://caolan.github.io/async/docs.html#each

frontend/express/public/javascripts/countly/countly.common.js
```diff
@@ -3351,4 +3351,24 @@
         return tmpAvgVal.toFixed(2);
     }
 
+    /**
+    * Async for each function
+    * @param {array} array - array to loop
+    * @param {function} fn - callback function
+    * @param {function} atEnd - optional function to will work at the end
+    */
+    countlyCommon.asyncForeach = function(array, fn, atEnd) {
```
ar2rsawseen: And same function in 2 places?  use async.js
Like this:
https://github.com/Countly/countly-server/blob/9ffae2ac5a9afea4707f586fc65d659cb4ce7d31/bin/upgrade/18.01/scripts/delete_drill_meta.js#L24

frontend/express/public/javascripts/countly/countly.helpers.js
```diff
@@ -37,6 +37,44 @@
...
+     /**
+    * Display modal popup that requires confirmation input from user and optional checkbox
+    * @param {string} msg - message to display in alert popup
+    * @param {string} type - type of alert red for errors and green for success
+    * @param {boolean} hasCheckbox - popup has checkbox? or not.
+    * @param {string} checkboxTitle - title of checkbox element 
+    * @param {function} callback - to determine result of the input
+    * @param {array=} buttonText - [0] element for cancle button text and [1] element for confirm button text
+    * @example
+    * CountlyHelpers.confirm("Are you sure?", "red", function (result) {
```
ar2rsawseen: Name in example is still different :)

plugins/star-rating/api/api.js
```diff
@@ -110,14 +496,90 @@ var plugin = {},
...
+        common.db.collection("events"+crypto.createHash('sha1').update("[CLY]_star_rating" + appId).digest('hex')).drop(function() {});
+        if(common.drillDb)
+            common.drillDb.collection("drill_events" + crypto.createHash('sha1').update("[CLY]_star_rating" + appId).digest('hex')).drop(function() {});
+
+    });
+    
+    plugins.register("/i/apps/clear", function(ob){
+        var appId = ob.appId;
+        var ids = ob.ids;
+        common.db.collection('web_feedback_widgets_'+appId).remove({ts:{$lt:ob.moment.unix()}}, function() {});
```
ar2rsawseen: do we even have ts in that collection? As I understood it stores the feedback forms and not feedback itself, so it should only be deleted/cleared on app delete and app reset

PR #542, opened 2018-07-05
[frontend](view): javascript bugfix

onokumus:
- unreachable code after return statement


ar2rsawseen:
I think this was on purpose, return null to remove such input from the configuration form for now, as it is not used, but remain the code, in case we want to bring it back later

PR #548, opened 2018-07-07
[feedback] integration instructions popup changes

frknbasaran:
- some localizations updated
- collection name prefixes updated as feedback instead of web-feedback
- api endpoints names updated as feedback instead of web-feedback
- widget instructions modal added
- widget row option menu updated
- feedback widget css moved to server-side as web-feedback-web.css
- feedback-popup.html updated
-- used jQuery ajax method instead of web-sdk request method
-- used device_id of web sdk value instead of device_id value of dashboard


plugins/star-rating/api/api.js
```diff
@@ -177,12 +178,12 @@ var plugin = {},
                                 "email": currEvent.segmentation.email,
                                 "comment": currEvent.segmentation.comment,
                                 "ts":currEvent.timestamp,
-                                "device_id":params.qstring.device_id,
+                                "device_id":currEvent.device_id,
```
ar2rsawseen: The way I see in popup dialog is that device_id is sent through `params.qstring.device_id` and not currEvent

plugins/star-rating/frontend/public/templates/feedback-popup.html
```diff
@@ -1,291 +1,408 @@
...
+                cly.type = 'text/javascript';
+                cly.async = true;
+                // Enter url of script here (see below for other option)
+                cly.src = '/javascripts/min/countly.min.js';
+                cly.onload = function() {
+                    Countly.init()
+                };
+                var s = document.getElementsByTagName('script')[0];
+                s.parentNode.insertBefore(cly, s);
+            })();
```
ar2rsawseen: I think we need to exclude Web SDK from here completely, it will not track what we want it to track eitherway.

plugins/star-rating/frontend/public/templates/feedback-popup.html
```diff
@@ -1,291 +1,408 @@
...
+                                    app_version: 1,
+                                    platform_version_rate: "",
+                                    rating: eventObject.segmentation.rating,
+                                    email: eventObject.segmentation.email,
+                                    comment: eventObject.segmentation.comment
+                                },
+                                timestamp: Date.now(),
+                                hour: h,
+                                dow: dow
+                            }])) + '&app_key=' + Countly.app_key + '&device_id=' + device_id + 'sdk_name=javascript_native_web&sdk_version=' + sdk_version + '&timestamp=' + Date.now() + '&hour=' + h + '&dow=' + dow,
```
ar2rsawseen: Same here. Of course question arises why we have two of the calls, etc, which leads to a bad code structuring. But ok :D

plugins/star-rating/api/api.js
```diff
@@ -26,13 +26,13 @@ var plugin = {},
             var triggerBgColor = params.qstring.trigger_bg_color;
             var triggerFontColor = params.qstring.trigger_font_color;
             var triggerButtonText = params.qstring.trigger_button_text;
-            var targetDevices = params.qstring.target_devices;
+            var targetDevices = JSON.parse(params.qstring.target_devices);
             var targetPage = params.qstring.target_page;
-            var targetPages = params.qstring.target_pages || [];
+            var targetPages = JSON.parse(params.qstring.target_pages) || [];
```
ar2rsawseen: JSON.parse should always we in try catch blocks

plugins/star-rating/frontend/public/javascripts/countly.views.js
```diff
@@ -611,360 +619,492 @@ window.starView = countlyView.extend({
...
-                        return target_pages.trim();
-                    },
-                    sType: "string",
-                    "sTitle": jQuery.i18n.map["feedback.target-pages"]
+                sType: "string",
+                "sTitle": jQuery.i18n.map["common.location"]
+            }, {
+                "mData": function(row) {
+                    var target_pages = "";
+                    if (typeof row.target_pages == "string") row.target_pages = eval(row.target_pages);
```
ar2rsawseen: eval is a very bad function

plugins/star-rating/frontend/public/javascripts/countly.views.js
```diff
@@ -611,360 +619,492 @@ window.starView = countlyView.extend({
...
+                    var vals = Object.values(td);
+                    var atLeastOneDevice = false;
+                    var deviceText = '';
+
+                    keys.forEach(function(key) {
+                        if(td[key]) {
+                            deviceText += key + ' ';  
+                            atLeastOneDevice = true;
+                        } 
+                    })
```
ar2rsawseen: This can be replace by this (without using unsupported functions as `Object.keys` and `Object.values`)

```
for(var key in td){
    if(td[key]) {
        deviceText += key + ' ';  
        atLeastOneDevice = true;
    } 
}
```

plugins/star-rating/frontend/public/javascripts/countly.views.js
```diff
@@ -611,360 +619,492 @@ window.starView = countlyView.extend({
...
+            Array.prototype.remove = function() {
+                var what, a = arguments, L = a.length, ax;
+                while (L && this.length) {
+                    what = a[--L];
+                    while ((ax = this.indexOf(what)) !== -1) {
+                        this.splice(ax, 1);
+                    }
+                }
+                return this;
+            };
```
ar2rsawseen: Think we don't need this anymore? no?

plugins/star-rating/frontend/public/javascripts/countly.views.js
```diff
@@ -611,360 +619,492 @@ window.starView = countlyView.extend({
...
-
+            $('body').off("click", ".options-item .show-instructions").on("click", ".options-item .show-instructions", function() {
+                $('#overlay').fadeIn();
+                $('#widgets-array').html($(this).data('id'));
+                $('.feedback-copy-code').attr("data-clipboard-text", "Countly.q.push(['enable_feedback',{'widgets':['" + $(this).data('id') + "']}]);");
+                $('.feedback-modal').css({
+                    "display": "block"
+                });
+                var id = $(this).data('id');
+                Object.values($('.edit-menu')).splice(0, Object.values($('.edit-menu')).length - 4).forEach(function(menu) {
```
ar2rsawseen: Again, no `Object.values` please, use loops

plugins/star-rating/frontend/public/localization/star-rating.properties
```diff
@@ -21,7 +21,7 @@ star.number-of-ratings=NUMBER OF RATINGS
 star.percentage=PERCENTAGE
 feedback.popup-header-text = What's your opinion about this page?
 feedback.popup-comment-callout = Add comment
-feedback.popup-email-callaout = Contact me with e-mail
+feedback.popup-email-callout = Contact me with e-mail
```
ar2rsawseen: I see this fixed, but don't see fixes of it where it was used, just making sure that usage of the string was also fixed.

PR #560, opened 2018-07-17
[UI][SERVER-759] UI - Trend indicator logic : red/green (Positive & N…

can-keklik:
…egative vs. Increase & Decrease)

### DIFF
--- a/plugins/compliance-hub/frontend/public/templates/compliance.html
+++ b/plugins/compliance-hub/frontend/public/templates/compliance.html
@@ -52,7 +52,7 @@
                             <div class="select">{{this.title}}</div>
                             <div><div style='display:inline-block;position:relative;'>
                                 <div class="number">{{getFormattedNumber this.total}}</div>
-                                <div class="trend {{this.trend}}">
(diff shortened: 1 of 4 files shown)


plugins/compliance-hub/frontend/public/templates/compliance.html
```diff
@@ -86,7 +86,7 @@
                 <div id="dashboard-purge-graph">
                     <div class="data">
                         <div class="name">{{purges.title}}</div>
-                        <div class="trend {{purges.trend}}">
+                        <div class="trend {{purges.trend}} inverse-trend">
```
ar2rsawseen: Why in some places we have additionalStyle in others hardcoded inverse-trend?

can-keklik: Ones with "additionalStyle" are parts in #each loops. One item of the array needs to be inverse, and the other doesn't. So, instead of using an if statement -which is the nearly the same in terms of information which will be passed-, I preferred the class to be dynamic.

PR #561, opened 2018-07-18
[db-viewer] switch between apps with dropdown

frknbasaran:


plugins/dbviewer/api/api.js
```diff
@@ -244,21 +262,30 @@ var plugin = {},
                 }
 			}
 			else{
-                
                 if (params.member.global_admin) {
-                    common.db.collection('apps').find({}).toArray(function (err, apps) {
+                    var query = params.qstring.app ? { "_id":common.db.ObjectID(params.qstring.app) } : {};
```
ar2rsawseen: I thought it was params.qstring.app_id and not params.qstring.app

plugins/dbviewer/api/api.js
```diff
@@ -120,14 +120,20 @@ var plugin = {},
...
                 
-                async.map(apps, getEvents, function (err, events) {
+                // if app_id provided, update app list
+                if (params.qstring.app_id) {
+                    var filteredApps = apps.filter((app) => app._id+"" === params.qstring.app_id+"");    
+                } else {
+                    var filteredApps = apps;
+                }
+                
+                async.map(filteredApps, getEvents, function (err, events) {
```
ar2rsawseen: I don't know if we need a filter here, I think everywhere where we call this method, **apps** array has been already prefiltered accordingly, at least it should be. But if it helps you sleep at night, sure :)

plugins/dbviewer/frontend/public/javascripts/countly.models.js
```diff
@@ -6,20 +6,29 @@
 		_document = {};
 
     //Public Methods
-    countlyDBviewer.initialize = function () {
+    countlyDBviewer.initialize = function (app_id) {
+    	var data = { api_key: countlyGlobal['member'].api_key };
+		// is app_id provided?
+		if (app_id !== "" && app_id == "all") delete data.app_id;
```
ar2rsawseen: we have not yet set app_id to data, why would we delete it?

plugins/dbviewer/frontend/public/javascripts/countly.models.js
```diff
@@ -6,20 +6,29 @@
...
 			type:"GET",
 			url:countlyCommon.API_URL + "/o/db",
-			data:{
-                api_key:countlyGlobal['member'].api_key
-            },
+			data:data,
 			success:function (json) {
 				_data = json;
-                for(var i = 0; i < _data.length; i++){
+				for(var i = 0; i < _data.length; i++){
```
ar2rsawseen: you need to stop using tabs btw, and switch to 4 spaces instead

plugins/dbviewer/frontend/public/javascripts/countly.models.js
```diff
@@ -6,20 +6,29 @@
...
     //Public Methods
-    countlyDBviewer.initialize = function () {
+    countlyDBviewer.initialize = function (app_id) {
+    	var data = { api_key: countlyGlobal['member'].api_key };
+		// is app_id provided?
+		if (app_id !== "" && app_id == "all") delete data.app_id;
+		else data.app_id = app_id;
+		// is there stored app_id?
+		if ((!app_id && store.get('selected_app')) && store.get('selected_app') !== "all") {
+			data.app_id = store.get('selected_app');
```
ar2rsawseen: selected_app storage key is too generic, as storage is shared, it might clash with others

plugins/dbviewer/frontend/public/javascripts/countly.models.js
```diff
@@ -6,20 +6,29 @@
...
+				for(var i = 0; i < _data.length; i++){
                     if(_data[i].collections){
-                        var list = [];
+                    	var list = [];
                         for(var j in _data[i].collections){
-                            list.push(j);
+                        	if (store.get('selected_app_name')) {
+                        		j = j.replace(store.get('selected_app_name'),"").replace(":","").replace("( ","(").toLowerCase().trim().replace("()","");
+							}
+							list.push(j);
```
ar2rsawseen: Why do we need something like this?
I thought the point was that server would already return us all needed collections with names

plugins/dbviewer/frontend/public/javascripts/countly.views.js
```diff
@@ -464,7 +513,7 @@ app.route('/manage/db/:dbs/:collection', 'dbs', function (db, collection) {
 	this.renderWhenReady(this.dbviewerView);
 });
 
-app.route('/manage/db/:dbs/:collection/*document', 'dbs', function (db, collection, document) {
+app.route('/manage/db/:dbs/:collection/:document', 'dbs', function (db, collection, document) {
```
ar2rsawseen: We need this as splash parameter, to allow document _ids containing /, so it would not break url.
Please recheck changes before you are commiting to ensure these are your own changes and you are not overwriting changes of others

plugins/dbviewer/frontend/public/localization/dbviewer.properties
```diff
@@ -19,4 +19,5 @@ dbviewer.descending = Descending
 dbviewer.countly-database = Countly Database
 dbviewer.countly-database-description = Common aggregated anonymous statistics on users and their segments.
 dbviewer.countly-drill-database = Countly Drill Database
-dbviewer.countly-drill-database-description = More granular data and data relations for users and their actions.
\ No newline at end of file
+dbviewer.countly-drill-database-description = More granular data and data relations for users and their actions.
+dbviewer.all-apps = All apps
```
ar2rsawseen: I think it would be enough to reuse https://github.com/Countly/countly-server/blob/master/frontend/express/public/localization/dashboard/dashboard.properties#L35
no need to create another translation string

plugins/dbviewer/frontend/public/templates/dbviewer.html
```diff
@@ -1,8 +1,24 @@
...
-			<div class="title">{{page-title}}</div>
-		</div>
+            <div style="overflow: auto;">
+                <div class="title small">{{page-title}}</div>
+            </div>
+            <div style="width:300px; display: inline-block;">
+                <div class="cly-select float">
+                    <div class="select-inner">
+                        <div class="text-container">
+                            <div class="text" id="app-selector" data-localize="dbviewer.all-apps"></div>
```
ar2rsawseen: Having locazitation hardcoded is dropd down selector is a bad idea, then value will be reset each time page is localized again

plugins/dbviewer/frontend/public/javascripts/countly.models.js
```diff
@@ -6,22 +6,31 @@
 		_document = {};
 
     //Public Methods
-    countlyDBviewer.initialize = function () {
+    countlyDBviewer.initialize = function (app_id) {
+    	var data = { api_key: countlyGlobal['member'].api_key, app_id:app_id };
+		// is app_id provided?
+		if (data.app_id && data.app_id == "all") delete data.app_id;
```
ar2rsawseen: Still confused about this
1) deleting is costly operation, adding something just to delete it seems bad in general
2) So if app_id is not provided, we have data.app_id = undefined.  I know it won't do anything bad, but the logic breaks me, sorry. :D 

can we just do 

```
var data = { api_key: countlyGlobal['member'].api_key};
if(data.app_id && data.app_id !== "all"){
    data.app_id = app_id;
}
```

plugins/dbviewer/frontend/public/javascripts/countly.views.js
```diff
@@ -32,12 +33,78 @@ window.DBViewerView = countlyView.extend({
...
+			} 
+			// clear app-list
+			$('#app-list').html("");
+			// prepend "all apps" link to list
+			$('#app-list').prepend('<div data-value="all" class="app-option item" data-localize=""><span class="app-title-in-dropdown">'+$.i18n.map["common.all"]+'</span></div>');
+			// append list items
+			for (var key in countlyGlobal.apps) {
+				if (store.get('dbviewer_selected_app') && (countlyGlobal.apps[key]._id+"" === store.get('dbviewer_selected_app'))) {
+					app_name = countlyGlobal.apps[key].name; 
+					store.set('dbviewer_selected_app_name',countlyGlobal.apps[key].name);
```
ar2rsawseen: Let's not store app name, it may change, be edited, etc.
We store app_id and that is enough, we can always get current up to date app name from
`countlyGlobal.apps[app_id].name`

PR #591, opened 2018-08-07
[feedback] some css conflicts fixed & new query params for "/feedback" url

frknbasaran:
- some little style conflicts resolved.
- app_version & sdk_name parameters supported by /feedback url now.


plugins/star-rating/frontend/public/templates/feedback-popup.html
```diff
@@ -162,6 +162,8 @@
                 var device_id = getParameterByName('device_id');
                 var sdk_version = getParameterByName('sdk_version');
```
ar2rsawseen: Can we have fallback for SDK version too, because here from debug perspective I am interested more in the feedback plugin version, rather than SDK version.

PR #592, opened 2018-08-07
Fixed username update bug if user rewrite the current username.

melihkorkmaz:


plugins/plugins/frontend/public/javascripts/countly.views.js
```diff
@@ -450,6 +450,11 @@ window.ConfigurationsView = countlyView.extend({
                     return false;
                 }
 
+                if($(this).val() === countlyGlobal["member"].username){
```
ar2rsawseen: Can you check this case

Assuming you have a dashboard with 2 users, test1 and test2

Currently logged in user is test1.
You should not be able to change it to test2.
You open dev console and paste in

`countlyGlobal["member"].username = "test2";`

then you enter value test2 in the input and it is allowed and form would be submitted. 
Would server side validation deal with such case?

melihkorkmaz: I tried it and I couldn't change username because it gives me a warning as the username already exists. There is a server side check at this line.

https://github.com/Countly/countly-server/blob/next/frontend/express/app.js#L1282

PR #605, opened 2018-08-14
Disable ability to enter negative numbers in Configuration parameters

melihkorkmaz:


ar2rsawseen:
I do wonder what will happen if some config would require negative values, but for now we have none, so merging this.

PR #614, opened 2018-08-16
[feedback] #SERVER-869

frknbasaran:
- Rating tests completed.
- Some API result strings changed.


plugins/star-rating/tests.js
```diff
@@ -7,55 +7,255 @@ var API_KEY_ADMIN = "";
...
+            .expect(200)
+            .end(function(err, res) {
+            	if (err) return done(err);
+            	var ob = JSON.parse(res.text);
+            	ob.should.have.property('result','Success');
+            	setTimeout(done, 100);
+            });
+        });
+    });
+
```
ar2rsawseen: Here it would be great to fetch widgets one more time, to ensure changes were stored in db and saved correctly.

plugins/star-rating/tests.js
```diff
@@ -7,55 +7,255 @@ var API_KEY_ADMIN = "";
...
+            APP_KEY = testUtils.get("APP_KEY");
+            WIDGET_ID = testUtils.get("WIDGET_ID");
+
+            var events = [{
+                "key": "[CLY]_star_rating",
+                "count": 1,
+                "sum": 1,
+                "segmentation": {
+                	"contactMe":true,
+                	"email":"fb@count.ly",
```
ar2rsawseen: Maybe better anonimize this

plugins/star-rating/tests.js
```diff
@@ -7,55 +7,255 @@ var API_KEY_ADMIN = "";
...
+            request.get('/o/feedback/multiple-widgets-by-id?app_id=' + APP_ID + '&api_key=' + API_KEY_ADMIN + '&widgets=' + JSON.stringify(array4query))
+            .expect(200)
+            .end(function(err, res) {
+                if (err) return done(err);
+                var ob = JSON.parse(res.text);
+                ob.length.should.eql(0);
+                setTimeout(done, 100);
+            });
+        });
+    });
```
ar2rsawseen: Here we should also verify feedback data, that it is empty and was removed on reset

plugins/star-rating/tests.js
```diff
@@ -7,55 +7,255 @@ var API_KEY_ADMIN = "";
...
-            .end(function(err, res) {
+            request.get('/o?method=star&period=60days&api_key=' + API_KEY_ADMIN + '&app_id=' + APP_ID).end(function(err, res) {
                 res.statusCode.should.equal(200);
                 data = JSON.parse(res.text);
-                console.log(data);
                 should(data.iOS && data.iOS.indexOf('1:23') >= 0).equal(true);
                 done();
             });
         });
     });
```
ar2rsawseen: I would want 2 more things here
1) create another widget but do not remove it, so we could check if resetting app also removes widgets
2) Create rating/feedback without widget, to ensure backwards compatibility with old approach for older SDKs

plugins/star-rating/tests.js
```diff
@@ -7,55 +7,255 @@ var API_KEY_ADMIN = "";
...
+            .expect(200)
+            .end(function(err, res) {
+                if (err) return done(err);
+                var ob = JSON.parse(res.text);
+                ob.length.should.eql(0);
+                setTimeout(done, 100);
+            });
+        });
+    });
+
```
ar2rsawseen: Here we should also test that rating/feedback data is empty

PR #622, opened 2018-08-21
update crashes logic

waiterZen:


ar2rsawseen:
Was there any ticket for this?
Last I know it was agreed to have 2 graph lines for crash/sessions.
It does not really make sense to show total crashes per session, as you most probably won't care about non fatal ones eitherway.

osoner:
It's either like this or we have single line, previous period ghost line
and no labels below the graph. We chose 3 lines in case fatal/non-fatal
info might be important.

bin/countly.install_ubuntu.sh
```diff
@@ -118,5 +118,6 @@ then
     else
         echo "Cannot reach Google. Disabling Google services. You can enable this from Configurations later."
         countly config "frontend.use_google" false
+        countly plugin enable EChartMap
```
ar2rsawseen: It would be great if we could do that before generating production files (`cd $DIR && grunt dist-al`), or else we need to generate them here again

PR #625, opened 2018-08-22
email reports placeholders

prikshittekta:


plugins/reports/frontend/public/templates/drawer.html
```diff
@@ -17,7 +17,7 @@
 
         <div class="section">
             <div class="label" data-localize='reports.emails-title' ></div>
-            <select id="email-list-input">
+            <select id="email-list-input" placeholder="Enter email">
```
ar2rsawseen: We can localize placeholder too, either by providing localized value as template variable in Handlebars, or using placeholder namepsace, like
`data-localize="placeholder.my-string"`

PR #636, opened 2018-08-29
[feedback] final changes.

frknbasaran:
- two api and one ui change included.


ar2rsawseen:
app_id change fixed on master branch already
UI change hard to merge, cause master has UI changed already
Widget edit change, incorrect fix, you need to use $set operator instead of replacing whole document, @Cookiezaurs will fix it on master

PR #642, opened 2018-09-05
Test timing tweaks and adding code coverage functionality

ArtursKadikis:


bin/config/supervisord.conf
```diff
@@ -22,6 +22,7 @@ programs=countly-dashboard, countly-api
 [program:countly-dashboard]
 environment=NODE_ENV=production
 command=node %(here)s/../../frontend/express/app.js
+;command=nyc node %(here)s/../../frontend/express/app.js ;replace previous line with this in case you want to run nyc
```
ar2rsawseen: We can't do this, it will run in prod like that
Better create separate config and copy it over inside travis installer script

ArtursKadikis: but it's commented out... :D

PR #645, opened 2018-09-06
guidelines changes.

frknbasaran:
- lint changes for dbviewer, ratings and populator plugins.


plugins/dbviewer/frontend/public/javascripts/countly.models.js
```diff
@@ -28,20 +29,20 @@
             data: data,
             success: function(json) {
                 _data = json;
-                for (var i = 0; i < _data.length; i++) {
+                for (let i = 0; i < _data.length; i++) {
```
ar2rsawseen: If only it would be that easy, but no let in browser, sorry :(

plugins/populator/frontend/public/javascripts/countly.models.js
```diff
@@ -805,14 +888,18 @@
...
+    **/    
     function generateRetention(callback) {
         if (typeof countlyRetention === "undefined") {
             callback();
             return;
         }
         var ts = endTs - 60 * 60 * 24 * 9;
         var ids = [ts];
-        var users = 10;
+        users = 10;
```
ar2rsawseen: This approach is very dangerous, with var it would create local function scope variable and be ok with it,
but without var, it would change the upper scope variable, change its value and lead to unpredictable results

PR #667, opened 2018-09-12
Crash plugin guideline fixes.

melihkorkmaz:


plugins/crashes/frontend/public/javascripts/countly.views.js
```diff
@@ -1386,15 +1387,15 @@ window.CrashgroupView = countlyView.extend({
...
 
             $("#tabs").tabs({
-                select: function(event, ui) {
+                select: function() {
                     $(".flot-text").hide().show(0);
                 }
             });
             $("#crash-notes").click(function() {
-                var crashData = countlyCrashes.getGroupData();
+                crashData = countlyCrashes.getGroupData();
```
ar2rsawseen: Such changes can be really dangerous, as you really change the variable from upper scope, and it may still be used in some event callback and would now have unexpected content

PR #696, opened 2018-09-28
[api] common.validateArgs method improvements.

frknbasaran:
- added optional parameter "returnErrors" to common.validateArgs() method.
- if returnErrors parameter passed as true, validateArgs method returns object with this format.
```
{
   result:true,
   errors: [],
   obj: {field1:value1, field2:value2}
}
```

or

```
{
   result:false,
   errors: ['Invalid type for arg2'],
   obj: {}
}
```

Old usage for `validateArgs` method

```javascript
if (!(args = common.validateArgs(params.qstring.args, argProps))) {
   log.d('Not enough params to create message: %j', params.qstring.args);
   common.returnMessage(params, 400, 'Not enough args');
   return;
}
```
Error response for old method:

```
{
   result: "Not enough args"
}
```
New usage

```javascript
var validation = common.validateArgs(params.qstring.args, argProps, true);
if (!(args = validation.obj)) {
   log.d('Not enough params to create message: %j', params.qstring.args);
   common.returnMessage(params, 400, 'Errors: ' + validation.errors);
   return;
}
```
Dynamic error response for new method:
```
{
   result: "Errors: Missing _id argument."
}
```

### DIFF
--- a/api/utils/common.js
+++ b/api/utils/common.js
@@ -677,20 +677,38 @@ common.getISOWeeksInYear = function(year) {
 * @param {string} argProperties.has-char - should string property has any latin character in it
 * @param {string} argProperties.has-upchar - should string property has any upper cased latin character in it
 * @param {string} argProperties.has-special - should string property has any none latin character in it
-* @returns {object|false} validated args or false if args do not pass validation
+* @param {boolean} returnErrors - return error details as array or only boolean result
+* @returns {object} validated args in obj property, or false as result property if args do not pass validation and errors array
 */
-common.validateArgs = function(args, argProperties) {
+common.validateArgs = function(args, argProperties, returnErrors) {
 
-    var returnObj = {};
-
+    if (arguments.length == 2) returnErrors = false;
+    if (returnErrors) {
+        var returnObj = {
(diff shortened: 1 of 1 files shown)


api/utils/common.js
```diff
@@ -677,60 +677,145 @@ common.getISOWeeksInYear = function(year) {
...
 
     if (!args) {
-        return false;
+        if (returnErrors) {
+            returnObj.result = false;
+            returnObj.errors.push("Missing 'args' parameter.");
+            return returnObj;
+        }
+        else {
+            return false;
```
prikshittekta: If the validateArgs function returns false (i.e. a boolean value) - 
var validation = common.validateArgs(params.qstring.args, argProps, false); 
//suppose validation is false
if (!(args = validation.obj)) { //returns true
   log.d('Not enough params to create message: %j', params.qstring.args);
   common.returnMessage(params, 400, 'Errors: ' + validation.errors); // output - 'Errors: undefined'
   return;
}

When we are not displaying proper messages to the users, should we show other text. Because, the api response message 'Errors: undefined' is not quite clear.

PR #706, opened 2018-10-16
upload icon support for app create & edit restful api

waiterZen:


api/parts/mgmt/apps.js
```diff
@@ -151,6 +153,34 @@ appsApi.getAppsDetails = function(params) {
...
+/**
+*  upload app icon function 
+*  @param {params} params - params object with args to create app
+**/
+const iconUpload = function (params) {
+    const appId = params.app_id || params.qstring.args.app_id;
+    if (params.files && params.files.app_image) {
+        const tmp_path = params.files.app_image.path,
+        target_path = __dirname + '/../../../frontend/express/public/appimages/' + appId + ".png",
+        type = params.files.app_image.type;
```
can-keklik: `type` should be declared before assignment
```suggestion
        var type = params.files.app_image.type;
```

api/parts/mgmt/apps.js
```diff
@@ -151,6 +153,34 @@ appsApi.getAppsDetails = function(params) {
...
 
+/**
+*  upload app icon function 
+*  @param {params} params - params object with args to create app
+**/
+const iconUpload = function (params) {
+    const appId = params.app_id || params.qstring.args.app_id;
+    if (params.files && params.files.app_image) {
+        const tmp_path = params.files.app_image.path,
+        target_path = __dirname + '/../../../frontend/express/public/appimages/' + appId + ".png",
```
can-keklik: `target_path` should be declared before assignment
```suggestion
        const target_path = __dirname + '/../../../frontend/express/public/appimages/' + appId + ".png",
```

waiterZen: it already declared as const, as line 163 end with ","  it is a multi-line declaration.

PR #711, opened 2018-10-22
[sub-directory] relative paths changes.

frknbasaran:
- confirm popup image path.
- countly_logo path.
- material-icon.css path.
- compare svg paths.
- populator after generation redirect path.
- populator svg path.
- heatmap logo svg path.


plugins/views/frontend/public/heatmap.js
```diff
@@ -69,7 +69,7 @@
 
             //TOPBAR IMAGE
             var img = document.createElement('img');
-            img.src = Countly.url + "/images/dashboard/countly_logo.svg";
+            img.src = countlyGlobal.path + "/images/dashboard/countly_logo.svg";
```
prikshittekta: I don't think this change is required here.
Mainly bcz this file is loaded on the client web application via the countly-web-sdk, which means countlyGlobal won't be exposed there and this won't work.
So i think it should be left as Countly.url (the url is set by the countly-web-sdk when the client inits the sdk into his application)

PR #712, opened 2018-10-26
[ratings] device_id fix script for broken records.

frknbasaran:
- find all broken app_users
- check is there any correct user for fixed_device_id
- map it correct uid and did to feedback and drill collections, and -remove broken app_user record
- if there is not correct user for fixed_device_id, use it as user and map it to feedback and drill collections

### DIFF
--- a/bin/scripts/fix_broken_ratings.js
+++ b/bin/scripts/fix_broken_ratings.js
@@ -1,0 +1,91 @@
+var pluginManager = require('../../plugins/pluginManager.js'),
+    async = require('async'),
+    crypto = require('crypto'),
+    countlyDb = pluginManager.dbConnection(),
+    countlyDrillDB = pluginManager.dbConnection("countly_drill");
+
+countlyDb.collection('apps').find({}).toArray(function(err, apps) {
+    if (!apps || err) {
+        console.log('No apps to fix');
+        countlyDb.close();
+        countlyDrillDB.close();
+        return;
+    }
+
+    function fixBrokenDeviceIDs(app, done) {
+        // find all broken app_user record containts sdk_name
+        countlyDb.collection('app_users' + app._id).find({did: /.*sdk_name=javascript_feedback_popup/}).toArray(function(err, users) {
+            if (!err) {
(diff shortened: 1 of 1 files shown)


bin/scripts/fix_broken_ratings.js
```diff
@@ -0,0 +1,91 @@
...
+        countlyDrillDB.close();
+        return;
+    }
+
+    function fixBrokenDeviceIDs(app, done) {
+        // find all broken app_user record containts sdk_name
+        countlyDb.collection('app_users' + app._id).find({did: /.*sdk_name=javascript_feedback_popup/}).toArray(function(err, users) {
+            if (!err) {
+                async.forEach(users, function(brokenUser, doneUser) {
+                    let fixed_device_id = brokenUser.did.substr(0, 36);
```
ar2rsawseen: Device id are not fixed length, it can be any custom string provided by a client too, so you need to strip out sdk_name... part specifically

bin/scripts/fix_broken_ratings.js
```diff
@@ -0,0 +1,91 @@
...
+                                                    console.log("Feedback fixed for " + app.name);
+                                                    // if there is correct user for this fixed device_id, remove broken app_user record.
+                                                    if (correctUser) {
+                                                        countlyDb.collection('app_users' + app._id).remove({_id: brokenUser._id}, function(err) {
+                                                            doneUser();
+                                                        });
+                                                    }
+                                                    // if there is no correct user for this fixed device_id, fix this broken app_user record and use it as user.
+                                                    else {
+                                                        countlyDb.collection('app_users' + app._id).findAndModify({_id: brokenUser._id}, {}, {$set: {did: fixed_device_id}}, function(err) {
```
ar2rsawseen: _id field of user document actually depends on device_id (https://github.com/Countly/countly-server/blob/master/api/utils/requestProcessor.js#L1109-L1111) so if device_id changes, then _id should also change.
And you can't change _id of the document in MongoDB, so you need to delete this user, and insert new one with same properties + corrected _id and did

PR #729, opened 2018-11-09
Sub directory

frknbasaran:


plugins/star-rating/frontend/public/templates/feedback-popup.html
```diff
@@ -165,6 +165,8 @@
                 var sdk_name = getParameterByName('sdk_name') || 'javascript_feedback_popup';
                 var app_version = getParameterByName('app_version');
                 var url = window.location.href.split("/")[0] + "//" + window.location.href.split("/")[2];
+                if (window.location.pathname.substr(0,8) === '/countly') url = url + '/countly';
```
ar2rsawseen: We can't hardcode `/countly` path, it must be configurable
the easiest here, is to just remove /feedback path section and use remaining as url

PR #730, opened 2018-11-09
Ratings fix script

frknbasaran:


bin/scripts/fix_broken_ratings.js
```diff
@@ -0,0 +1,100 @@
...
+                                                                countlyDb.collection("feedback" + app._id)
+                                                                    .updateMany({device_id: brokenUser.did}, {
+                                                                        $set: {
+                                                                            device_id: brokenUser.did,
+                                                                            uid: brokenUser.uid
+                                                                        }
+                                                                    }, function(err) {
+                                                                        if(!err) doneUser();
+                                                                    });
+                                                            } else console.log(err);
```
ar2rsawseen: Here doneUser never get's called

bin/scripts/fix_broken_ratings.js
```diff
@@ -0,0 +1,100 @@
...
+                                                            if (!err) {
+                                                                // map correct uid and did values to broken feedback records
+                                                                countlyDb.collection("feedback" + app._id)
+                                                                    .updateMany({device_id: brokenUser.did}, {
+                                                                        $set: {
+                                                                            device_id: brokenUser.did,
+                                                                            uid: brokenUser.uid
+                                                                        }
+                                                                    }, function(err) {
+                                                                        if(!err) doneUser();
```
ar2rsawseen: If there is an error, doneUser never gets called

bin/scripts/fix_broken_ratings.js
```diff
@@ -0,0 +1,100 @@
...
+                                                                    });
+                                                            } else console.log(err);
+                                                        });
+                                                    } else {
```
ar2rsawseen: We can enter this state when there is error, but correctAppUser still does not exist, and script will crash when accessing correctAppUser properties

bin/scripts/fix_broken_ratings.js
```diff
@@ -0,0 +1,100 @@
...
+                                                        // map correct uid and did values to broken feedback records
+                                                        countlyDb.collection("feedback" + app._id)
+                                                            .updateMany({device_id: brokenUser.did}, {
+                                                                $set: {
+                                                                    device_id: correctAppUser.did,
+                                                                    uid: correctAppUser.uid
+                                                                }
+                                                            }, function(err) {
+                                                                if(!err) doneUser();
```
ar2rsawseen: if there is error, doneUser never gets called

PR #732, opened 2018-11-14
Initial screen

frknbasaran:
- first app create ui changed.
- hided unnecessary elements from navbar.
- added first app view support to related plugins.


ar2rsawseen:
I am not sure this one should go into master directly, but rather next branch

PR #735, opened 2018-11-15
changing render host to localhost

prikshittekta:


ar2rsawseen:
Can we not override it in that part, but rather custom dashboard or email reports provides it as options.host? So plugins could still override it, just in case :)
Also we can possibly host countly in subdirectory, so it should take into account countlyConfig.path too, when navigating to views

PR #744, opened 2018-11-28
[dashboard] change /logout api method from GET to POST

waiterZen:


test/1.frontend/2.login.test.js
```diff
@@ -243,17 +256,6 @@ describe('Accessing with login', function() {
...
-    describe('GET /logout', function() {
-        it('should redirect to login', function(done) {
-            agent
-                .get('/logout')
-                .expect('location', '/login')
-                .expect(302, done);
-        });
-    });
-});
-
```
ar2rsawseen: I think we should do POST logout here, or else next test should fail and do not redirect you to login page, shouldn't it?

waiterZen: Right, just rechecked it is affecting the previous login testing.  Thanks :)

PR #747, opened 2018-11-29
[dbviewer] refresh optimization

frknbasaran:
- instead of refreshing whole view, fetching new needed data.

### DIFF
--- a/plugins/dbviewer/frontend/public/javascripts/countly.views.js
+++ b/plugins/dbviewer/frontend/public/javascripts/countly.views.js
@@ -52,12 +52,11 @@ window.DBViewerView = countlyView.extend({
             $('#dbviewer').css({"height": (window.innerHeight - 150) + "px"});
             $('#accordion > div').css({"height": (window.innerHeight - 150) + "px"});
         }
-
         if (store.get('dbviewer_selected_app')) {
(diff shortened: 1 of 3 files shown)


plugins/dbviewer/frontend/public/javascripts/countly.views.js
```diff
@@ -36,34 +79,53 @@ window.DBViewerView = countlyView.extend({
...
-                $('#accordion > div').css({"height": (window.innerHeight - 150) + "px"});
-            }
+        $('body').on('click', '.db-detail', function() {
+            self.db = $(this).data('db');
+            self.renderCollectionList();
+        });
 
-            if (store.get('dbviewer_selected_app')) {
-                $('#app-selector').html(store.get('dbviewer_selected_app').name);
+        if (window.location.hash.split('/').length === 4) {
```
ar2rsawseen: You cannot rely on `window.location.hash` because depending on server settings, sometimes it can include app_id, sometimes it won't, so the paths will differ

Instead use `Backbone.history.getFragment`
https://github.com/Countly/countly-server/blob/master/frontend/express/public/javascripts/countly/countly.template.js#L3272

which returns hash fragment without app_id, always.

PR #774, opened 2018-12-28
[cohort-widget] updated getDashboardData() method.

frknbasaran:
- updated getDashboardData() method.

### DIFF
--- a/frontend/express/public/javascripts/countly/countly.common.js
+++ b/frontend/express/public/javascripts/countly/countly.common.js
@@ -2795,12 +2795,13 @@
 
             if (_periodObj.isSpecialPeriod) {
                 isEstimate = true;
+
                 for (j = 0; j < (_periodObj.currentPeriodArr.length); j++) {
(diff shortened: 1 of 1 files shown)


ar2rsawseen:
For what property exactly did you face undefined behavior?
Because `tmp_x = clearObject(tmp_x);` should have 0 in all required properties.

frknbasaran:
I send cohort data from API for widget, but somehow in that method returning some values as NaN. When I debugged I noticed because of that behavior. It was looks to good idea to add OR 0 to me.

Here is where data came from exactly.
https://bitbucket.org/osoner/countly-enterprise-plugins/pull-requests/215/cohorts-widget-created-cohorts-widget/diff#Lplugins/cohorts/api/api.jsT432

And here is how formatted data that received from server.
https://bitbucket.org/osoner/countly-enterprise-plugins/pull-requests/215/cohorts-widget-created-cohorts-widget/diff#Lplugins/cohorts/frontend/public/javascripts/countly.views.jsT518

Just combined "i" and "o" values as "total" in response object.

PR #775, opened 2018-12-31
[aggregate-support] aggregate support added to dbviewer api.

frknbasaran:
- SERVER-1097
- now db viewer api accepts parameter below for aggregation support
-- aggregate=1 // required
-- match={"key":"val"}
-- group={"_id":"$platform", "average_rating": {"$avg": "$rating"}}

### DIFF
--- a/plugins/dbviewer/api/api.js
+++ b/plugins/dbviewer/api/api.js
@@ -263,6 +263,12 @@ var common = require('../../../api/utils/common.js'),
         * @param {object} aggregation - aggregation object
         * */
         function aggregate(collection, aggregation) {
+            if (params.qstring.skip) {
+                aggregation.push({"$skip": parseInt(params.qstring.skip)});
+            }
+            if (params.qstring.limit) {
+                aggregation.push({"$limit": parseInt(params.qstring.limit)});
+            }
             dbs[dbNameOnParam].collection(collection).aggregate(aggregation, function(err, result) {
                 if (!err) {
                     common.returnOutput(params, result);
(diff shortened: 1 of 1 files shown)


plugins/dbviewer/api/api.js
```diff
@@ -275,6 +294,63 @@ var common = require('../../../api/utils/common.js'),
                     });
                 }
             }
+            else if ((params.qstring.dbs || params.qstring.db) && params.qstring.collection && params.qstring.collection.indexOf('system.indexes') === -1 && params.qstring.collection.indexOf('sessions_') === -1 && params.qstring.aggregate) {
+                var match, group = {};
+                if (params.member.global_admin) {
+                    if (params.qstring.group) {
+                        try {
+                            if (params.qstring.match) {
+                                match = JSON.parse(params.qstring.match) || {};
```
prikshittekta: match will never be assigned the {}, if incase JSON.parse fails.
I think you should initialize match = {}, just like done with group.

plugins/dbviewer/api/api.js
```diff
@@ -275,6 +294,63 @@ var common = require('../../../api/utils/common.js'),
...
+                    else {
+                        common.returnMessage(params, 400, 'Group object must be provided for aggregation');
+                    }
+                }
+                else {
+                    dbUserHassAccessToCollection(params.qstring.collection, function(hasAccess) {
+                        if (hasAccess) {
+                            if (params.qstring.group) {
+                                try {
+                                    match = JSON.parse(params.qstring.match) || {};
```
prikshittekta: same goes here.

plugins/dbviewer/api/api.js
```diff
@@ -257,6 +257,25 @@ var common = require('../../../api/utils/common.js'),
...
+        /**
+        * Get aggregated result by the parameter on the url
+        * @param {string} collection - collection will be applied related query
+        * @param {object} match - match object for aggregation
+        * @param {object} group - group object for aggregation
+        * */
+        function aggregate(collection, match, group) {
+            dbs[dbNameOnParam].collection(collection).aggregate([
+                { $match: match },
+                { $group: group }
```
ar2rsawseen: This is very limited aggregation
It would be better to accept single parameter `pipeline` or the same `aggregation` param you have, which will array of commands and pass them to aggregate method directly

PR #778, opened 2019-01-03
[version-info] added more configuration options to version.info file.

frknbasaran:
- documentationLink, feedbackLink properties set true as default.
- users can change that properties as true or false, also can pass custom url as string.


frontend/express/views/dashboard.html
```diff
@@ -167,18 +167,32 @@
 							<span data-localize="sidebar.settings"></span>
 						</a>
 						<% if (!countlyType) { %>
-							<% if(intercom && member['global_admin'] || config.autonomous) { %>
-							<a class="item activate_intercom">
+							<% if((intercom && member['global_admin'] || config.autonomous) && (typeof feedbackLink === "string")) { %>
+							<a href="<%= feedbackLink %>" class="item activate_intercom">
```
ar2rsawseen: In this case link is both intercom and feedback link. Don't they clash?

PR #790, opened 2019-01-15
auth token issue headless chrome

prikshittekta:


api/utils/render.js
```diff
@@ -63,6 +63,22 @@ exports.renderView = function(options, cb) {
...
+            page.on('response', response => {
+                if (response.status() != 200) {
+                    console.log("HEADLESS CHROME Response - ", response.status(), response.url());
+                }
+            });
+
+            page.on('requestfailed', request => {
+                console.log("HEADLESS CHROME Failed Request - ", request.failure().errorText, request.url());
+            });
+
```
ar2rsawseen: Not sure if we should include pure console log statements for all renders
You can try to use our default log utility to get log level and if it is debug, then run this code:
https://code.count.ly/server/api/module-api_utils_log.html#.getLevel

PR #796, opened 2019-01-21
[overview-bars] created generic getTop() method for overview screen and integrated plugins.

frknbasaran:
- created countly.analytics.js library for generic getTop() method.
- integrated mobile, web and desktop plugins.

### DIFF
--- a/frontend/express/public/javascripts/countly/countly.analytics.js
+++ b/frontend/express/public/javascripts/countly/countly.analytics.js
@@ -1,21 +1,16 @@
-/*global countlyAnalyticsAPI, $, countlyCommon, countlyGlobal */
+/*global countlyAnalyticsAPI, $, countlyCommon */
 (function() {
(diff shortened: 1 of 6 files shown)


frontend/express/public/javascripts/countly/countly.analytics.js
```diff
@@ -0,0 +1,30 @@
...
+            },
+            dataType: "json",
+            success: function(json) {
+                countlyAnalyticsAPI.data = json;
+            }
+        });
+    };
+
+    countlyAnalyticsAPI.getTop = function(metric) {
+        return countlyAnalyticsAPI.data[metric];
```
prikshittekta: if countlyAnalyticsAPI.data is undefined or null, this will throw an error.
return countlyAnalyticsAPI.data && countlyAnalyticsAPI.data[metric] should work.

PR #806, opened 2019-01-30
[update-geo-ip] update geo ip job created and scheduled for 30 days.

frknbasaran:


api/jobs/updateGeoIp.js
```diff
@@ -0,0 +1,25 @@
...
+    /**
+     * Run the update geo-ip job
+     * @param {done} done callback
+     */
+    run(done) {
+        log("Updating geo-ip's..");
+        exec('countly update geoip', (err) => {
+            if (err) {
+                log("Error:" + err);
+                return;
```
prikshittekta: incase of error, done callback will never be called.
is it the required behaviour?
shouldn't it call the done callback?

prikshittekta: https://github.com/Countly/countly-server/blob/428535f64ace72224f3d5db3c4d178b4bae04c3f/api/parts/jobs/job.js#L804

prikshittekta: is there no need for "countly update devices"?

api/jobs/updateGeoIp.js
```diff
@@ -0,0 +1,31 @@
...
+            }
+            log("Updating devices...");
+            exec("countly update devices", (err) => {
+                if (err) {
+                    log("Error occurred while updating devices:" + err);
+                }
+                done();
+            });
+        });
+    }
```
prikshittekta: in this whole scenario, callback(done) is being called twice
1. if error occurs
2. after countly update device

is this desired?

prikshittekta: point being, incase of error, after the done callback is called,
the countly update devices command will run anyway.

PR #818, opened 2019-02-06
[slipping-away-users] change query

waiterZen:


plugins/slipping-away-users/api/api.js
```diff
@@ -22,9 +22,9 @@ const plugins = require('../../pluginManager'),
             const dayOfYear = now.dayOfYear();
 
             periods.forEach((p) => {
-                timeList[p] = now.dayOfYear(dayOfYear - p).utc().unix();
+                timeList[p] = now.dayOfYear(dayOfYear - p).utc().unix() * 1000;
                 conditions.push({
-                    ls: {$lt: timeList[p]},
+                    lac: {$lt: timeList[p]},
```
ar2rsawseen: I think we also need to change in the view on click action to use `lac` instead of `ls`, when viewing users in user profiles list

PR #826, opened 2019-02-08
Remote config 

prikshittekta:


api/parts/data/usage.js
```diff
@@ -268,6 +268,76 @@ usage.processLocation = function(params) {
...
+};
+
+/**
+ * Set user location in params
+ * @param  {params} params - params object
+ * @param  {object} loc - location info
+ */
+usage.setUserLocation = function(params, loc) {
+    params.user.country = loc.country;
+    params.user.city = loc.city;
```
ar2rsawseen: Here I think we also have a region now in core, does it makes sense to have it here too?
https://github.com/Countly/countly-server/blob/next/api/parts/data/usage.js#L136

api/utils/requestProcessor.js
```diff
@@ -89,6 +89,20 @@ const processRequest = (params) => {
     if (!params.req || !params.req.url) {
         return common.returnMessage(params, 400, "Please provide request data");
     }
+
+    if (params.res) {
+        //PROXY THE RESPONSE WRITE FUNCTION HERE TO SET RESPONSE OBJECT TO PARAMS
+        var write = params.res.write.bind(params.res);
+        params.res.write = function(data, encoding, callback) {
```
ar2rsawseen: There are actually cases where we do not write response, but return response in a callback. Like using requestProcessor programmatically. Maybe it makes more sense rather than overwriting write function, to better cover 3 response methods that we have here:
https://github.com/Countly/countly-server/blob/next/api/utils/common.js#L974-L1132
to store response in params as object. And not messing with nodejs internals too :)

api/utils/requestProcessor.js
```diff
@@ -2148,6 +2202,132 @@ const validateAppForWriteAPI = (params, done, try_times) => {
...
+ * @param  {function} done - callback when processing done
+ * @returns {function} done
+ */
+const ignorePossibleDevices = (params, done) => {
+    //ignore possible opted out users for ios 10
+    if (params.qstring.device_id === "00000000-0000-0000-0000-000000000000") {
+        common.returnMessage(params, 400, 'Ignoring device_id');
+        common.log("request").i('Request ignored: Ignoring zero IDFA device_id', params.req.url, params.req.body);
+        params.cancelRequest = "Ignoring zero IDFA device_id";
+        plugins.dispatch("/sdk/cancel", {params: params});
```
ar2rsawseen: If IDFA is 0, this event will log request in logger, 
I think we have some flag if log was already processed or not for request, but just noting :)

api/utils/requestProcessor.js
```diff
@@ -2148,6 +2202,132 @@ const validateAppForWriteAPI = (params, done, try_times) => {
...
+        }
+
+        params.app_id = app._id;
+        params.app_cc = app.country;
+        params.app_name = app.name;
+        params.appTimezone = app.timezone;
+        params.app = app;
+        params.time = common.initTimeObj(params.appTimezone, params.qstring.timestamp);
+
+        if (params.app.checksum_salt && params.app.checksum_salt.length) {
```
ar2rsawseen: Maybe we can have single method for checksum check for both write and read functions?
To avoid code duplication etc

api/parts/data/usage.js
prikshittekta: So far region is not yet being used in the metrics.
Also we only use city in the metrics as it is specified as a default user prop.
It won't matter if region is present or not.
But will add region anyhow, maybe in future it could be used.

api/utils/requestProcessor.js
prikshittekta: done.

PR #839, opened 2019-02-13
Server 1315

waiterZen:


plugins/alerts/api/api.js
```diff
@@ -230,4 +230,19 @@ const _ = require('lodash');
         }, paramsInstance);
         return true;
     });
+
+    plugins.register("/i/apps/delete", function(ob) {
+        var appId = ob.appId;
+        common.db.collection('alerts').remove({selectedApps: {$all: [appId]}}, function() {});
```
ar2rsawseen: Here we don't need to remove job itself? only document?

PR #841, opened 2019-02-15
added a project editorconfig file

LVMBDV:
I propose we add an editorconfig file to help contributors by allowing them to start hacking around right away without adjusting their editor settings according to the project whitespace conventions :)

For more info, see: https://editorconfig.org

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = false
+
+[*.{js,html}]
+indent_style = space
+indent_size = 4


ar2rsawseen:
My only concern is, how does it work together with eslint. If editor finds both, which is more important?
Cause eslint has much more restrictions and they are based on file paths too

LVMBDV:
Well there are tools to generate editorconfig files from eslint configs but editorconfig is just a convenience to help the developer adjust some editor settings while they are working on the project, not a linter like eslint. In short, editorconfig helps you stick to the project's indentation rules, eslint checks if you have been sticking to all the rules of the project, e.g. `space-before-blocks`, `brace-style`. There shouldn't be a case of a conflict as the editorconfig should be adjusted accordingly to the eslintrc as it changes.

PR #860, opened 2019-03-06
[async-plugins-check] server-1248

frknbasaran:
- added async plugins check mechanism to plugins plugin.

### DIFF
--- a/plugins/plugins/api/api.js
+++ b/plugins/plugins/api/api.js
@@ -58,10 +58,10 @@ var plugin = {},
     plugins.register('/o/plugins-check', function(ob) {
         var params = ob.params;
         ob.validateUserForDataReadAPI(params, function() {
-            common.db.collection('plugins').count({"_id": "failed"}, function(err, failedCount) {
-                if (!err && failedCount < 1) {
-                    common.db.collection('plugins').count({"_id": "busy"}, function(err, count) {
-                        if (err) {
+            common.db.collection('plugins').count({"_id": "failed"}, function(failedErr, failedCount) {
+                if (!failedErr && failedCount < 1) {
+                    common.db.collection('plugins').count({"_id": "busy"}, function(busyErr, count) {
+                        if (busyErr) {
                             common.returnMessage(params, 200, "failed");
                         }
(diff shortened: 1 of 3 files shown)


plugins/plugins/api/api.js
```diff
@@ -36,12 +38,12 @@ var plugin = {},
...
-                        process.send({ cmd: "endPlugins" });
-                        if (err) {
-                            common.returnOutput(params, 'Errors');
+                        if (!err) {
+                            process.send({ cmd: "endPlugins" });
+                            updatePluginState("end");
                         }
                         else {
-                            common.returnOutput(params, 'Success');
+                            common.returnMessage(params, 500, "failed");
```
prikshittekta: I don't think a (failed) response will be sent from here because you have already sent the (started) message in line no. 27

plugins/plugins/frontend/public/javascripts/countly.views.js
```diff
@@ -195,35 +195,40 @@ window.PluginsView = countlyView.extend({
...
             }
-            else {
-                overlay.hide();
-                loader.hide();
-                msg.title = jQuery.i18n.map["plugins.error"];
-                msg.message = res;
-                msg.info = jQuery.i18n.map["plugins.retry"];
-                msg.sticky = true;
-                msg.type = "error";
+            else if (res.result === "error") {
```
prikshittekta: shouldn't this be res.result == "failed" since nowhere you are returning error message from the server.

plugins/plugins/frontend/public/javascripts/countly.views.js
```diff
@@ -195,35 +195,40 @@ window.PluginsView = countlyView.extend({
...
+                        }
+                    });
+                }, 5000);
             }
-            else {
-                overlay.hide();
-                loader.hide();
-                msg.title = jQuery.i18n.map["plugins.error"];
-                msg.message = res;
-                msg.info = jQuery.i18n.map["plugins.retry"];
```
prikshittekta: This case is not covered, with the retry message.

PR #865, opened 2019-03-18
[SERVER-1220] Control all graph colors from javascript instead of CSS

frknbasaran:


mt-sebastien-robert:
Awesome work, will be very helpful for us too.

The function `hexToRgb` could check the length of the input in order to deduce if the color contains an alpha channel, in stead of truncating to 6 characters.
Beware, in such a case red has to be shifted by 24, green by 16, blue by 8 and alpha by 0.

frknbasaran:
hello @mt-sebastien-robert 

Thanks for your feedback, but `hexToRGB` method is not using for all graph color convertions, it's only for the times of day plugin. The place that we using that method should accept alpha channel from our times of days algorithm. Because the opacity value depend data density for related time range.
I placed below the link of where and how are we using exactly that method.
https://github.com/Countly/countly-server/blob/7078614871ee539f105aa3a9a85dddb231cfa1db/plugins/times-of-day/frontend/public/javascripts/countly.models.js#L239

mt-sebastien-robert:
I see, thanks for the explanation.

PR #875, opened 2019-03-29
[Reports-manager] change table render to server side 

waiterZen:


ar2rsawseen:
Very good
the only additional thing to consider it data export 😃

PR #880, opened 2019-04-10
[server-1397] feedback widget system logs wordings changed.

frknbasaran:


plugins/star-rating/api/api.js
```diff
@@ -66,7 +66,7 @@ var exported = {},
             common.db.collection(collectionName).insert(widget, function(err) {
                 if (!err) {
                     common.returnMessage(ob.params, 201, "Success");
-                    plugins.dispatch("/systemlogs", {params: params, action: "Widget added", data: widget});
+                    plugins.dispatch("/systemlogs", {params: params, action: "Feedback widget added", data: widget});
```
ar2rsawseen: Can we better have them as localized strings, so the code here would not actually change in future, but rather only localized string changes here:
https://github.com/Countly/countly-server/blob/master/plugins/systemlogs/frontend/public/localization/systemlogs.properties#L58

plugins/star-rating/api/api.js
```diff
@@ -2,7 +2,8 @@ var exported = {},
     common = require('../../../api/utils/common.js'),
     crypto = require('crypto'),
     countlyCommon = require('../../../api/lib/countly.common.js'),
-    plugins = require('../../pluginManager.js');
+    plugins = require('../../pluginManager.js'),
+    localize = require('../../../api/utils/localization.js');
```
ar2rsawseen: Now you don't need the localization module here then :)

frknbasaran: yep I removed that dependencies on my last commit.
https://github.com/Countly/countly-server/pull/880/commits/5bf4d7b5a07b01ec15fba08f4f20b7a5f7471fa5

PR #906, opened 2019-05-22
[server-1364] added optional member image and custom profile image upload feature.

frknbasaran:


frontend/express/public/javascripts/countly/countly.views.js
```diff
@@ -3328,6 +3335,55 @@ window.ManageUsersView = countlyView.extend({
         $(".change-password").off("click").on('click', function() {
             $(this).parents(".row").next().toggle();
         });
+
+        $('body').off('change').on('change', '.pp-uploader', function() {
+            $('.pp-menu-list').hide();
+            var member_id = $(this).data('member-id');
+            $(this).simpleUpload("/member/icon", {
```
ar2rsawseen: If we use library in the core, we should better add it to the core and not to the plugin. And it might be useful to other parts too may be.
And don't forget to add it to gruntfile too.

frontend/express/public/javascripts/countly/countly.views.js
```diff
@@ -3328,6 +3335,55 @@ window.ManageUsersView = countlyView.extend({
...
+                        }
+                    }
+                },
+                error: function() {
+                    CountlyHelpers.notify(jQuery.i18n.map["plugins.errors"]);
+                }
+            });
+        });
+
+        $('.delete-symbol').on('click', function() {
```
ar2rsawseen: What does symbol has to do with this?

frontend/express/public/javascripts/countly/countly.views.js
frknbasaran: added.

frontend/express/public/javascripts/countly/countly.views.js
frknbasaran: https://github.com/Countly/countly-server/pull/906/files#diff-2d40c6d737f10cb1de3e55b86434ca35R3419

ar2rsawseen: Ok, but still let's renamed to something more valid, like `delete-member-image`

frontend/express/public/javascripts/countly/countly.views.js
```diff
@@ -3328,6 +3335,55 @@ window.ManageUsersView = countlyView.extend({
         $(".change-password").off("click").on('click', function() {
             $(this).parents(".row").next().toggle();
         });
+
+        $('body').off('change').on('change', '.pp-uploader', function() {
```
ar2rsawseen: We need to remove listener only for specific element not all listeners
```
$('body').off('change','.pp-uploader').on('change', '.pp-uploader', function() {
```

frontend/express/public/javascripts/countly/countly.views.js
```diff
@@ -3328,6 +3335,55 @@ window.ManageUsersView = countlyView.extend({
...
+            $('.pp-menu-list > div:nth-child(2)').css({'display': 'none'});
+            $('#pp-circle-' + member_id).prepend('<span style="text-style:uppercase">' + name[0][0] + name[name.length - 1][0] + '</span>');
+            if (member_id === countlyGlobal.member._id) {
+                $('.member_image').html("");
+                $('.member_image').css({'background-image': 'url("images/avatar-sprite.png?now=' + Date.now() + '")', 'background-size': 'auto', 'background-position': defaultAvatarSelector + 'px'});
+                $('.member_image').prepend('<span style="text-style: uppercase;color: white; position: absolute; top: 5px; left: 6px; font-size: 16px;">' + name[0][0] + name[name.length - 1][0] + '</span>');
+            }
+        });
+
+        $('body').off('click').on('click', function(e) {
```
ar2rsawseen: Same here

plugins/plugins/frontend/public/javascripts/countly.views.js
```diff
@@ -851,6 +883,31 @@ window.ConfigurationsView = countlyView.extend({
                 }
             });
 
+            $('.pp-menu-trigger').off('click').on('click', function() {
+                $('.pp-menu-list').show();
+            });
+
+            $('body').off('change').on('change', '#pp-uploader', function() {
```
ar2rsawseen: And here

plugins/plugins/frontend/public/javascripts/countly.views.js
```diff
@@ -871,6 +928,19 @@ window.ConfigurationsView = countlyView.extend({
                 }
             });
 
+            $('body').off('click').on('click', function(e) {
+                if (e.target.className !== 'pp-menu-trigger') {
+                    $('.pp-menu-list').hide();
+                }
+            });
+
+            $('body').off('click').on('click', '.delete-symbol', function() {
```
ar2rsawseen: And here

plugins/plugins/frontend/public/javascripts/countly.views.js
```diff
@@ -871,6 +928,19 @@ window.ConfigurationsView = countlyView.extend({
                 }
             });
 
+            $('body').off('click').on('click', function(e) {
```
ar2rsawseen: And here we need some control mechanism, not removing all click listeners

ar2rsawseen: Assuming you need this for hiding popup, you can try using blur event like we do in other places:
https://github.com/Countly/countly-server/blob/master/frontend/express/public/javascripts/countly/countly.helpers.js#L1119-L1123

PR #910, opened 2019-05-27
[SERVER-1322] Application Locking

LVMBDV:
Application lock logic enables locking an app to prevent accidental delete/clear/reset and running data populator.


LVMBDV:
Fixing eslint errors :)

LVMBDV:
I have no idea why DeepScan decided to point those issues after that eslint fix commit :thinking:. They are unrelated to the changes in this PR.

frontend/express/public/javascripts/countly/countly.views.js
```diff
@@ -1482,19 +1482,6 @@ window.ManageAppsView = countlyView.extend({
...
-        $("#clear-data > li").on("selectstart", function(e) {
-            e.preventDefault();
-        });
-        $("#clear-app-data").on("selectstart", function(e) {
-            e.preventDefault();
-        });
-        $("#delete-app").on("selectstart", function(e) {
-            e.preventDefault();
-        });
-
```
ar2rsawseen: Any reason why these were removed?
I understand they were replaces with other button ids?
But maybe the new buttons need the same?
Just rechecking

plugins/populator/frontend/public/javascripts/countly.views.js
```diff
@@ -22,6 +22,16 @@ window.PopulatorView = countlyView.extend({
...
+
+        if (countlyGlobal.apps[countlyCommon.ACTIVE_APP_ID].locked) {
+            $("#populator-locked").show();
+            $("#populator > .content").hide();
+        }
+        else {
+            $("#populator-locked").hide();
+            $("#populator > .content").show();
+        }
+
```
ar2rsawseen: Maybe additionally to browser side protection, we can also do server side protetions
1) Add populator=true parameter to all SDK populator requests
2) On server side check for each SDK request and if app is locked and request contains populator=true, then block request (both /i and /i/bulk) like this: https://github.com/Countly/countly-server/blob/master/api/utils/requestProcessor.js#L2034

That will also help us identify populator requests in Event logs too

frontend/express/public/javascripts/countly/countly.views.js
LVMBDV: You're right, these should be adapted to the new menu buttons, on it 👍

plugins/populator/frontend/public/javascripts/countly.views.js
LVMBDV: I wanted to block populator SDK requests server-side as well but I couldn't decide if it fit in the scope of this feature's requirements or was worth the overhead it would add to the SDK request processor. I'm down to implement it in this PR if you authorize it :)

ar2rsawseen:
Yep, let's do that :)

api/utils/rights.js
```diff
@@ -345,15 +350,30 @@ exports.validateUser = function(params, callback, callbackParam) {
...
                 params.member = member;
 
-                if (plugins.dispatch("/validation/user", {params: params})) {
-                    if (!params.res.finished) {
-                        common.returnMessage(params, 401, 'User does not have permission');
-                        reject('User does not have permission');
-                    }
-                    return false;
-                }
+                // The populator creates push messages for one app so params.qstring.apps[0] should be fine in this case
```
ar2rsawseen: Let's not do validation for app lock on `validateUser` as it has nothing to do with app context and app_id should never be passed here.

LVMBDV: Well, the populator uses a lot of plugins that use `validateUserForWriteAPI` which is just `validateUser` (https://github.com/Countly/countly-server/blob/next/api/utils/requestProcessor.js#L19). Here are a few examples:
- https://github.com/Countly/countly-server/blob/next/plugins/push/api/api.js#L122
- https://github.com/Countly/countly-server/blob/next/plugins/star-rating/api/api.js#L16

In fact, nothing uses `validateAppForWriteAPI` besides `/i/` and `/i/bulk` inside `requestProcessor.js` so I assume it's for sdk requests only. If so, we need an app write validator for non-sdk create/update requests and add it to wherever needed (most CE and EE plugins) or just hack some app validation into the validator most stuff is already using :smile:

PR #914, opened 2019-05-28
[server-1429] added custom field support dbviewer sort.

frknbasaran:


plugins/dbviewer/frontend/public/javascripts/countly.views.js
```diff
@@ -839,7 +871,7 @@ window.DBViewerView = countlyView.extend({
             });
 
             // when the filter button fired
-            $("#dbviewer-apply-filter-button").on('click', function() {
+            $('body').off('click').on('click', "#dbviewer-apply-filter-button", function() {
```
ar2rsawseen: This is very bad, as it removes all click listeners from body, it should be
```
$('body').off('click', "#dbviewer-apply-filter-button").on('click', "#dbviewer-apply-filter-button", function() {
```

PR #920, opened 2019-05-31
[server-1425] cleared properties before db comparison.

frknbasaran:


api/utils/common.js
```diff
@@ -2036,8 +2036,8 @@ common.checkDatabaseConfigMatch = (apiConfig, frontendConfig) => {
                 // mongodb0.example.com:27017
                 let apiMongoHost = apiConfig.split("/")[2];
                 let frontendMongoHost = frontendConfig.split("/")[2];
-                let apiMongoDb = apiConfig.split("/")[3];
-                let frontendMongoDb = frontendConfig.split("/")[3];
+                let apiMongoDb = apiConfig.split("/")[3].split('?')[0];
+                let frontendMongoDb = frontendConfig.split("/")[3].split('?')[0];
```
ar2rsawseen: This would still crash on malformed database connection string like "test"

PR #923, opened 2019-06-10
Improve docker image

### LINKED ISSUE #921: Run countly with non-root in docker
For security reason our cluster is configured to run the docker container with random UIDs.

When I use this image everything breaks because of many permission problems. It starts with the base image which tries to access a file which is only accessible for the docker deamon.

I would appreciate to have this docker image follow the cloud native best practices and have the core image also published on Docker Hub.

sapkra:
Fixes #921 

#### What I did:
- Updated dependencies
- Use smaller debian-lite instead of ubuntu as base image
- Container will run without root permissions now
- Removed mongodb & nginx from image **(breaking change)**
- Added docker-compose.yml
- Simplified install scripts and fix some linting errors using ShellCheck
- Improved build speed of image
- Choose which countly service should be started via environment variable

#### TODO:
- [ ] test docker-compose
- [ ] test install scripts for non docker systems
- [ ] test complete container functionality
- [ ] add example kubernetes manifests
- [ ] add persistent volume for plugins etc.
- [ ] install plugins without necessary restart?

#### Problems:
-  Localization error in frontend:
    - `GET http://countly.add-analytics.dev.tazeapp.com/localization/min/locale_en.properties?19.02.1&798e88b356bca172862865bd5c80a9e74aa75be3 404 (Not Found)`
    - `Cannot GET /localization/min/locale_en.properties`
- Database error:
```bash
2019-06-15T09:27:33.153+0000 I NETWORK  [conn44] received client metadata from 10.128.0.1:33420 conn44: { driver: { name: "nodejs", version: "3.2.7" }, os: { type: "Linux", name: "linux", architecture: "x64", version: "3.10.0-957.10.1.el7.x86_64" }, platform: "Node.js v10.16.0, LE, mongodb-core: 3.2.7", application: { name: "node: countly_fs(10) 20" } }
2019-06-15T09:27:33.153+0000 I ACCESS   [conn44] Supported SASL mechanisms requested for unknown user 'countly@countly_fs'
2019-06-15T09:27:33.166+0000 I ACCESS   [conn44] SASL SCRAM-SHA-1 authentication failed for countly on countly_fs from client 10.128.0.1:33420 ; UserNotFound: Could not find user countly@countly_fs
```
- Can only see page views in the dashboard. Sessions are not shown. My frontend is showing: `Session extended 61` and the Request which also has an `{"result":"Success"}` result.
I can see the latest visitors but the total visitors and new visitors are 0. The data seems to be completely broken.
image


sapkra:
If you want I can also rebase it onto the `next` branch because there are some breaking changes.

iartem:
It doesn't seem to address zombie process issue inside the contnainer, does it?

sapkra:
I don't understand in which case it would appear. Can you explain?

Additionally I think the normal docker way would be to have only one service in a container. So that docker will restart the container automatically when the process crashed.
The api server, the frontend server, nginx and mongodb should be in separated containers. But I had one problem. I couldn't reach the frontend server nor the api server from outside the container because I think they will only respond when the request is asking for `localhost:<PORT>`. Maybe you now an easy way how to fix it.

iartem:
> The api server, the frontend server, nginx and mongodb should be in separated containers. But I had one problem.

That's true. But not in case of your PR as it has (correct me if I'm wrong): at least 2 processes for API, 1 for frontend and at least 2 for nginx. Whenever you send a push notification, at least one more is spawned (usually more).

> I don't understand in which case it would appear. Can you explain?

Please take a look at [this SO question](https://stackoverflow.com/questions/49162358/docker-init-zombies-why-does-it-matter) and [this article](https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/). Basically, it's the reason we use baseimage with `runit` inside. At the moment we made this image supervisor didn't handle the issue and we indeed had zombies. I'm not sure if supervisor does handle it now. 

> I think they will only respond when the request is asking for localhost:<PORT>.

They do, it's in `config.js` in api & frontend folders. But why do you it need to be otherwise? 

Overall I don't get why did you put nginx into your image? Since you use docker compose, wouldn't it be better to have it separately? And maybe building 2 images out of one Dockerfile for API & frontend switching with some `--build-arg`? I'm asking because we cannot just ditch mongo & nginx from our standard image (docker for us is more of a tool for easy trial), but I guess your contribution might work for people who use dockerized Countly in production. But that, in my opinion, would require having api/frontend/nginx separated.

sapkra:
> Please take a look at this SO question and this article. Basically, it's the reason we use baseimage with runit inside. At the moment we made this image supervisor didn't handle the issue and we indeed had zombies. I'm not sure if supervisor does handle it now.

Interesting, I didn't know that. I will read some docs and blogs how the best way is to solve this problem in 2019. Maybe it's enough to provide the `init` option or just use `tini` inside the container like described [here](https://www.elastic.io/nodejs-as-pid-1-under-docker-images/).

> They do, it's in config.js in api & frontend folders. But why do you it need to be otherwise?
Overall I don't get why did you put nginx into your image? Since you use docker compose, wouldn't it be better to have it separately? And maybe building 2 images out of one Dockerfile for API & frontend switching with some --build-arg? I'm asking because we cannot just ditch mongo & nginx from our standard image (docker for us is more of a tool for easy trial), but I guess your contribution might work for people who use dockerized Countly in production. But that, in my opinion, would require having api/frontend/nginx separated.

The reason is that I'm using OpenShift/Kubernetes. Containers are running in a pod, each pod has a service (something like a load balancer) and all pods are communicating over them. The problem is that each service and also each pod has it's own internal IP and are not reachable over localhost. When I use nginx inside the container instead it can communicate to the frontend and API service via localhost and can expose port 8080 outside of the container.
I tried to change the config from `localhost` to the Service IP and also to the domain but neither of them worked. I just got port binding errors.

Another approach would be to put all containers inside one Pod so that they can communicate via localhost but then they are not separately scalable anymore.

iartem:
Sure, Nginx from a different container won't be able to access 3001 / 6001 ports of Countly pod. You'd need to expose these ports from pods/services and use them as k8s DNS service hosts in Nginx configuration. Or, maybe, through environment variables with service IPs. I mean the problem should be solved from another way around, you don't need to change anything in Countly's `config.js`s. 

I guess we'll be doing a "reference design" for k8s in coming months as more and more customers using Countly in k8s. So far the most correct way to do it seems to be:

1. API image/pod/service exposes port 3001. 
2. Frontend image/pod/service exposes port 6001.
3. Nginx role completely removed from setup and left for k8s Ingress which mimics our Nginx role (not sure it'd work yet).
4. MongoDB role intentionally left blank or filled with some single-pod MongoDB stub.

I don't have enough experience with Ingress, do you think it'd work?

The biggest issue with this approach is that we'll need to do something with the ability to change the plugin list and to restart the server from the dashboard.

sapkra:
I seperated the two node process into seperated containers but like I've said earlier I have the problem that I can't reach the servers through a service.

Inside the frontend container
```bash
curl localhost:6001
Found. Redirecting to /login
```

Inside the API container
```bash
curl $COUNTLY_FRONTEND_SERVICE_HOST:$COUNTLY_FRONTEND_SERVICE_PORT
curl: (7) Failed to connect to 172.30.32.43 port 6001: Connection refused
```

The service of the frontend pod
image

iartem:
That's perfectly fine, API container doesn't need frontend one at all.

sapkra:
I've just chosen this for an example because Ingress also tries to access the frontend and api via the Service. That means that neither the frontend nor the api is available via Ingress at all.
I think I know the reason for it and will investigate in your code how to solve it.

iartem:
I have no ideas other than changing `localhost` to `0.0.0.0` in `config.js` (or setting `COUNTLY_CONFIG_FRONTEND_WEB_HOST` env variable instead). And as far as I remember it doesn't need to be `0.0.0.0`, `localhost` should be ok. Therefore, something is wrong with pod or service definition.

sapkra:
Setting `COUNTLY_CONFIG_FRONTEND_WEB_HOST` to `0.0.0.0` fixed the problem somehow. Thanks.

sapkra:
I rebased the branch and added my latest changes.
I also found a few problems which I've documented in the PR description. Maybe you can help with that. 😉

iartem:
Looks great, thanks! Especially "tiny" thing.

Regarding problems:
1. Localization. It's ok if it cannot find only `_en.properties`, we don't need it. Yet there is a problem in case other languages cannot be downloaded as well.

2. Database. [we have it covered here](https://resources.count.ly/docs/securing-mongodb#section-adding-single-user-for-all-countly-databases)

3. Must be something with nginx conf. Check its access logs.

You don't need to fix these as I guess I'll start from your work a bit later. Just FYI, overall issues:
1. What's wrong with `bash smth.sh` as opposed to `node smth.js`? Removing "bin/scripts/countly.install.plugins.sh" affects a ton of things. Breaking change, won't merge.
2. Updated dependencies, won't merge. At least we had issues with `uglify 4`, not sure if it's already fixed. I won't even comment on `"fs.extra": "^1.3.2"` vs ` "fs-extra": "^8.0.1"` :)
3. Supervisor still there, not sure about that. Do logs go to stdout/stderr? I guess supervisor still catches all logs to file, which is not docker-friendly.
4. No `EXPOSE` in Dockerfile?

sapkra:
Regarding problems:
1. It seems like there is only a problem with the English language.
2. Ok, I will have a look at it.
3. I don't use nginx in my environment - I'm using Ingress instead. I will have a look.

Overall issues:
1. In my opinion it makes no sense to have a script which just executes one command. You could execute the commands directly instead. It's just a bit simpler. I thought I changed all references to this file I've deleted or are there any other scripts I missed?

2. If I remember correctly I couldn't find any file where `fs.extra` is used. But generally dependencies shall be regularly updated to fix possible security problems and breaking changes with new node versions. This repository still uses some deprecated dependencies which should be refactored soon. 

3. I don't know what you mean? I don't see any supervisor anymore.

4. I didn't added the `EXPOSE` because this container is opening different ports depending on the environment variable. So shall we add both `3001` and `6001`?

PS: I would be nice to have to code commented instead of one comment with 1. 2. 3....

bin/countly.install_ubuntu.sh
```diff
@@ -52,15 +52,12 @@ apt-get -y install nginx || (echo "Failed to install nginx." ; exit)
...
+if ! (command -v npm >/dev/null) then
     apt-get -y install npm
 fi
 
 #install supervisor
-if [ "$INSIDE_DOCKER" != "1" ]
-then
-	apt-get -y install supervisor || (echo "Failed to install supervisor." ; exit)
-fi
+apt-get -y install supervisor || (echo "Failed to install supervisor." ; exit)
```
iartem: Regarding supervisor, I guess it was this one. It doesn't seem you use it though

iartem:
Regarding ports, I guess it should just expose one port with API / Frontend on it depending on env variable. Did Ingress work out after all?

iartem:
I've finally started working on this, in case you want to take a look: [api](https://github.com/Countly/countly-server/blob/feature/new-whale/Dockerfile-api) & [frontend](https://github.com/Countly/countly-server/blob/feature/new-whale/Dockerfile-frontend). There's a lot from your work there, so thanks! 👍 Yet it's a bit different, mostly in regards to plugins handling.

sapkra:
@iartem Currently I don't have time to test it but I still have the kubernetes manifests I was using to test the environment. I can share them if you want, shall I?

iartem:
@sapkra Actually, some of our customers already use it in production, so I guess we can close this one. Thanks for your help!

sapkra:
@iartem So you don't want that I share my kubernetes manifests? I could create a new PR for them.

iartem:
Sure, why not, maybe you have some additions or improvements.

iartem:
@sapkra Sure, why not, maybe you have some additions or improvements.

PR #929, opened 2019-06-19
Add minifyHTML plugin for production

cmontgomerymt:


frontend/express/app.js
```diff
@@ -397,6 +398,18 @@ app.set('views', __dirname + '/views');
...
+    app.use(minifyHTML({
+        override: true,
+        exception_url: false,
+        htmlMinifier: {
+            removeComments: true,
+            collapseWhitespace: true
+        }
+    }));
+}
+
```
ar2rsawseen: Sorry, for taking so long to review.
But there is only one thing that concerns me, the config should be runtime enabled/disabled.
But in this case this check is run only once on server start. And if I go and disable production mode, then it won't take any effect.

Would it be rather possible to have the production flag check inside the middleware to use it or skip it?

cmontgomerymt: @ar2rsawseen Thank you for the feedback. I just want to clarify where I should move the check.

By `middleware`, do you mean this one?

```
app.use(function(req, res, next) {
    add_headers(req, res);
    next();
});
```

So, the modification could be something like:

```
function check_minify() {
  if (plugins.getConfig("frontend").production) {
    app.use(minifyHTML({
        override: true,
        exception_url: false,
        htmlMinifier: {
            removeComments: true,
            collapseWhitespace: true
        }
    }));
  }
}

app.use(function(req, res, next) {
    add_headers(req, res);
    check_minify();
    next();
});
```
Let me know if I am going about this the right way. Thanks.

mt-sebastien-robert: With this, for each request (`app.use(function(req, res, next) {`) you will register a new middleware (`check_minify()` -> `app.use(minifyHTML(...`).

What you want to do is register your middleware only once (`app.use(function(req, res, next) {`), and in the `check_minify()` function, manually call a minification process (hopefully minifyHtml provide a way to do it manually).

Let me know if what I say is not clear.

ar2rsawseen: Middlewares are just functions, so does not matter if minifyHTML provides it or not, something like this should work 😄 
```
    var minify = minifyHTML({
        override: true,
        exception_url: false,
        htmlMinifier: {
            removeComments: true,
            collapseWhitespace: true
        }
    });
    app.use(function(req, res, next) {
        if (plugins.getConfig("frontend").production) {
            minify(req, res, next);
        }
        else {
          next(); 
       }
    });
```

mt-sebastien-robert: Yeah you can use the middleware as is.
However you don't want the `else` and run the `next()` all the time.

cmontgomerymt:
Updated

mt-sebastien-robert:
@ar2rsawseen Can you have a look at the security issue please ? We can't access Synk.
@cmontgomerymt The Travis issue is due to code formatting, they use `else` on new line, like that:
```
if (condition) {
    // ...
}
else {
    // ...
}
```

ar2rsawseen:
I am checking why tests fail on Travis now, could be that we rely on some html information to be parsed from body for verification, etc. But I will deal with it.

But about Snyk, so it seems this dependency introduces issue there:
image


More on the issue: https://app.snyk.io/vuln/SNYK-JS-LODASHMERGE-173732

So not sure, maybe there are better alternatives or maybe author promised a fix soon?

ar2rsawseen:
Fixed tests to work with minified output

ronkorving:
@cmontgomerymt I think you'll need to update this PR to resolve some conflicts.
Also, please consider using https://www.npmjs.com/package/express-minify-html-2 which contains some bugfixes (express-minify-html seems to be unmaintained).

cmontgomerymt:
We solved it on our end.

PR #931, opened 2019-06-20
[server-1437] Runtime replacement of word Countly.

frknbasaran:


frontend/express/public/javascripts/countly/countly.template.js
```diff
@@ -1224,6 +1224,11 @@ var AppRouter = Backbone.Router.extend({
             path: [countlyGlobal.cdn + 'localization/min/'],
             mode: 'map',
             callback: function() {
+                if (countlyGlobal.company) {
+                    for (var key in jQuery.i18n.map) {
+                        jQuery.i18n.map[key] = jQuery.i18n.map[key].replace("Countly", countlyGlobal.company);
+                    }
+                }
```
ar2rsawseen: As far as I know, this will replace only first occurance of Countly, not others, if there are more than one Countly word in localized string

frontend/express/public/javascripts/countly/countly.common.js
```diff
@@ -228,6 +228,9 @@
         * @returns {string} encode string
         */
         countlyCommon.encodeSomeHtml = function(html, options) {
+            if (countlyGlobal.company) {
+                html.replace("Countly", countlyGlobal.company);
+            }
```
ar2rsawseen: I think ticket meant, not applying replacement in this method, but rather applying encodeSomeHtml to each and every localization string

frontend/express/public/javascripts/countly/countly.template.js
```diff
@@ -1526,6 +1531,11 @@ var AppRouter = Backbone.Router.extend({
                     path: [countlyGlobal.cdn + 'localization/min/'],
                     mode: 'map',
                     callback: function() {
+                        if (countlyGlobal.company) {
+                            for (var key in jQuery.i18n.map) {
+                                jQuery.i18n.map[key] = jQuery.i18n.map[key].replace("Countly", countlyGlobal.company);
+                            }
+                        }
```
ar2rsawseen: Can you also recheck prelogin pages if they are handled, they should be handled in countly/frontend/express/public/javascripts/prelogin.js

mt-sebastien-robert:
So this PR is all about massively replacing "Counly" by "<your-company-name>" ? What about differentiating Countly the company and Countly the product ? (if this makes sense)

ar2rsawseen:
For whitelabeling purposes don't know if it is needed. Mostly references are to Countly dashboard and Countly server.

PR #932, opened 2019-06-24
[server-1179] added update geoip shell script.

frknbasaran:
- scheduled geoip update script for 30 days.

### DIFF
--- a/bin/countly.install_rhel.sh
+++ b/bin/countly.install_rhel.sh
@@ -137,6 +137,9 @@ else
     countly plugin enable EChartMap
 fi
 
+#update geoip dataset and create cron job for this process
+bash $DIR/scripts/geoip.sh
+
 #compile scripts for production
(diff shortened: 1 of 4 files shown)


ar2rsawseen:
So couple of more points

- [ ]  Does it mean we can close this PR https://github.com/Countly/countly-server/pull/806?

- [ ]  Let's not update geoip database on installation, as it will increase install script run time for 5+ minutes

- [ ] bash nodejs script does not seem to run on travis, so we need to understand why and how to ensure it will work on any system, or else move to nodejs script and call it as nodejs script directly

- [ ]  When updating geoip data in crontab, countly process won't know it changed, so we need to enable filesystem watcher https://github.com/bluesmoon/node-geoip#automatic-start-and-stop-watching-for-data-updates
and I want to test it more with watcher to ensure it does not cause any problems

So let's not merge for this release, but rather test more thoroughly while working on next release

PR #979, opened 2019-07-29
[reports] Smart regeneration in report manager

waiterZen:


api/jobs/task.js
```diff
@@ -18,14 +17,41 @@ class MonitorJob extends job.Job {
...
         common.db.collection("long_tasks").find({
             autoRefresh: true,
-            r_hour: targetHour
         }).toArray(function(err, tasks) {
             log.d('Running Task Monitor Job ....');
             log.d("job info:", self._json, tasks);
-            tasks.forEach((task)=>{
+            const filteredTasks = tasks.filter(tasksFilter);
+            filteredTasks.forEach((task) => {
                 return Promise.coroutine(function *() { // eslint-disable-line require-yield
```
ar2rsawseen: I think we also need is running state, or else we might run some long running reports multiple times even before they finished.
And just in case we should reset running flag on master process start, just like we do with cohorts

waiterZen: In jobs  data schema, we dont have "in_progress" property, instead, we prvoide  "status":https://github.com/Countly/countly-server/blob/master/api/parts/jobs/job.js#L13-L19

So here I can add status checking.

ar2rsawseen: I am not talking about the job, but about tasks. They don't have statuses, jobs have.
So what happens if tasks gets run for more then 1 hour?
Next job that will run in 1 hour, will rerun this task again, even though previous one is not yet finished,

ar2rsawseen: here we have status for the task, just recheck this, so you won't run the task, that is already running
https://github.com/Countly/countly-server/blob/master/api/utils/taskmanager.js#L361

waiterZen: OK updated

PR #986, opened 2019-07-30
[SERVER-1460] large numbers

mksglu:


plugins/plugins/frontend/public/javascripts/countly.views.js
```diff
@@ -1118,7 +1118,7 @@ window.ConfigurationsView = countlyView.extend({
             return input;
         }
         else if (typeof value === "number") {
-            return "<input type='number' id='" + id + "' value='" + value + "' max='2147483647' min='0' onkeyup='this.value= (parseInt(this.value) > 2147483647) ? 2147483647 : this.value;'/>";
```
ar2rsawseen: I think we agreed to leave this one

plugins/pluginManager.js
```diff
@@ -1494,22 +1494,36 @@ var pluginManager = function pluginManager() {
...
             }
             else {
-                ob[i] = (!isNaN(ob[i]) && ob[i] > 2147483647) ? 2147483647 : ob[i];
+                const configsKey = Object.keys(configs);
+                const defaultConfigsKey = Object.keys(defaultConfigs);
+                const objectConfigs = {...configs, ...defaultConfigs};
+                const arrConfigsKey = [...configsKey, ...defaultConfigsKey];
+                for (let configKey of arrConfigsKey) {
+                    const objectKey = Object.keys(objectConfigs[configKey]);
+                    objectKey.map(key=>{
```
ar2rsawseen: Here `forEach` maybe is better suitable than `map`

ar2rsawseen: In overall, this is really not optimal, we are traversing all the nodes of object, and inside each traverse, we do an iteration on all possible keys. It creates complexity exponentially.
I think you can just get the needed key value from prefix variable (it's all upper keys with dot as divider) and use that to check defaultConfigs value type in constant time

PR #1020, opened 2019-08-12
Travis

ar2rsawseen:


ronkorving:
@ar2rsawseen Just mentioning this in case you're not aware, but you can test against more than one version of Node at once:

```yaml
language: node_js
node_js:
  - "10"
  - "8"
```

Assuming you still officially support 8... which you may not, I don't know :)

ar2rsawseen:
We upgrade nodejs version together with Countly, so current release needs to support only specific nodejs version it will run on :)

ronkorving:
@ar2rsawseen Ah, so we're on 10 now? :) Cool

ar2rsawseen:
Yea, I hope so :)

PR #1035, opened 2019-08-21
[server-1509] Using localization visits for mobile apps.

frknbasaran:


ar2rsawseen:
Was fixed here:
https://github.com/Countly/countly-server/commit/c348a2301848851583da350cd228820a03f81cd6

PR #1043, opened 2019-08-29
[crashes-build-id] added right border to datatablesubrow.

frknbasaran:


plugins/crashes/frontend/public/stylesheets/main.css
```diff
@@ -256,4 +256,8 @@ td.details table tr.header, td.details table tr.header:hover {
 
 #crashgroup-manipulation-menu .item {
     user-select: none;
+}
+
+.datatablesubrow table tbody tr td:last-child {
```
ar2rsawseen: can we set some parent class/id so it is applied only to crashes table?

plugins/crashes/frontend/public/javascripts/countly.views.js
```diff
@@ -1835,7 +1835,7 @@ window.CrashgroupView = countlyView.extend({
         // `d` is the original data object for the row
         var str = '';
         if (data) {
-            str += '<div class="datatablesubrow">' +
+            str += '<div id="crash-data-subrow" class="datatablesubrow">' +
```
ar2rsawseen: There can be multiple subrows, so it is not a good idea to use `id` there, as id should identify single unique element.

Instead you could just add the id of whole table or of whole crash plugin view

PR #1048, opened 2019-09-08
gracefully skip puppeteer if not installed

riyadshauk:
I installed Countly on my own Archlinux server, and didn't want to go through the hassle of dealing with compiling/installing Puppeteer. Since I noticed that it wasn't that essential to use Puppeteer, I removed it from my own package.json. I figured this would be a nice RT check: if the user (server admin / developer) has opted out of installing Puppeteer, then simply ignore the functionality which depends on Puppeteer.

### DIFF
--- a/api/utils/render.js
+++ b/api/utils/render.js
@@ -5,7 +5,17 @@
 * @module api/utils/render
 */
 
-var puppeteer = require('puppeteer');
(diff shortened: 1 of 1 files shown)


ar2rsawseen:
Thank you, that seems to be really useful :)

PR #1049, opened 2019-09-09
[s3rver-1495] Correctly display ObjectId objects in DBViewer.

frknbasaran:


plugins/dbviewer/api/api.js
```diff
@@ -11,6 +11,24 @@ var common = require('../../../api/utils/common.js'),
...
+        * Check properties and manipulate values
+        * @param {object} doc - document
+        * @returns {object} - returns manipulated document object
+        **/
+        function objectIdCheck(doc) {
+            if (typeof doc === "string") {
+                doc = JSON.parse(doc);
+            }
+            for (var key in doc) {
+                if (doc[key] && typeof doc[key].toHexString !== "undefined") {
```
ar2rsawseen: let's check === "function" just to make sure there is no case when {"toHexStrin":true} is stored in database :D

PR #1053, opened 2019-09-11
[core] improve countly user password change experience

waiterZen:


api/parts/mgmt/users.js
```diff
@@ -259,8 +259,8 @@ usersApi.createUser = function(params) {
 
                 member[0].api_key = common.md5Hash(member[0]._id + (new Date().getTime()));
                 common.db.collection('members').update({ '_id': member[0]._id }, { $set: { api_key: member[0].api_key } }, function() { });
-
-                mail.sendToNewMember(member[0], passwordNoHash);
+                const escapedHTMLPass = passwordNoHash.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;').replace(/"/g, '&quot;');
+                mail.sendToNewMember(member[0], escapedHTMLPass);
```
ar2rsawseen: Maybe better do this inside `sentToNewMember` method, so everyone who is using it will have the fix
And also I think another method named `sendToUpdatedMember` was also using password.

PR #1055, opened 2019-09-13
[server-1551] added custom sort algorithm for app_version field.

frknbasaran:


frontend/express/public/javascripts/countly/countly.views.js
```diff
@@ -1634,7 +1634,21 @@ window.AppVersionView = countlyView.extend({
...
+                        "mData": "app_versions",
+                        "sTitle": jQuery.i18n.map["app-versions.table.app-version"],
+                        "mRender": function(data, type) {
+                            if (type === "sort") {
+                                var numbers = data.split(".");
+                                data = "";
+                                for (var i in numbers) {
+                                    data += numbers[i];
+                                }
+                                parseInt(data);
```
ar2rsawseen: Firstly parseInt data should be used somewhere
Secondly, it will incorrectly sort app_versions 2.3.5 and 2.35
we need something more sophisticated
Maybe something like this:
https://github.com/Countly/countly-server/blob/master/bin/scripts/checking_versions.js#L63-L96
Or this:
https://github.com/Countly/countly-server/blob/master/api/utils/common.js#L1653-L1708

frontend/express/public/javascripts/countly/countly.views.js
```diff
@@ -1634,7 +1634,28 @@ window.AppVersionView = countlyView.extend({
...
+                        "mRender": function(d, type) {
+                            if (type === "sort") {
+                                var numbers = d.split(".");
+                                var sortValue = 0;
+                                var multipler = 10000;
+                                for (var index in numbers) {
+                                    var valueArr = numbers[index].split("");
+                                    if (valueArr.length > 1) {
+                                        valueArr.splice(1, 0, '.');
+                                        numbers[index] = parseFloat(valueArr.join(""));
```
ar2rsawseen: How would it work with app versions like
1.b
2.1.alpha
?

frontend/express/public/javascripts/countly/countly.template.js
```diff
@@ -2650,6 +2650,55 @@ var AppRouter = Backbone.Router.extend({
...
+            }
+        
+            return 0;
+        }
+
+        jQuery.fn.dataTableExt.oSort['app_versions-asc'] = function(x, y) {
+            var versions = [];
+            versions.push(x);
+            versions.push(y);
+            return versions.sort(compareVersions);
```
ar2rsawseen: Are you sure it works as expected?
from my understanding, it should have been just
```
return compareVersions(x, y);
```

ar2rsawseen:
Also recheck, why travis builds are failing :)

frontend/express/public/javascripts/countly/countly.template.js
frknbasaran: It's works as expected now but yep your code is looks better. Just changed as yours.

PR #1058, opened 2019-09-23
[server-1562] removed cly-select search condition.

frknbasaran:
- allowed searching for value in dropdown.


ar2rsawseen:
Ok, it was a misunderstanding, so let's leave it as is with less than 10 without search. Thank you :)

PR #1064, opened 2019-10-08
[server-1550] same arrow key handle applied to dashboard selector.

frknbasaran:


ar2rsawseen:
Needs to be done from the plugin

PR #1082, opened 2019-10-17
SERVER-1513

can-keklik:


frontend/express/public/javascripts/countly/countly.helpers.js
```diff
@@ -1167,12 +1172,16 @@
             e.stopPropagation();
         });
 
-        element.off("click", ".cly-select .select-items .item").on("click", ".cly-select .select-items .item", function() {
-            var selectedItem = $(this).parents(".cly-select").find(".text");
+        element.off("click", ".cly-select .select-items .item").on("click", ".cly-select .select-items .item", function(e) {
```
waiterZen: Are you sure this selector would not affect another " .cly-select" component?

PR #1084, opened 2019-10-22
Added local module to resolve relative path from Countly root path

### LINKED ISSUE #1083: Easier way to require
This is an improvement request.

### Expected Behavior
In plugins code, I'm expecting to be able to do:
```
const countlyRoot = require('countly-root');
const myModule = require(countlyRoot('./myModules/myModule'));
```

### Current Behavior
```
// slightly exaggerating
const myModule = require('../../../../../../../../../../../../myModules/myModule');
```

The problem is two-fold:
1) Many `../`, hard to keep the count
2) Need to update the count of `../` each time files move

### Possible Solution
Have a "local" node modules that resolves to the root of Countly.

The code that resolves has to be a local node module, otherwise it would have to be required first through `require('../../../countly-root')` and breaking the point 2 described above.

This is made possible thanks to this: https://docs.npmjs.com/files/package.json#local-paths
Available since NPM 2.0.0, so this is fine.

I've already made a quick proof of concept, and it works. If you like this, please let me know and I will make a PR in no time.

TanukiSharp:
## What

This PR tries to solve the problem reported here: https://github.com/Countly/countly-server/issues/1083

## Notes

There are kind of existing soluctions, partially suitable, such as:
- https://github.com/sindresorhus/pkg-dir
- https://github.com/sindresorhus/find-up
- https://www.npmjs.com/package/app-root-path

The `pkg-dir` and `find-up` dependencies are already present in Countly, but as transitive dependencies, therefore would still require to add them as direct dependencies to implement this. I don't think there is a need to add 2 additional dependencies for that.

The `app-root-path` is close to what I want, but not solid enough in my opinion.

The implementation could be as simple as
```
function fromCountlyRoot(relativePath) {
    return path.join('/opt/countly', relativePath);
}
```
but the current one is more robust by looking for the right thing, and will work no matter containerized or running local, from anywhere.

Feel free to adjust naming and path (you may not like `local-modules`)

## Example of usage

```
const fromCountlyRoot = require('from-countly-root');
const myModule = require(fromCountlyRoot('./my/module'));

...
```

## Benefits

1. Now files can be moved for re-organization with no (or few) changes required
2. Referencing relative to a fixed point is always more logical and intuitive than relative to a moving point
3. No more crazy amount of dot-dots
4. `from-countly-root` does not suffer dot-dots, Node does the lookup for us no matter where our script is located

---

Close #1083


TanukiSharp:
As usual, many CI errors. I would like to hear your opinion about them before spending time on those issues.

Thanks.

ronkorving:
I think because NPM is guaranteed to install the local module into the root's node_modules folder (please correct me if I'm wrong), this should be enough:

```js
module.exports = (relativePath) => {
  return path.join(__dirname, '../..', relativePath);
};
```

or simply a string returning module:

```js
module.exports = path.join(__dirname, '../..');
```

mt-sebastien-robert:
Those rules of Codacy are kind of ridiculous, because anyway this module has to be completely synchronous.

If we go with @ronkorving 's suggestion, that will solve the problem. If you are fine with the original, please let me know how I can add inline comment to disable those check.

In any case, please do not wait for me to solve the Codacy issues before starting to review.

Thanks a lot.

ar2rsawseen:
Hello,
yeah not waiting on Codacy and I am probably ok with the way it is too.
Just have been preoccupied and did not yet have enough time for PR reviews, also need to make decisions about module placement, etc

But not waiting on anything from you, thank you for contribution :)

ronkorving:
Just keep in mind that this implementation depends on the "name" field in package.json being `countly-server`. If that ever changes, (as innocent as that change may seem in the future) this will break. Anything against going with the one-liner I suggest? Less moving parts is less that can go wrong, right?

ar2rsawseen:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 2
           

Complexity increasing per file
==============================
- local-modules/from-countly-root/index.js  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ar2rsawseen/countly-server/pullRequest?prid=4358141&bid=14803481)

local-modules/from-countly-root/index.js
```diff
@@ -0,0 +1,85 @@
...
+ * @param {string} root The reference path to start lookup from.
+ * @param {string} fileToSearch The name of the file to search.
+ * @returns {string} The path where the file to search as been found.
+ */
+function findReferenceFile(root, fileToSearch) {
+    var prevRoot;
+
+    do {
+        var filename = path.join(root, fileToSearch);
+        if (fs.existsSync(filename) && fs.statSync(filename).isFile()) {
```
ar2rsawseen: Codacy Issue found: [Found fs.statSync with non literal argument at index 0 (security/detect-non-literal-fs-filename)](https://app.codacy.com/manual/ar2rsawseen/countly-server/pullRequest?prid=4358141)

local-modules/from-countly-root/index.js
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/**
+ * Get the project name from a package.json file.
+ * @param {string} jsonFilename The package.json full filename.
+ * @returns {string} The name of the package (property "name" of the package.json).
+ */
+function getProjectName(jsonFilename) {
+    try {
+        return JSON.parse(fs.readFileSync(jsonFilename)).name;
```
ar2rsawseen: Codacy Issue found: [Found fs.readFileSync with non literal argument at index 0 (security/detect-non-literal-fs-filename)](https://app.codacy.com/manual/ar2rsawseen/countly-server/pullRequest?prid=4358141)

ronkorving:
@ar2rsawseen Why not go with `module.exports = path.join(__dirname, '../..');` instead? It would achieve the same with a lot less to maintain.

ar2rsawseen:
Will see, I will refactor and change location either way :)

PR #1088, opened 2019-10-30
[server-1570] hided empty sections/menus.

frknbasaran:


frontend/express/public/javascripts/countly/countly.template.js
```diff
@@ -1085,6 +1085,25 @@ var AppRouter = Backbone.Router.extend({
...
+                        }
+                    }
+                    if (isEmpty) {
+                        $(categories[j]).hide();
+                    }
+                    else {
+                        $(categories[j]).show();
+                    }
+                }
+
```
ar2rsawseen: Can we do a different logic here?
Like by default all categories are hidden but as soon as at least one member is added here:
https://github.com/Countly/countly-server/blob/master/frontend/express/public/javascripts/countly/countly.template.js#L673

The reason for this is, because menu items can be added dynamically. For example, one of the plugins is waiting for data from API to check if menu item needs to be added or not. But submenu.init would already be called by that time.

frontend/express/public/javascripts/countly/countly.template.js
```diff
@@ -579,6 +579,35 @@ $.expr[":"].contains = $.expr.createPseudo(function(arg) {
...
 });
 
+/**
+* Set menu items by restriction status, hiding empty menu-categories
+* @name setMenuItems
+* @global
+*/
+function setMenuItems() {
+    // hide empty section headers for types listed below
+    var types = ['web', 'mobile', 'desktop'], categories;
```
ar2rsawseen: We cannot hardcode this. Types can be dynamically added/removed/absent etc
I think it would be enough if it would work on the currently selected app type?
Or if not, you can use it on `this.appTypes`

PR #1093, opened 2019-11-05
[notes] change apps id query logic for notes

waiterZen:


api/parts/mgmt/users.js
```diff
@@ -811,19 +811,70 @@ usersApi.deleteUserNotes = async function(params) {
...
+            }
+        }
+        if (params.member.user_of) {
+            for (let i = 0; i < params.member.user_of.length ;i++) {
+                appIds.push(params.member.user_of[i]);
+            }
+        }
+    }
+    if (appIds.length > 0) {
+        query._id = {$in: appIds};
```
ar2rsawseen: If we already know app_ids here, do we need to make query to get them?

PR #1100, opened 2019-11-07
[core] Make app creation 1 step process

waiterZen:


api/parts/mgmt/apps.js
```diff
@@ -241,15 +241,27 @@ appsApi.createApp = function(params) {
...
+                if (err) {
+                    reject(err);
+                }
+                resolve(buf.toString('hex'));
+            });
+        });
+    }
+    catch (e) {
+        console.log(e);
+        return common.returnMessage(params, 500, "Error creating App: " + e);
```
ar2rsawseen: Can we maybe fallback to some other seed method here? Like OS generated random numbers?

PR #1137, opened 2019-12-23
its needed to be patched for no new properties 

0sus0:
because new message can leads to inje.\ct.\ion

### DIFF
--- a/frontend/express/public/javascripts/pre-login.js
+++ b/frontend/express/public/javascripts/pre-login.js
@@ -14,9 +14,9 @@
 * @memberof Pre Login
 */
 /* exported showMessage */
-function showMessage(key, prop) {
-    $("#message").data("localize", key);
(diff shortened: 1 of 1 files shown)


ar2rsawseen:
Hello,
can you provide an example of injection?
As far as I see, all loaded strings are escaped before used:
https://github.com/Countly/countly-server/blob/25f61108dacc8316a00889999aa30f5f0de72840/frontend/express/public/javascripts/pre-login.js#L75

0sus0:
yes pls chk your securty@count.ly mail i have a conversation with hello@count.ly there,
i cannot disclous it here as your policy .

ar2rsawseen:
I see,
I think this will fix it:
https://github.com/Countly/countly-server/commit/4d066388c1dfbac7de73003c95422c6598f9c35c

In overall, we need to allow passing additionally properties too to show message, so cannot accept this PR, but let's continue the conversation at the other one

PR #1138, opened 2019-12-23
pacth2 pull 1137

0sus0:
because new message can leads to inje.\ct.\ion + pacth one / 1137 / +


ar2rsawseen:
Hello,
thank you for reporting.
Do you think this would fix it enough?
https://github.com/Countly/countly-server/commit/4d066388c1dfbac7de73003c95422c6598f9c35c

0sus0:
yes just close it without changes

PR #1139, opened 2019-12-27
[server-1461] moved log creator to 'countly upgrade' command.

frknbasaran:


bin/commands/countly.sh
```diff
@@ -63,7 +63,7 @@ run_upgrade (){
                         continue
                     fi
                 fi
-                #bash "$DIR/../upgrade/$i/upgrade_fs.sh";
+                bash "$DIR/../upgrade/$i/upgrade_fs.sh" combined 2>&1 | tee -a "$DIR/../../../log/countly-upgrade-$i-$DATE.log";
```
ar2rsawseen: combined need to actually be passed only from upgrade.sh indicating that it is full upgrade run, not db or fs separately. So combined should be removed from this and passed only from upgrade.sh

PR #1140, opened 2020-01-06
[server-1660] added edited cohort name support to systemlogs plugin.

frknbasaran:


plugins/systemlogs/frontend/public/javascripts/countly.views.js
```diff
@@ -326,9 +326,19 @@ window.SystemLogsView = countlyView.extend({
...
+                        ret += "<th>" + jQuery.i18n.map["systemlogs.name"] + "</th>";
+                    }
+                    ret += "<th style='width:20%;'>" + jQuery.i18n.map["systemlogs.field"] + "</th><th style='width:40%;'>" + jQuery.i18n.map["systemlogs.before"] + "</th><th style='width:40%;'>" + jQuery.i18n.map["systemlogs.after"] + "</th>";
+                    ret += '</tr>';
                     for (var i in row.i.before) {
-                        ret += "<tr><td>" + i + "</td><td>" + self.renderField(i, row.i.before[i]) + "</td><td>" + self.renderField(i, row.i.after[i]) + "</td></tr>";
+                        ret += "<tr>";
+                        if (typeof row.i.name !== "undefined") {
+                            ret += "<td>" + row.i.name + "</td>";
+                        }
```
ar2rsawseen: Since this is done in the loop, if there are many before changes, I think this will be displayed multiple times for single log record, wouldn't it?

plugins/systemlogs/frontend/public/localization/systemlogs.properties
```diff
@@ -24,6 +24,7 @@ systemlogs.for-campaign = For Campaign
 systemlogs.for-crash = For Crash
 systemlogs.for-id = For ID
 systemlogs.data = Data
+systemlogs.name = Name
```
ar2rsawseen: I think we dont need this localization anymore

PR #1152, opened 2020-01-30
[server-1668] export bug fixed when custom path enabled.

frknbasaran:


frontend/express/public/javascripts/countly/countly.helpers.js
```diff
@@ -686,6 +686,22 @@
...
+    CountlyHelpers.getCustomPathIfExist = function() {
+        var pathString = "";
+        var urlArray = window.location.href.split("/");
+        var dashboardIndex = urlArray.indexOf("dashboard#");
+        if (dashboardIndex === 4) {
+            pathString = urlArray[3];
+        }
+        return pathString;
+    };
+
```
ar2rsawseen: I think we already had this variable: https://github.com/Countly/countly-server/blob/master/frontend/express/public/javascripts/countly/countly.config.sample.js#L13
for that purpose, we can just use `countlyCommon.API_URL` to prefix all API calls, and there is no need for custom functions :)

PR #1176, opened 2020-03-04
[server-1696] design reviews.

frknbasaran:


plugins/star-rating/frontend/public/templates/star.html
```diff
@@ -328,24 +328,24 @@
...
                             <div class="emotion tooltip-wrapper">
-                                <span class="tooltip-text">Nice</span>
-                                <img title="Nice" class="grow rating-emotion tooltip" data-score="4" src="star-rating/images/star-rating/3_gray.svg">
+                                <span style="left: -4px !important;" class="tooltip-text">Nice</span>
+                                <img class="grow rating-emotion tooltip" data-score="4" src="star-rating/images/star-rating/3_gray.svg">
                             </div>
                             <div class="emotion tooltip-wrapper">
-                                <span class="tooltip-text">Great</span>
-                                <img title="Great" class="grow rating-emotion tooltip" data-score="5" src="star-rating/images/star-rating/4_gray.svg">
+                                <span style="left: -4px !important;" class="tooltip-text">Great</span>
```
ronkorving: If you use !important here, it becomes very difficult for themes to override the look and feel of the UI. Do you need that though? Doesn't an inline style always override any pre-existing CSS?

ar2rsawseen: Ideally there should be no inline css at all, and all should go to CSS files and without important :)

PR #1179, opened 2020-03-06
Dep upgrade

ar2rsawseen:


plugins/dbviewer/api/api.js
```diff
@@ -243,18 +243,18 @@ var common = require('../../../api/utils/common.js'),
...
                                         }
-                                        done(false, true);
+                                        done();
                                     });
                                 }
                                 else {
-                                    done(false, true);
+                                    done();
                                 }
                             }, function(mapError) {
```
ronkorving: Since map became each, it's not a mapError anymore :) Nitpicking here ;)

package.json
```diff
@@ -17,19 +17,19 @@
...
     "test": "grunt --verbose"
   },
   "dependencies": {
-    "argon2": "0.24.0",
-    "async": "2.6.3",
+    "argon2": "0.26.0",
+    "async": "3.2.0",
     "basic-auth": "2.0.1",
-    "bluebird": "3.5.5",
+    "bluebird": "3.7.2",
```
ronkorving: Oh nice, you have an `allSettled` now :)
https://github.com/petkaantonov/bluebird/releases/tag/v3.7.0

bin/upgrade/dev/upgrade_fs.sh
```diff
@@ -5,13 +5,29 @@ echo "Running filesystem modifications"
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/../.." && pwd )"
 CUR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
+#upgrade nodejs
+if [ -f /etc/redhat-release ]; then
+	curl -sL https://rpm.nodesource.com/setup_10.x | bash -
```
ronkorving: Awesome :) I guess if we're officially Node 10 now, the Dockerfiles will need an update too

package.json
ar2rsawseen: Interesting, but it is not in the docs :D
http://bluebirdjs.com/docs/api-reference.html

ronkorving: > Updated 06 Nov 2018 

Those docs don't see a lot of action :-/

PR #1243, opened 2020-03-28
[feedback-ui-fixes] time series, sticker position on safari, svg alignment problems fixed.

frknbasaran:


plugins/star-rating/frontend/public/stylesheets/main.css
```diff
@@ -1,3 +1,13 @@
+.routename-star #dashboard-graph-time {
+    font-family: Oswald;
+    font-size: 14px;
+    margin: 0 30px !important;
```
ar2rsawseen: Let's not use important for CSS, but rather construct a better CSS selector

PR #1291, opened 2020-04-21
[SERVER-1783] Populator templates fixes

LVMBDV:
This fixes both SERVER-1783 and SERVER-1784.


LVMBDV:
DeepScan is wrong on this one, `!_.isObject` is a null check :^)

PR #1299, opened 2020-04-23
[zoom-fixes] on feedback and push charts.

frknbasaran:


plugins/push/frontend/public/javascripts/countly.views.js
```diff
@@ -351,4 +351,9 @@ $(document).ready(function() {
             }
         });
     }
+
+    $('body').on('click', '.tab-buttons > div', function() {
+        if ($(this).next().length === 1) $('.widget-content').addClass('hide-zoom');
+        else $('.widget-content').removeClass('hide-zoom');
+    });
```
ar2rsawseen, line 358: Eslint complaint

plugins/push/frontend/public/javascripts/countly.views.js
```diff
@@ -351,4 +351,13 @@ $(document).ready(function() {
             }
         });
     }
+
+    $('body').on('click', '.tab-buttons > div', function() {
```
ar2rsawseen, line 355: It would be a good practice to do `.on('click', '.tab-buttons > div').on('click', '.tab-buttons > div', function() {`

ar2rsawseen: Also, the selector is too broad, it will work on other cases outside push with the same classes too

PR #1324, opened 2020-05-02
[SERVER-1694] MongoDB installation and check script

kaanklky:
Completed tasks
- Using the XFS filesystem is strongly recommended with the WiredTiger storage engine
    - Not formatting disk by script
- Turn off atime for the storage volume containing the database files.
- Setup log rotation
- Setting swappiness to 1
- NUMA hardware check and configs
- Ensure NTP is installed
    - Checking but not installing NTP, `check` function of this script doesn't require internet connection to use in any condition
- Configure sufficient file handles
- Linux Kernel and Glibc Version

Denied because of virtualization
- Check it has SSD
- Set the readahead setting between 8 and 32 regardless of storage media type (spinning disk, SSD, etc.).


ar2rsawseen:
> I'm afraid that won't work. Pushing mongo install up would do I guess, right?

Which line are you talking about exactly?

bin/scripts/mongodb.install.sh
```diff
@@ -2,76 +2,265 @@
...
+
+    if [ -f /etc/lsb-release ]; then
+        if [[ "$(/sbin/init --version)" =~ upstart ]]; then
+            restart mongod || echo "mongodb upstart job does not exist"
+        else
+            systemctl restart mongod || echo "mongodb systemctl job does not exist"
+        fi
+    fi
+
+    bash $0 check
```
ar2rsawseen, line 87: In /home/travis/build/Countly/countly-server/bin/commands/scripts/../../../bin/scripts/mongodb.install.sh line 87:
    bash $0 check
         ^-- SC2086: Double quote to prevent globbing and word splitting.
Did you mean: 
    bash "$0" check

iartem:
@kaanklky you've commented out GCC exports for rhel install script, push compilation will most likely fail in this case.

iartem:
As a suggestion, instead of commenting things out, just move mongo install script before the lines you've commented out.

iartem:
Ok, @ar2rsawseen told me that those lines were invalid. Then in case it does work, that is no errors during push compilation, then I have no more objections.

PR #1332, opened 2020-05-07
[heatmap-pages] added 5 different heatmap page types.

frknbasaran:


plugins/populator/frontend/public/javascripts/countly.models.js
```diff
@@ -542,10 +542,11 @@
 
             if (id === "[CLY]_view") {
                 event.segmentation = {};
+                var populatorType = $(".populator-template-name.cly-select").clySelectGetSelection().substr(7).toLowerCase();
                 Object.keys(viewSegments).forEach(function(key) {
                     var values = [];
                     if (countlyGlobal.apps[countlyCommon.ACTIVE_APP_ID].type === "web" && key === "name") {
-                        values = ["/" + countlyCommon.ACTIVE_APP_KEY + "/demo-page.html"];
+                        values = ["/" + countlyCommon.ACTIVE_APP_KEY + "/demo-" + populatorType + ".html"];
```
ar2rsawseen, line 549: What will happen if custom template is used?

frontend/express/public/stylesheets/heatmap-demos/navigation.css
```diff
@@ -0,0 +1,39 @@
+.top-bar {
```
ar2rsawseen, line 1: Can we move all these files into populator plugin folder rather than core?

frontend/express/public/stylesheets/heatmap-demos/navigation.css
```diff
@@ -0,0 +1,39 @@
...
+  height: 60px;
+  background-color: #13B94D;
+  color: white;
+  font-size: 16px;
+  border-bottom-left-radius: 5px;
+  border-bottom-right-radius: 5px;
+}
```
ar2rsawseen, line 8: Can you recheck Codacy complaining about these styles?

plugins/populator/frontend/app.js
```diff
@@ -5,8 +5,13 @@ var exported = {},
 
 (function(plugin) {
     plugin.init = function(app) {
-        app.get(countlyConfig.path + '/:id/demo-page.html', function(req, res) {
-            var pageSourceCode = fs.readFileSync(path.resolve(__dirname, 'public/templates/demo-page.html'));
+        const predefinedTypes = ["demo-banking.html", "demo-gaming.html", "demo-navigation.html", "demo-healthcare.html", "demo-ecommerce.html"];
+        app.get(countlyConfig.path + '/:id/:page', function(req, res) {
```
ar2rsawseen, line 9: I think this path pattern breaks tests
that is why Travis cannot pass the tests.
I think we cannot use any substring pattern there, but there must be some unique prefix, like 
```
app.get(countlyConfig.path + '/populator/:id/:page', function(req, res) {
```
Also, can you recheck the warnings on LGTM about escaping?

PR #1343, opened 2020-05-18
[SERVER-1251] Streaming data exports to CSV and excel files

Cookiezaurs:
[SERVER-1251] Streaming data exports to CSV and excel files

[SERVER-1251]: https://countly.atlassian.net/browse/SERVER-1251

### DIFF
--- a/api/parts/data/exports.js
+++ b/api/parts/data/exports.js
@@ -337,7 +337,7 @@ exports.fromDatabase = function(options) {
     options.query = options.query || {};
     options.projection = options.projection || {};
     if (options.limit && options.limit !== "") {
-        options.limit = parseInt(options.limit);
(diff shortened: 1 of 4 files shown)


api/parts/data/exports.js
```diff
@@ -132,45 +169,151 @@ exports.output = function(params, data, filename, type) {
...
-        });
+            stream.once('close', function() {
+                if (listAtEnd) {
+                    sheet.write(paramList);
+                }
+                sheet.end();
+                xc.finalize();
+            });
+        }
+        else {
```
can-keklik, line 299: Just out of curiosity, is the purpose of this `else {` branch to handle JSON?

Cookiezaurs: yes

PR #1351, opened 2020-05-21
[SERVER-1810] users be able to specify a custom value like this: 1, 2…

frknbasaran:
…, 3, 4, 5.


plugins/populator/frontend/public/javascripts/countly.views.js
```diff
@@ -378,7 +378,7 @@ window.PopulatorView = countlyView.extend({
             if (["true", "false"].indexOf(s) !== -1) {
                 return s === "true";
             }
-            else if (/^[1-9][0-9]+|0$/.test(s)) {
+            else if (/^[1-9][0-9]+|0|[1-9]$/.test(s)) {
```
ronkorving, line 381: Replacing `+` (1 or more) with `*` (0 or more) would've also done the trick.

PR #1380, opened 2020-06-11
[feature/crud-context] new permission mechanism implementation.

frknbasaran:


bin/scripts/member_permission_generator.js
```diff
@@ -0,0 +1,50 @@
...
+    if (!members && err) {
+        console.log(err);
+        countlyDb.close();
+        return;
+    }
+
+    function upgrade(member, done) {
+        if (!member.admin_of || !member.user_of) {
+            done();
+        }
```
mt-sebastien-robert, line 15: You need a `return` statement here.

mt-sebastien-robert, line 15: You need a `return` statement here.

bin/scripts/member_permission_generator.js
```diff
@@ -0,0 +1,50 @@
...
+            }
+            done();
+        });
+    };
+
+    async.forEach(members, upgrade, function() {
+        console.log("Finished upgrading member permissions.");
+        countlyDb.close();
+    });
+});
```
mt-sebastien-robert, line 50: Linefeed missing ?

ar2rsawseen:
Maybe we can move it to the upgrade script path directly, for next release
bin/upgrade/dev/scripts/member_permission_generator.js

PR #1386, opened 2020-06-17
[Server-1813]Events overview links

waiterZen:


frontend/express/public/javascripts/countly/countly.views.js
```diff
@@ -4872,6 +4872,13 @@ window.EventsOverviewView = countlyView.extend({
             self.refresh(true, true);
             $("#update_overview_button").removeClass('disabled');
         });
+        $(".top-events-widget .col").off("click").on("click", function(e) {
+            window.location.href = $(e.currentTarget).children("h4").children("a").attr("href");
+        });
+        $(".event-overview-grid-block").off("click").on("click", function(e) {
+            window.location.href = $(e.currentTarget).children(".title").children("a").attr("href");
+        });
+
```
ar2rsawseen, line 4881: If they are really anchor elements with href attribute, do they need this? they should automatically change the URL on click, shouldn't they?

waiterZen: Because we want to support card block level link, if embered <a> tag with whole card dom is more complex for css control, so got this approach.

PR #1391, opened 2020-06-23
[comments-tab-fix] fixed comments tab sorting problem.

frknbasaran:


plugins/star-rating/frontend/public/javascripts/countly.views.js
```diff
@@ -856,7 +857,8 @@ window.starView = countlyView.extend({
                     }
                 },
                 sType: "string",
-                "sTitle": jQuery.i18n.map["management-users.email"]
+                "sTitle": jQuery.i18n.map["management-users.email"],
+                "bSortable": false
```
ar2rsawseen, line 861: Why do you disable sorting on these columns if you just enabled sorting on the server side?

frknbasaran: when I first create this comments tab, I didn't know how to use legacy version of this datatable efficiently. I didn't know bSortable property, so I should already did these columns non-sortable when I was create this table first time. Why? Because sorting data by comment and email is weird way to sort data. These columns are not a logically sortable fields, so it just sorting text by ascii code and it's meanless for customer I guess. I allowed rating and date columns for sorting on server-side but if you want to add sortability to email and comments columns again, I can :D

PR #1396, opened 2020-06-25
Bugfix. Don't look for tepmplate if path not set.

Cookiezaurs:


frontend/express/public/javascripts/countly/countly.template.js
```diff
@@ -486,9 +486,14 @@ window.countlyManagementView = countlyView.extend({
 
     beforeRender: function() {
         var self = this;
-        return $.when(T.render(this.templatePath, function(src) {
-            self.template = src;
-        }));
+        if (this.templatePath && this.templatePath !== "") {
```
ronkorving, line 489: If templatePath is truthy, it can't be empty string, so this would behave identical:

```js
if (this.templatePath) {
```

ar2rsawseen: better being thorough than sorry? :D

iartem: At least someone does actually review :D

ronkorving: @ar2rsawseen The `!== ""` test is a no-op, so it's not really "thorough" IMHO :)

ar2rsawseen: Operation wise, sure
but it is more readable, as I can understand type and requirements by looking just on that case

iartem: Honestly, that thing makes it way less readable

PR #1401, opened 2020-06-30
[datatable-sort-fix] fixed comments table sort problem.

frknbasaran:


plugins/star-rating/frontend/public/javascripts/countly.views.js
```diff
@@ -1179,6 +1179,16 @@ window.starView = countlyView.extend({
...
+        $('body').on('click', '.dataTable .sorting, .dataTable .sorting_asc, .dataTable .sorting_desc', function(e) {
+            var dtSelector = $(e.target).parent().parent().parent().parent().attr('id').split("_")[0];
+            var sortIndex = $(e.target).index();
+            var sortType = e.target.className === 'sorting' ? 'asc' : (e.target.className === 'sorting_asc' ? 'desc' : false);
+            if (sortType) {
+                var sortObject = [sortIndex, sortType, (sortType === "asc" ? 0 : 1)];
+                store.set(dtSelector + '_sort', sortObject);
+            }
+        });
+
```
ar2rsawseen, line 1191: Do we really need to persistently remember last sorted column?

If yes and Antonin confirms it, then this looks like something that would need to be global for all tables, not just feedback?
Possibly somewhere here, where we already have the object for persistent table settings:
https://github.com/Countly/countly-server/blob/master/frontend/express/public/javascripts/countly/countly.template.js#L3009

PR #1413, opened 2020-07-13
[server-1863] added 20 constant ip addresses.

frknbasaran:


plugins/populator/frontend/public/javascripts/countly.models.js
```diff
@@ -298,12 +299,7 @@
         this.isRegistered = false;
 
         this.hasSession = false;
-        if (ip_address.length > 0 && Math.random() >= 0.5) {
-            this.ip = ip_address.pop();
-        }
-        else {
-            this.ip = chance.ip();
-        }
+        this.ip = predefined_ip_addresses[Math.floor(chance.random() * 20)];
```
ar2rsawseen, line 302: Can we use array length - 1 in this case instead of hardcoded 20
so in case we modify array (like adding or removing more IP addresses) we won't have to modify other places too :)

PR #1416, opened 2020-07-16
New Lint rules

ar2rsawseen:


ronkorving:
Nice :)

.stylelintrc.json
```diff
@@ -0,0 +1,27 @@
...
+    "color-hex-length": "short",
+    "unit-case": "lower",
+    "number-leading-zero": "always",
+    "value-keyword-case": "lower",
+    "declaration-block-single-line-max-declarations": 1,
+    "no-descending-specificity": "off",
+    "selector-list-comma-newline-after": "never-multi-line",
+    "selector-list-comma-newline-before": "never-multi-line",
+    "selector-list-comma-space-after": "always-single-line",
+    "selector-list-comma-space-after": "never-single-line",
```
ar2rsawseen, line 16: Codacy found an issue: [Duplicate field 'selector-list-comma-space-after'](https://app.codacy.com/manual/ar2rsawseen/countly-server/pullRequest?prid=5864534)

ar2rsawseen:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 2
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ar2rsawseen/countly-server/pullRequest?prid=5864534&bid=19196795)

.eslintrc.json
```diff
@@ -234,7 +264,13 @@
                     "off"
                 ],
                 "no-await-in-loop": [
-                    "off"
+                    "error"
```
iartem, line 267: Too limiting. There're good examples when one would want a sequential execution of promises instead of parallel one for performance reasons, for example. No await in `while` or `for` doesn't bring any advantage except for very few very obvious cases which I hope we're all capable of avoiding.

.eslintrc.json
```diff
@@ -234,7 +264,13 @@
...
+                ],
+                "no-return-await": [
+                    "error"
```
iartem, line 270: Again, too limiting without any good advantages. For example:
```
async function findRows () {
    let rows = await db.find(),
         transformed = rows.map(x => x * 2);
    
    return await this.calculate(transformed);
}
```
A decent case. I don't see any reasons why that's bad.

.eslintrc.json
```diff
@@ -234,7 +264,13 @@
...
+                ],
+                "no-promise-executor-return": [
+                    "error"
```
iartem, line 273: This one I'm not that sure about. As I understand it'd prevent returning a promise from executor function, right? In that case I'd be against this one as well because I see that practice very useful. Yet I guess it's somewhat a bad practice, not sure. What do you think?

.eslintrc.json
```diff
@@ -206,6 +210,32 @@
...
+                    "version": ">=10.0.0"
+                }],
+                "node/no-unsupported-features/node-builtins": ["error", {
+                    "version": ">=10.0.0"
+                }],
+                "no-process-exit": "off",
+                "node/callback-return": "error",
+                "node/exports-style": "error",
+                "node/no-sync": "error",
+                "promise/no-promise-in-callback": "off",
```
iartem, line 234: Not sure about this one. I think it'd be pretty useful to force these limitations on us. Like if `plugins.register()` argument is not a promise, then don't use promises in it, make the callback a promise instead. 

Second one is questionable as it'd make `.map()` in an `async` function impossible. What do you think?

.eslintrc.json
```diff
@@ -234,7 +264,13 @@
...
                 "no-await-in-loop": [
-                    "off"
+                    "error"
+                ],
+                "no-return-await": [
+                    "error"
+                ],
+                "no-promise-executor-return": [
+                    "error"
                 ],
```
iartem, line 274: For the [promise](https://cnpmjs.org/package/eslint-plugin-promise) plugin, I'd set following:

```
{
  "rules": {
    "promise/always-return": "error",
    "promise/no-return-wrap": "off",
    "promise/param-names": "error",
    "promise/catch-or-return": "error",
    "promise/no-native": "off",
    "promise/no-nesting": "off",
    "promise/no-promise-in-callback": "warn" / "error",
    "promise/no-callback-in-promise": "off",
    "promise/avoid-new": "off",
    "promise/no-new-statics": "error",
    "promise/no-return-in-finally": "warn",
    "promise/valid-params": "warn"
  }
}
```

.eslintrc.json
ar2rsawseen: I would say forcing tasks in parallel is a good idea and sequential can be treated as an exception, would make you think twice do you really need it as a sequential process or not

.eslintrc.json
prikshittekta: This would prevent returning a value from the promise. (not the promise)
Rule explanation says - "The return value of the executor is ignored. Returning a value from an executor function is a possible error because the returned value cannot be used and it doesn't affect the promise in any way"
Which makes sense, so I have no problem with this rule.

ar2rsawseen: I guess main point here is this
```
new Promise((resolve, reject) => {
    if (someCondition) {
        return defaultResult;
    }
    getSomething((err, result) => {
        if (err) {
            reject(err);
        } else {
            resolve(result);
        }
    });
});
```
if you do return default value, promise is never handled at all (nor resolves and nor rejected)
same here:
```
new Promise(() => {
    return 1;
});
```
So the rule enforces that promise is always resolved or rejected

ar2rsawseen: But I guess javascript garbage collector would be smart enough to collect pending promises once they have no references, right?

.eslintrc.json
ar2rsawseen: Agree about map and async, but can you provide examples what you mean about `plugins.register`?

.eslintrc.json
can-keklik: Agree with Artem. Any examples?

ar2rsawseen: I agree that it would be easier for me to then use the result of that findRows function rather than dealing with another promise :D

But as rule suggests, if returning await, then this current function is kept in stack until the promise resolves which kind of pollutes the stack and has a minor performance hit (https://eslint.org/docs/rules/no-return-await)

But I could not find any hard evidence of that and benchmarks that would prove performance hit. 
So I guess we can turn that off for the sake of usability

.eslintrc.json
can-keklik: Even though the garbage collector handles it, it would be a nice to have standard for codebases.

.eslintrc.json
ar2rsawseen: Oh, actually
```
Since the return value of an async function is always wrapped in Promise.resolve, return await doesn’t actually do anything except add extra time before the overarching Promise resolves or rejects.
```

ar2rsawseen: http://hassansin.github.io/Why-return-await-Is-a-Bad-Idea

.eslintrc.json
can-keklik: I am 50/50. Sequential execution can be necessary sometimes. But it is vulnerable to misuse for example sequential db calls (I think this is what Artem means while saying very obvious cases).

.eslintrc.json
iartem: Oh, how didn't I think about that. Then yes, let's keep it.

.eslintrc.json
ar2rsawseen: yes, but eslint idea is to prevent possible errors, and using this rule would prevent of processing promises in sequence unintentionally

.eslintrc.json
iartem: So it's not about returning a promise in a `.then()`. Then sure, why not.

prikshittekta: A somewhat similar implementation can be seen here - https://github.com/Countly/countly-server/blob/bbf1aa88676b4024ff89dc37069b0ec987c8034c/api/utils/requestProcessor.js#L2425
The promise is called recursively, and once the most recent promise is resolved, the others behind it are neither resolved nor rejected.

.eslintrc.json
iartem: Ok, I want this to be [forbidden](https://github.com/xjamundx/eslint-plugin-promise/blob/development/rules/lib/is-promise.js#L12-L14):
```
plugins.register('/i/whatever', function (params){
    whatever.doSomething().then(smth => {
        common.returnOutput(params, smth);
    });
});
```

So we'd have to do this:
```
plugins.register('/i/whatever', async params => {
    let smth = await whatever.doSomething();
    common.returnOutput(params, smth);
});
```

... because of callback-promise incompatibility due to which we don't have a centralised error-catching & error-reporting. Basically I'd like any plugin dispatch error to result in `common.returnMessage` in `pluginManager.js`, not in individual plugins.

Yet this rule would possibly result in lots of changes. But it's must IMO in case we're really up to the task to move to async/await.

ar2rsawseen:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 2
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ar2rsawseen/countly-server/pullRequest?prid=5864534&bid=19196795)

.eslintrc.json
ar2rsawseen: but not all dispatches should have output to the browser, because they can just be internal events?

.eslintrc.json
iartem: Oh, good point BTW. It'd prevent returning from the executor function scope, right? 

```
new Promise((res, rej) => {
    if (something.notExists()) {
        return res(); // <<<<< error
    }

    res('something exists');
});
````

Comment (prikshittekta) [reply .eslintrc.json:273]: yes

File: .eslintrc.json (hunk 2 above)
Comment (iartem) [reply .eslintrc.json:267]: I think our CTO should fire the person who commit that effective immediately :D

File: .eslintrc.json (hunk 5 above)
Comment (ar2rsawseen) [reply .eslintrc.json:234]: So adding this rule adds
```
76 problems (76 errors, 0 warnings)
```
which is manageable, so adding it

File: .eslintrc.json (hunk 6 above)
Comment (ar2rsawseen) [reply .eslintrc.json:274]: I am not sure if warn is useful, as most of checks are run on CI pipelines they won't be seen, and only errors are actionable.
So I would vote to put both last 2 options as error

File: .eslintrc.json (hunk 2 above)
Comment (ar2rsawseen) [reply .eslintrc.json:267]: everyone can make mistakes :)

File: .eslintrc.json (hunk 4 above)
Comment (iartem) [reply .eslintrc.json:273]: Then I'm back against this :)

File: .eslintrc.json (hunk 5 above)
Comment (iartem) [reply .eslintrc.json:234]: > but not all dispatches should have output to the browser, because they can just be internal events?

Sure, let `pluginManger` throw error from `dispatch` which would be handled in `requestProcessor` if needed for a particular request.

File: .eslintrc.json (hunk 4 above)
Comment (ar2rsawseen) [reply .eslintrc.json:273]: It does not prevent return from executor scope in this case (as flow control), just make sure all routes resolve or reject
```
new Promise((resolve, reject) => {
    if (someCondition) {
        resolve(defaultResult);
        return;
    }
    getSomething((err, result) => {
        if (err) {
            reject(err);
        } else {
            resolve(result);
        }
    });
});
```

File: .eslintrc.json (hunk 5 above)
Comment (ar2rsawseen) [reply .eslintrc.json:234]: still not sure how that will work, but I am ok for this rule

File: .eslintrc.json (hunk 4 above)
Comment (prikshittekta) [reply .eslintrc.json:273]: As per my understanding, this rule won't check if we are resolving a promise or not.
But rather check for the returns from the promise.
The rule title says - Disallow returning values from Promise executor functions
So it will throw an error if a value is returned from within the promise.
You can still return but without a value bcz a returned value from promise is meaningless.
.then just gets the resolved or rejected values afterall.

Comment (ar2rsawseen) [reply .eslintrc.json:273]: true, so it won't guard against cases of unhandled promises
I guess then we can turn it off

File: .eslintrc.json (hunk 2 above)
Comment (ar2rsawseen) [reply .eslintrc.json:267]: So @iartem this is the last one, will you budge on this or you stand your position? :D

File: .eslintrc.json (hunk 4 above)
Comment (ar2rsawseen) [reply .eslintrc.json:273]: reiterating on this actually :D
so does returning value from executor even make sense? it would not be used anywhere, right, only resolved value is passed returned outside of promise. So it actually prevents you from accidentally thinking you are returning value but you don't

And for @iartem you simply need to do this:
```
new Promise((res, rej) => {
    if (something.notExists()) {
        res();
        return;
    }

    res('something exists');
});
```

File: .eslintrc.json (hunk 2 above)
Comment (iartem) [reply .eslintrc.json:267]: I just don't like meaningless limitations, but I won't be too loud about this one :D

**Comment by ar2rsawseen** [comment]
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 2
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ar2rsawseen/countly-server/pullRequest?prid=5864534&bid=19196795)

ar2rsawseen:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ar2rsawseen/countly-server/pullRequest?prid=5864534&bid=19196795)

ar2rsawseen:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ar2rsawseen/countly-server/pullRequest?prid=5864534&bid=19196795)

PR #1583, opened 2020-10-22
[SERVER-1940] renamed configurations to settings

LVMBDV:


frontend/express/public/localization/help/help.properties
```diff
@@ -64,7 +64,7 @@ help.manage-apps.app-clear-button = Deletes all the data of this application but
 help.manage-apps.app-edit-button = Activates editing mode for this application. Users with a global administrator account and administrators of this application can perform this action.
 help.manage-apps.app-lock-button = Locks this application to prevent accidental delete/clear/reset. Users with a global administrator account and administrators of this application can perform this action.
 help.manage-apps.checksum-salt = Providing this value will enable checking request's checksum to prevent parameter tampering. You must configure same salt on SDK side too.
-help.manage-apps.app-clear-button-hint = Purge all data or data older than a selected date, but keep the configurations
+help.manage-apps.app-clear-button-hint = Purge all data or data older than a selected date, but keep the settings
```
ar2rsawseen, line 67: So these 2 (including the one in dashbaord.properties file) are actually referencing what we currently will call configurations (or basically data that config trasnfer will transfer), so I think their names should still be configurations :)

ar2rsawseen (changes requested):

frontend/express/public/localization/help/help.properties
LVMBDV: Yup, you're right, I was a bit too eager to rename all mentions of configurations lol.

PR #1589, opened 2020-10-25
[SERVER-1913] feedback submenu

prikshittekta:


frontend/express/public/javascripts/countly/countly.template.js
```diff
@@ -1347,6 +1347,9 @@ var AppRouter = Backbone.Router.extend({
             if (countlyGlobal.member.global_admin || (countlyGlobal.admin_apps && Object.keys(countlyGlobal.admin_apps).length)) {
                 self.addSubMenu("events", {code: "manage-events", url: "#/analytics/events/blueprint", text: "sidebar.events.blueprint", priority: 100});
             }
+
+            app.addMenu("reach", {code: "feedback", text: "sidebar.feedback", icon: '<div class="logo ion-android-star-half"></div>', priority: 20});
+
```
ar2rsawseen, line 1352: So we cannot add this from the core, we need to add it from both CE and EE plugins
so in both plugins we need to check if #feedback-menu exists and if not, then add it

ar2rsawseen (changes requested):

frontend/express/public/javascripts/countly/countly.template.js
prikshittekta: okay.
Is there a common utility to check if the menu item exists or not?

PR #1593, opened 2020-10-26
Fix Link to Docker Hub Image

SethiPandi:
Hey! I was just trying to get the project running and reading some docs, so I could tackle the initial exercise later.
It seems the link to the Docker image is incorrect though, it should be `/r/`, not `/u/`. :thinking: 

This PR just corrects that so users get taken to the page they expect.


SethiPandi:
(Don't mind that, clicked the wrong thing. ^-^')

ar2rsawseen:
Hello,
thank you for noticing this and fixing :)

PR #1603, opened 2020-11-01
[SERVER-1807] widgets orchestration in dashbaords

prikshittekta:


prikshittekta:
TODO for future - Move all dashboard widgets related code into separate js files.

PR #1714, opened 2021-01-26
[SERVER-2089] - analytics data inconsistency fixes

itsiprikshit:


api/lib/countly.common.js
```diff
@@ -1030,19 +1036,15 @@ countlyCommon.extractBarData = function(db, rangeArray, clearFunction, fetchFunc
...
-    for (let i = maxItems - 1; i >= 0; i--) {
-        var percent = Math.floor((rangeTotal[i] / sum) * 100);
+    for (var i = 0; i < rangeNames.length; i++) {
+        var percent = countlyCommon.round((rangeTotal[i] / sum) * 100, 0);
         totalPercent += percent;
 
-        if (i === 0) {
+        if (i === (rangeNames.length - 1)) {
             percent += 100 - totalPercent;
         }
```
ar2rsawseen, line 1049: Hello
previously we went through values backwards from less to top and added the accumulated difference to 1st top value
We did this, because there were cases when accumulated percentage difference was quite big and if we added it to last value, then last value would have more percentage than previous value. That is why order was reversed.

So just checking if some situation is possible here?

Example:
Platform top
Android: 97.3%
iOS: 1.4%
Windows: 1.3%

Then if we add rounding error to the last element, then result would be:
Android: 97%
iOS: 1%
Windows: 2%

Which would be incorrect

So just want to make sure this situation is handled or not possible with new approach :)

itsiprikshit: Hello this can still happen at places where we don't sort the array based on percentage after calculating the percentages.
We don't sort the array everywhere after calculating percentages.
Why I am calculating percentages from the top (and not end like before) because in pie charts and horizontal charts, all the initial values are rounded off and the last value is called "Other" which is calculated by 100 - totalPercent.
So to maintain the consistency, i started calculated percentages from top by rounding off.
To fix the case you suggested, i can sort the arrays based on percentages whereever we are not sorting them.
I found one such place in the code where we don't. All other does the sorting.
Sounds ok?

ar2rsawseen: Yep, sounds great :)

ar2rsawseen: Sorry, maybe I misunderstood
but in piecharts, it makes sense to round off in last because of the "Others" option
but in tops we should always round off in the first to avoid the provided scenario, in all the places, I think

PR #1728, opened 2021-02-05
report email view permisson & unsubscribe support

waiterZen:


frontend/express/app.js
```diff
@@ -1048,11 +1048,30 @@ Promise.all([plugins.dbConnection(countlyConfig), plugins.dbConnection("countly_
...
+            countlyDb.collection('reports').update({_id: common.db.ObjectID(reportID)}, { $pull: {'emails': email}},function(err, result) {
+                // silence procedure
+                res.redirect(countlyConfig.path + '/login?message=report.unsubscribe-successful&no_redirect=true');
+            });
+        } catch (e) {
+            console.log(e);
+            res.redirect(countlyConfig.path + '/login?message=report.unsubscribe-failed&no_redirect=true');
+        }
+    });
+
```
ar2rsawseen, line 1069: I think this can actually go into reports plugin's frontend js?

frontend/express/public/localization/pre-login/pre-login.properties
```diff
@@ -15,6 +15,10 @@ login.locked = Your account was locked by admin
 login.button = Login
 login.token-expired = Your session was expired. Please login again
 
+#report
+report.unsubscribe-successful = You have successfully unsubscribed from this email report.
+report.unsubscribe-failed = Email report unsubscribe failed.
```
ar2rsawseen, line 20: Can you recheck this text with Mariano?

plugins/reports/api/api.js
```diff
@@ -166,11 +179,16 @@ var common = require('../../../api/utils/common.js'),
                         return common.returnMessage(params, 401, 'User does not have right to access this information');
                     }
 
-                    common.db.collection('reports').findOne({_id: common.db.ObjectID(id), user: common.db.ObjectID(params.member._id)}, function(err_update, report) {
+                    const query = {_id: common.db.ObjectID(id)};
+                    if (params.member.global_admin !== true) {
+                        query.user = common.db.ObjectID(params.member._id); 
+                    }
```
ar2rsawseen, line 185: I guess we don't need this piece of code, as it is in `recordUpdateOrDeleteQuery` function

frontend/express/app.js
```diff
@@ -1048,11 +1048,30 @@ Promise.all([plugins.dbConnection(countlyConfig), plugins.dbConnection("countly_
...
     app.get(countlyConfig.path + '/login', function(req, res) {
         res.header('Cache-Control', 'no-cache, private, no-store, must-revalidate, max-stale=0, post-check=0, pre-check=0');
         res.header('Expires', '0');
         res.header('Pragma', 'no-cache');
-        if (req.session.uid) {
+        if (req.session.uid && !req.query.no_redirect) {
```
ar2rsawseen, line 1074: I don't think forcing Login screen to show message is good idea UX wise, you should consult Antonin maybe on proper approach here

plugins/reports/api/reports.js
```diff
@@ -577,15 +599,62 @@ var metricProps = {
...
-                    html: message
+                    // if report contains customize message for each email address, use reports.messages[i]
+                    html: report.messages && report.messages[i] && report.messages[i].html || message,
                 };
+
+                if (report.messages && report.messages[i]) {
+                    msg.list = {
+                        unsubscribe: {
+                            url: report.messages[i].unsubscribeLink, 
+                            comment: 'Countly Report Unsubscribe'
```
ar2rsawseen, line 653: I think we need this to be localized too, and it can be simply "Unsubscribe"

frontend/express/public/localization/pre-login/pre-login.properties
ar2rsawseen: And would really want to see them inside plugin instead of core.
You can load external localization files in prelogin, like this: https://github.com/Countly/countly-server/blob/master/plugins/enterpriseinfo/frontend/public/templates/login.html#L82

plugins/reports/api/reports.js
```diff
@@ -577,15 +599,62 @@ var metricProps = {
...
                     to: report.emails[i],
                     from: versionInfo.title,
                     subject: report.subject,
-                    html: message
+                    // if report contains customize message for each email address, use reports.messages[i]
+                    html: report.messages && report.messages[i] && report.messages[i].html || message,
                 };
+
+                if (report.messages && report.messages[i]) {
+                    msg.list = {
```
ar2rsawseen, line 650: And can we add that `List-Unsubscribe` header with link too? I know it won't work in all cases, but should be a nice addition, for future.
It will mean modifying mail.js to accept custom headers too.

ar2rsawseen (changes requested):

plugins/reports/api/reports.js
waiterZen: Yes, we already add  "List-Unsubscribe" header  to nodemail lib, same example here: https://nodemailer.com/message/list-headers/

plugins/reports/api/reports.js
waiterZen: done

plugins/reports/api/api.js
waiterZen: done

plugins/reports/api/reports.js
ar2rsawseen: Where do we add it?

PR #1733, opened 2021-02-08
[core][crashes] update for hooks data

waiterZen:


plugins/crashes/api/api.js
```diff
@@ -480,6 +480,12 @@ plugins.setConfigs("crashes", {
                                 }];
                                 plugins.dispatch("/plugins/drill", {params: params, dbAppUser: dbAppUser, events: events});
```
ar2rsawseen, line 481: Hello, instead of having a separate event for incoming data, maybe hooks can reuse this event that we already have and all future manual events will use it to write data to drill either way :)

PR #1762, opened 2021-03-11
[feature/crud-context] CRUD based user management.

frknbasaran:
Current commits are only contain core changes. 
Other commits for community plugins on the way.


api/utils/requestProcessor.js
```diff
@@ -273,12 +273,12 @@ const processRequest = (params) => {
                 }
                 switch (paths[3]) {
                 case 'save':
-                    validateUserForWriteAPI(params, () => {
+                    validateCreate(params, 'core', () => {
```
ar2rsawseen, line 276: I think this works both as create and update, right? so we need to decide what to do in those situations
but my guess we would go with update right

api/utils/requestProcessor.js
```diff
@@ -537,27 +537,21 @@ const processRequest = (params) => {
...
                 switch (paths[3]) {
                 case 'create':
-                    validateUserForWriteAPI(() => {
-                        if (!(params.member.global_admin)) {
-                            common.returnMessage(params, 401, 'User is not a global administrator');
-                            return false;
-                        }
-                        countlyApi.mgmt.apps.createApp(params);
-                    }, params);
+                    validateUserForGlobalAdmin(params, countlyApi.mgmt.apps.createApp);
```
ar2rsawseen, line 540: This would not be Create for "global_applications" too?

extend/mail.example.js
```diff
@@ -1,85 +0,0 @@
-//file should be placed in countly/extend
```
ar2rsawseen, line 1: we should put this file back :)

bin/upgrade/dev/scripts/member_permission_generator.js
```diff
@@ -0,0 +1,52 @@
+var pluginManager = require('../../../plugins/pluginManager.js'),
+    async = require('async'),
+    countlyDb = pluginManager.dbConnection();
```
ar2rsawseen, line 3: Now this method returns promise and not db object, since we got rid of mongoskin

package-lock.json
```diff
@@ -0,0 +1,8351 @@
+{
```
ar2rsawseen, line 1: Can we remove this file?

ar2rsawseen, line 1: Can we remove this file?

ar2rsawseen (changes requested):

api/utils/requestProcessor.js
frknbasaran: yes, nice catch. but if we go with "update" rights control in here, we can't allow users who have only "create" rights, so I think we need a custom solution here. maybe we can add some dependency rule for "core" permission that force user to give both "update" and "create" permission on the frontend. or custom permission check method like validateCreateOrUpdate (). but the first solution sounds better to me.

I'm leaving this part as it is, and I took notes for the new-UI permission box about this.

api/utils/requestProcessor.js
frknbasaran: yes, it should be and changed already. :)

ar2rsawseen (approved):

PR #1802, opened 2021-04-20
Feature/allow disabling request logger for apps

LumbardhAgaj:
- Allows the user to manually configure request plugin in app and global settings.

### DIFF
--- a/plugins/logger/api/api.js
+++ b/plugins/logger/api/api.js
@@ -1,8 +1,6 @@
 var exported = {},
     common = require('../../../api/utils/common.js'),
-    plugins = require('../../pluginManager.js'),
-    pluginManager = require('../../pluginManager.js');
-
+    plugins = require('../../pluginManager.js');
 
 var RequestLoggerStateEnum = {
     on: "on",
@@ -84,7 +82,7 @@ plugins.setConfigs("logger", {
 
     var turnRequestLoggerOffIfNecessary = function(params, requestLoggerConfiguration) {
         if (requestLoggerConfiguration.state === RequestLoggerStateEnum.automatic && automaticStateRequestLoggerManager.shouldTurnOffRequestLogger(requestLoggerConfiguration.limit)) {
-            pluginManager.updateApplicationConfigs(common.db, params.app._id, "logger", Object.assign(requestLoggerConfiguration, {state: RequestLoggerStateEnum.off}));
+            plugins.updateApplicationConfigs(common.db, params.app._id, "logger", Object.assign(requestLoggerConfiguration, {state: RequestLoggerStateEnum.off}));
         }
     };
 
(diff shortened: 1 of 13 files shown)


plugins/logger/frontend/public/javascripts/countly.views.js
```diff
@@ -369,6 +370,28 @@ window.LoggerView = countlyView.extend({
...
+                '</div>';
+        setTimeout(function() {
+            $(".routename-logger #logger-alert").append(warningAlertContainer);
+            $(".event-remind-tooltip .fa-times").off("click").on("click", function(e) {
+                $(e.currentTarget.parentElement.parentElement).remove();
+            });
+        }, 0);
+    },
+    displayLoggerStateWarningAlertIfNecessary: function() {
+        var loggerRequestAppState = countlyGlobal.apps[countlyCommon.ACTIVE_APP_ID].plugins.logger.state;
```
ar2rsawseen, line 391: Hello,
I think if the plugin does not have any separate configs changed, it is possible the app document would not have plugins property at all. So all descending properties should be null checked.

plugins/logger/api/api.js
```diff
@@ -2,9 +2,22 @@ var exported = {},
     common = require('../../../api/utils/common.js'),
     plugins = require('../../pluginManager.js');
 
+plugins.setConfigs("logger", {
+    state: "on"
```
ar2rsawseen, line 6: Any reason why you went with the string "on/off" instead of boolean true/false?
I think boolean would also provide input out of the box too, so you would not have to implement your own.

ar2rsawseen (changes requested):

plugins/logger/api/api.js
LumbardhAgaj: As we discussed. Since we are thinking of introducing a new state that will automatically disable the logging og requests when a certain amount is reached, I kept it as string values.

plugins/logger/frontend/public/javascripts/countly.views.js
```diff
@@ -369,6 +370,28 @@ window.LoggerView = countlyView.extend({
...
+        setTimeout(function() {
+            $(".routename-logger #logger-alert").append(warningAlertContainer);
+            $(".event-remind-tooltip .fa-times").off("click").on("click", function(e) {
+                $(e.currentTarget.parentElement.parentElement).remove();
+            });
+        }, 0);
+    },
+    displayLoggerStateWarningAlertIfNecessary: function() {
+        var activeAppPlugins = countlyGlobal.apps[countlyCommon.ACTIVE_APP_ID].plugins;
+        if (activeAppPlugins && activeAppPlugins.logger.state === 'off') {
```
ar2rsawseen, line 392: I think there could still be a state when plugins object is defined (because the app has configs modified previously), but the logger object is not, because it does not propagate automatically to all apps, I think
so logger.state check would throw an exception

LumbardhAgaj: good to know as I did not expect that would be the case if the plugin is loaded. I just put in the null pointer check for logger plugin.

ar2rsawseen (approved):

PR #1828, opened 2021-05-12
SERVER-2202/Implement slipping away users using Vue

LumbardhAgaj:
- Initial implementation of slipping away users page using Vue

- BREAKS: 
- export as slipping away users table
- UI mockup in reference to filter (query builder that renders on modal is not supported yet)


plugins/slipping-away-users/frontend/public/javascripts/countly.models.js
```diff
@@ -1,30 +1,106 @@
...
-(function(countlySlippingPlugin, $) {
-    var _data = {};
-    /**
-     * This is for initializing model
-     * @param {object} query user drill filter query
-     * @return {func} ajax func to request data and store in _data
-    */
-    countlySlippingPlugin.initialize = function(query) {
+/*global countlyCommon,jQuery,countlyVue*/
+(function(countlySlippingAwayUsers, $) {
```
can-keklik, line 3: Is there another plugin which uses this? If so, deleting the old API could break things.

plugins/slipping-away-users/frontend/public/javascripts/countly.models.js
```diff
@@ -1,30 +1,106 @@
...
-            url: countlyCommon.API_URL + "/o/slipping",
-            data: {
-                app_id: countlyCommon.ACTIVE_APP_ID,
-                method: 'slipping',
-                query: JSON.stringify(query),
-            },
-            success: function(json) {
-                _data = json;
+    countlySlippingAwayUsers.helpers = {
+        buildFilters: function(filters) {
```
can-keklik, line 5: For now, this could stay. But in the future we might need to transfer this functionality to query builder itself.

plugins/slipping-away-users/frontend/public/javascripts/countly.models.js
```diff
@@ -1,30 +1,106 @@
...
+                slippingAwayUsersFilters: { query: {}, byVal: []},
+                isLoading: false,
+                hasError: false,
+                error: null,
+            };
+        };
+
+        var slippingAwayUsersActions = {
+            fetchSlippingAwayUsers: function(context) {
+                context.dispatch('onFetchInit');
```
can-keklik, line 51: Using "slippingAwayUsers" in action names (or mutations) feels a bit verbose.

plugins/slipping-away-users/frontend/public/javascripts/countly.models.js
```diff
@@ -1,30 +1,106 @@
...
+        };
+
+        var slippingAwayUsersMutations = {
+            setSlippingAwayUsers: function(state, value) {
+                state.slippingAwayUsers = value;
+            },
+            setSlippingAwayUsersFilters: function(state, value) {
+                state.slippingAwayUsersFilters = value;
+            },
+            fetchInit: function(state) {
```
can-keklik, line 83: Handling {Init, Error, Success} cases is a good pattern but naming could be better. Another word for "fetch", for instance.

plugins/slipping-away-users/frontend/public/javascripts/countly.views.js
```diff
@@ -1,296 +1,154 @@
...
+var SlippingAwayUsersBarChart = countlyVue.components.BaseComponent.extend({
+    template: "#slipping-away-users-bar-chart",
+    data: function() {
+        return {};
+    },
+    computed: {
+        slippingAwayUsers: function() {
+            return this.$store.state.countlySlippingAwayUsers.slippingAwayUsers;
+        },
+        slippingAwayUsersOptions: function() {
```
can-keklik, line 31: Current cly-chart implementation offers almost no abstractions in terms of axes and other configuration. Once chart implementation is finalized we might need to get back to this. @itsiprikshit

plugins/slipping-away-users/frontend/public/javascripts/countly.views.js
```diff
@@ -1,296 +1,154 @@
...
+            }
+            window.location.hash = '/users/query/' + JSON.stringify(data);
         }
-        this._query = filter;
-        app.navigate("/analytics/slipping-away/" + JSON.stringify(filter), false);
-        this.refresh();
+    }
+});
+
+var SlippingAwayUsersPage = countlyVue.components.BaseComponent.extend({
```
can-keklik, line 110: By convention view names should be suffixed with "View". So it should be SlippingAwayUsersView. In addition, views should extend countlyVue.views.BaseView. This would save you from importing i18n and a few other mixins.

plugins/slipping-away-users/frontend/public/templates/SlippingAwayUsersPage.html
```diff
@@ -0,0 +1,79 @@
...
+                <el-table-column sortable="custom" prop="count" :label="i18n('slipping-away-users.table-count')">
+                </el-table-column>
+                <el-table-column sortable="custom" prop="percentage" :label="i18n('slipping-away-users.table-percentage')">
+                    <template slot-scope="scope">
+                        <div class="bu-columns bu-is-gapless">
+                            <div class="bu-column bu-is-2">
+                                <span>{{scope.row.percentage}}%</span>
+                            </div>
+                            <div class="bu-column bu-is-10">
+                                <el-progress :percentage="parseInt(scope.row.percentage)" class="slipping-away-users-td-progress-bar" :show-text="false" :color="progressbarColor"></el-progress>
```
can-keklik, line 66: As far as I know, we do not use el-progress, but another progress bar provided by Bulma. It is not a Vue component for the time being, but we can wrap it and use it as primary progress component. @itsiprikshit

plugins/slipping-away-users/frontend/public/templates/SlippingAwayUsersPage.html
```diff
@@ -0,0 +1,79 @@
...
+        </cly-main>
+    </div>
+</script>
+
+<!-- NOTE-LA: <slipping-away-users-filter> is a temporary created component until query filter with modal becomes available-->
+<script type="text/x-template" id="slipping-away-users-filter">
+    <cly-section style="min-height:150px">
+        <div class="bu-columns bu-is-gapless bu-is-mobile bu-is-centered" style="padding-top:1%">
+            <div class="bu-column bu-is-12-tablet bu-is-9-desktop bu-is-7-widescreen bu-is-6-fullhd">
+                <cly-qb-segmentation
```
can-keklik, line 24: In some cases (e.g. when drill is not active) this component wouldn't be found. We need to handle those cases.

can-keklik: But anyway it is a temporary one right now.

plugins/slipping-away-users/frontend/public/localization/slipping.properties
```diff
@@ -1,17 +0,0 @@
-#slipping
```
can-keklik, line 1: @ar2rsawseen Would changing the file name cause an issue for translations?

plugins/slipping-away-users/frontend/public/javascripts/countly.models.js
```diff
@@ -1,30 +1,106 @@
...
-        });
+            return newQuery;
+        }
     };
 
-    countlySlippingPlugin.getData = function() {
-        return _data;
+    countlySlippingAwayUsers.service = {
+        fetchSlippingAwayUsers: function(filters, callback) {
+            $.ajax({
```
can-keklik, line 20: Please check CV.$.ajax in docs, if you want to promisify this.

can-keklik (changes requested):

plugins/slipping-away-users/frontend/public/javascripts/countly.models.js
LumbardhAgaj: Assuming that other endpoints will want to send the same query format then I agree it should move inside query billder (I left a note).

plugins/slipping-away-users/frontend/public/javascripts/countly.models.js
LumbardhAgaj: I was not able to find any plugin that does it (except for its former view beforeRender() method, which has become obsolete now).

plugins/slipping-away-users/frontend/public/javascripts/countly.models.js
LumbardhAgaj: I, myself, have problems to succinctly use naming variables (count. working on it). Please let me know if you have any suggestions in this particular case. I must say, I am not fond of short, general variable names that lack context.

plugins/slipping-away-users/frontend/public/javascripts/countly.models.js
LumbardhAgaj: you're right,fetch does not have any meaning in our case since we do not use the fetch() method to make requests. please let me know if you think replacing it with ''request'' is better or if you have other ideas.

plugins/slipping-away-users/frontend/public/templates/SlippingAwayUsersPage.html
LumbardhAgaj: If that's the case then probably it's better we create a component for it. I can always do that.

plugins/slipping-away-users/frontend/public/javascripts/countly.models.js
can-keklik: Since countlySlippingAwayUsers already creates a context, everything defined under that is automatically namespaced. I also agree with you regarding using short and vague naming, but examples such as "countlySlippingAwayUsers/fetchSlippingAwayUsers" sounds/feels a bit verbose. Instead of that, for instance, we could use something like "countlySlippingAwayUsers/fetchAll".

plugins/slipping-away-users/frontend/public/localization/slipping.properties
LumbardhAgaj: actually there were two file names that were serving the same purpose (slipping-away-users.properties and slipping.properties). So I removed slipping.properties and merged its content inside slipping-away-users (this is also consistent with other int. naming convensions.

plugins/slipping-away-users/frontend/public/javascripts/countly.models.js
can-keklik: My point wasn't about using the fetch() API, but rather about the action (don't confuse with the Vuex actions) we describe. I mean the thing you do there is not fetching something, but mutating the state depending on the status of fetch you triggered. So, perhaps we should use a wording such as "setRequestStatus". You could also refactor the logic so that there is only one mutator function (e.g. setRequestStatus('loading', 'error', 'success')) and do state changes in it.

plugins/slipping-away-users/frontend/public/templates/SlippingAwayUsersPage.html
can-keklik: Can you please add this to our Vue wishlist then?

plugins/slipping-away-users/frontend/public/javascripts/countly.models.js
LumbardhAgaj: I understand and in a hindsight I saw it too that there was a discrepancy between method name and what it was doing. Since, I have used fetch throughout the model, I will keep it then and instead add ''set'' as prefix for all the fetch methods, e.g. setFetchInit(), setFetchSuccess() and setFetchError() (I don't mind using a single function but then we will have to operate on a flag argument and it can easily become bloated).

PR #1834, opened 2021-05-20
Feature/SERVER-2203 implement user activity page

LumbardhAgaj:
- Implements user activity page in Vue

BREAK:
 - bar chart legend is not customized to support tooltip icon, user activity information and position

BONUS:
- convert css to saas for slipping away users plugin
- add route view with query parameter for slipping away users page


frontend/express/public/templates/user-activity/UserActivity.html
```diff
@@ -0,0 +1,124 @@
...
+                    </div>
+                </div>
+            </div>
+        </div>
+    </cly-section>
+</script>
+
+<script type="text/x-template" id="user-activity-bar-chart">
+    <cly-section>
+        <cly-chart-bar :option="userActivityOptions" style="min-height: 400px;" :autoresize="true" v-loading="isLoading"></cly-chart-bar>
```
itsiprikshit, line 48: charts accept height as a prop, so instead of using style, you can pass :height="400". 
Note: height should be a Number.
autoresize is also true by default so no need to set it.

frontend/express/public/javascripts/countly/countly.views.js
```diff
@@ -638,410 +638,186 @@ window.UserView = countlyView.extend({
...
+        userActivity: function() {
+            return this.$store.state.countlyUserActivity.userActivity;
+        },
+        seriesTotal: function() {
+            return this.$store.state.countlyUserActivity.seriesTotal;
+        },
+
+        userActivityOptions: function() {
+            return {
+                legend: {
```
itsiprikshit, line 687: charts automatically has a default legend setting.
unless you want to change that, no need to pass legend options explicitly.

frontend/express/public/javascripts/countly/countly.views.js
```diff
@@ -638,410 +638,186 @@ window.UserView = countlyView.extend({
...
+                    top: "bottom",
+                    padding: [0, 0, 20, 0]
+                },
+                xAxis: {
+                    type: "category",
+                    data: this.xAxisUserActivitySessionBuckets
+                },
+                yAxis: {
```
itsiprikshit, line 695: you can even skip yAxis options, by default its type is value.

frontend/express/public/javascripts/countly/countly.views.js
```diff
@@ -638,410 +638,186 @@ window.UserView = countlyView.extend({
...
+                return {
+                    data: self.userActivity[userActivityKey].map(function(item) {
+                        return item.count;
+                    }),
+                    type: "bar",
+                    name: self.barChartItemsLegends[userActivityKey],
+                    itemStyle: {
+                        borderRadius: [4, 4, 0, 0],
+                        color: self.barChartItemsColors[userActivityKey]
+                    },
```
itsiprikshit, line 716: bar charts auto selects a color from an array of provided colors.
i dont think you need to provide color aswell.
also i think i should add borderRadius as default to the bar chart.

plugins/slipping-away-users/frontend/public/javascripts/countly.views.js
```diff
@@ -35,25 +35,13 @@ var SlippingAwayUsersBarChart = countlyVue.views.BaseView.extend({
...
-                    type: "category",
                     data: this.xAxisSlippingAwayUsersPeriods
                 },
-                yAxis: {
-                    type: "value",
-                },
                 series: [{
                     data: this.yAxisSlippingAwayUsersCount,
                     name: CV.i18n('slipping-away-users.barchart-description'),
                     type: "bar",
```
itsiprikshit, line 44: same here.

PR #1845, opened 2021-05-25
Feature/server 2244 implement progress bar component

LumbardhAgaj:
- Implement progress bar component with single and multiple bar stacks

BONUS:
- re-use progress bar component in user activity and slipping away users pages
- refactor user activity model to fill empty or not found series values


frontend/express/public/javascripts/countly/vue/templates/progress.html
```diff
@@ -0,0 +1,8 @@
+<script type="text/x-template" id="cly-progress-bar">
```
itsiprikshit, line 1: i dont think we need this file. do we?

LumbardhAgaj: good catch.  I will remove it.

PR #1847, opened 2021-05-25
changes to add cly-vue-menubox and its css

abhisheksingla02:
Cly-vue-menubox and its css added


can-keklik:
Hi,

We've already deprecated the cly-menubox as we'll be using Element's dropdowns. This one seems no different from the former, so I'm afraid this is irrelevant and I need to reject it.

itsiprikshit:
I think abhishek was working on it and its not the latest code maybe.
Is it @abhisheksingla02 ?

abhisheksingla02:
Hi Can, Yes you can reject it. This is outdated code and we wont be using this component as it is. I'll raise a seperate pr for the required component :)

PR #1867, opened 2021-06-02
Update NGINX parameters

kaanklky:


kaanklky:
Yep, I forgot about Ubuntu support, ignore my existence...

PR #1870, opened 2021-06-04
Info / Warning Tooltip Icon

LVMBDV:
From the Vue.js Components Wishlist.


frontend/express/public/stylesheets/vue/clyvue.scss
```diff
@@ -3324,3 +3324,9 @@
         margin: 16px;
     }
 }
+
```
itsiprikshit, line 3327: can you please move this to /stylesheets/styles/blocks/helpers.scss and @use it in manifest.scss :)
you might have to create the helpers.scss file in blocks.
clyvue.scss will be deprecated once we move to the new ui entirely.

frontend/express/public/javascripts/countly/vue/components/helpers.js
```diff
@@ -140,4 +140,21 @@
                     </div>'
     }));
 
+    Vue.component("cly-tooltip-icon", countlyBaseComponent.extend({
+        mixins: [
+            _mixins.i18n
```
LumbardhAgaj, line 145: Not sure if we still want to add the Int mixins even though we are not using them inside the component (just in case you missed it, otherwise I don't think it has a major impact).

LumbardhAgaj (approved):

frontend/express/public/javascripts/countly/vue/components/helpers.js
itsiprikshit: makes sense, we can remove it if its not being used.

frontend/express/public/stylesheets/vue/clyvue.scss
LVMBDV: done 👍

PR #1871, opened 2021-06-04
Feature/server 2208 implement session frequency

LumbardhAgaj:
Feature: implement session frequency using Vue


frontend/express/public/stylesheets/styles/manifest.scss
```diff
@@ -29,4 +29,7 @@
 @use "states/state";
 
 // All core "Plugins" views specific SASS partials
-@use "../../core/user-activity/stylesheets/main"
+@use "../../core/user-activity/stylesheets/main" as user-activity-main-style;
+@use "../../core/session-overview/stylesheets/main" as session-overview-main-style;
+@use "../../core/session-durations/stylesheets/main" as session-durations-main-style;
+@use "../../core/session-frequency/stylesheets/main" as session-frequency-main-style;
```
itsiprikshit, line 35: for all the imported scss partials, i dont think we need a namespace for them.

LumbardhAgaj: the compiler complains because of the same file name being used, hence I had to namespace it.

PR #1897, opened 2021-06-24
[new-user-management] initialized new user management vue transition.

frknbasaran:


frontend/express/public/core/user-management/stylesheets/_main.scss
```diff
@@ -0,0 +1,301 @@
+.user-management-view {
```
can-keklik, line 1: Too much custom CSS, but OK 😄

frknbasaran: let's find the right way together on DM. :D

PR #1911, opened 2021-07-07
Feature/new UI

Cookiezaurs:
New design for analytics/technology.


can-keklik (commented):
I just couldn't understand the purpose of using a separate date picking. Other than that, a great PR, thanks :)

PR #1914, opened 2021-07-07
made a new component to have multiple dropdowns in a box and handle t…

abhisheksingla02:
…he essential events for them

### DIFF
--- a/frontend/express/public/javascripts/countly/vue/components/dropdown.js
+++ b/frontend/express/public/javascripts/countly/vue/components/dropdown.js
@@ -1,8 +1,9 @@
-/* global jQuery, Vue, ELEMENT */
+/* global jQuery, Vue, ELEMENT, CV */
 
 (function(countlyVue) {
 
-    var countlyBaseComponent = countlyVue.components.BaseComponent;
+    var countlyBaseComponent = countlyVue.components.BaseComponent,
+    _mixins = countlyVue.mixins;
 
     Vue.component("cly-input-dropdown-trigger", countlyBaseComponent.extend({
         props: {
@@ -187,8 +188,9 @@
         }
     }));
 
-    Vue.component("cly-vue-menubox", countlyBaseComponent.extend({
-        template: '<cly-dropdown ref="dropdown" v-on="$listeners" class="cly-vue-menubox-new__dropdown">\
+    Vue.component("cly-fields-select", countlyBaseComponent.extend({
+        mixins: [_mixins.i18n],
+        template: '<cly-dropdown ref="dropdown" v-on="$listeners" class="cly-fields-select__dropdown">\
                         <template v-slot:trigger="dropdown">\
                         <slot name="trigger">\
                             <cly-input-dropdown-trigger\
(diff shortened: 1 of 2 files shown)


frontend/express/public/javascripts/countly/vue/components/dropdown.js
```diff
@@ -187,6 +187,123 @@
...
+        },
+        data: function() {
+            return {
+                internalValue: function() {
+                    return {};
+                },
+            };
+        },
+        watch: {
+            value: {
```
itsiprikshit, line 267: do we need to watch it?
i believe value never changes since its not being used in the template.
i speculate its only changed if the parent changes it.

frontend/express/public/javascripts/countly/vue/components/dropdown.js
```diff
@@ -187,6 +187,123 @@
         }
     }));
 
+    Vue.component("cly-vue-menubox", countlyBaseComponent.extend({
```
can-keklik, line 190: I think this needs a better name:

-Shouldn't contain "vue" keyword
-Should emphasize the function 

For instance, cly-fields-select

frontend/express/public/javascripts/countly/vue/components/dropdown.js
```diff
@@ -187,6 +187,123 @@
...
+        template: '<cly-dropdown ref="dropdown" v-on="$listeners" class="cly-vue-menubox-new__dropdown">\
```
can-keklik, line 191: Class names should be updated in accordance with the component name change, of course.

frontend/express/public/javascripts/countly/vue/components/dropdown.js
```diff
@@ -187,6 +187,123 @@
...
+                            </cly-input-dropdown-trigger>\
+                        </slot>\
+                        </template>\
+                        <div class="cly-vue-menubox-new menubox-default-skin">\
+                        <div class="menu-body">\
+                            <div>\
+                            <div>\
+                            <span class="cly-vue-menubox-new__title">{{title}}</span>\
+                            <span class="cly-vue-menubox-new__reset" @click="reset"> Reset Filters</span>\
+                            </div>\
```
can-keklik, line 210: You should use el-button instead of custom elements.

frontend/express/public/javascripts/countly/vue/components/dropdown.js
```diff
@@ -187,6 +187,123 @@
...
+                        </div>\
+                    </cly-dropdown>',
+        props: {
+            cancelLabel: {
+                type: String,
+                default: 'Cancel'
+            },
+            confirmLabel: {
+                type: String,
+                default: 'Confirm'
```
can-keklik, line 239: These default values should be localized using CV.i18n

frontend/express/public/stylesheets/styles/blocks/_dropdown.scss
```diff
@@ -0,0 +1,79 @@
+@use "../base/typography-variables" as t;
+
+.cly-vue-menubox-new {
```
can-keklik, line 3: Component name should be changed as I mentioned earlier :)

frontend/express/public/stylesheets/styles/blocks/_dropdown.scss
```diff
@@ -0,0 +1,79 @@
...
+        height:32px;
+        border-radius: 4px;
+        background: #F6F6F6;
+        padding: 10px 12px;
+      }
+      &__dropdown {
+        .el-input__inner::placeholder{
+            color:black;
+        }
+      }
```
can-keklik, line 78: These are probably some debris from the legacy component I wrote, but you should definitely avoid this much of custom styling for fundamental elements. For instance, buttons shouldn't be styled here, font-family shouldn't be specified here. The bottom line is that custom styling should be the last resort.

frontend/express/public/javascripts/countly/vue/components/dropdown.js
can-keklik: I think this is fine Prikshit. This mechanism ensures that parent changes are always reflected on the component, while the opposite is not, so that it requires the user's explicit confirmation for value update.

can-keklik (changes requested):

frontend/express/public/javascripts/countly/vue/components/dropdown.js
itsiprikshit: ok got it.

itsiprikshit: but even if internalValue is changed within the component, it wont reflect in the parent bcz we are not changing value.
I suppose if we set internalValue as computed property it should work.

abhisheksingla02: So basically without the watcher here , using the computedProperty the value will change while selecting the internal dropdowns and hence that would effect the main dropdown label :)

can-keklik: @itsiprikshit But we still want to modify internalValue. Setting it computed would make it read-only, right?

abhisheksingla02: Hey @can-keklik . I think @itsiprikshit meant something like this...        internalValue: function() {
                return this.value;
            }

PR #1930, opened 2021-07-22
sidebar

itsiprikshit:


frontend/express/public/javascripts/countly/vue/templates/sidebar/users-menu.html
```diff
@@ -0,0 +1,198 @@
+<el-popover
```
can-keklik, line 1: As I recall, el-popover was a leaky component. So I avoided using it. I'll double check.

frontend/express/public/javascripts/countly/vue/core.js
```diff
@@ -51,7 +51,7 @@
     // @vue/component
     var commonFormattersMixin = {
         methods: {
-            formatTimeAgo: countlyCommon.formatTimeAgo,
+            formatTimeAgo: countlyCommon.formatTimeAgoText,
```
can-keklik, line 54: It feels like this will break some stuff, but let it break. We need "text" version anyway.

frontend/express/public/javascripts/countly/vue/components/nav.js
```diff
@@ -102,19 +102,12 @@
...
-
-                    Only emit input event when no route provided
-                */
                 if (tab.length && tab[0].route) {
                     if (this.noHistory) {
                         Backbone.history.noHistory(tab[0].route);
                     }
                     else {
-                        window.location.hash = tab[0].route;
+                        app.navigate(tab[0].route, true);
```
can-keklik, line 110: Are you sure that (`.hash=` and `app.navigate`) both things yield the exact result?

frontend/express/public/javascripts/countly/vue/components/sidebar.js
```diff
@@ -0,0 +1,425 @@
...
+                    "menus": "/sidebar/analytics/menu",
+                    "submenus": "/sidebar/analytics/submenu"
+                })
+            ],
+            data: function() {
+                var apps = _.sortBy(countlyGlobal.apps, function(app) {
+                    return (app.name + "").toLowerCase();
+                });
+
+                apps = apps.map(function(a) {
```
can-keklik, line 49: forEach would do the same thing as you just mutate the existing items in the array, rather than creating new ones.

frontend/express/public/javascripts/countly/vue/core.js
```diff
@@ -121,6 +121,27 @@
                         }
                     }
                 }
+            },
+            countlySidebar: {
```
can-keklik, line 125: That's totally fine, I mean registering the sidebar in core, as it is a permanent thing, that doesn't need to be unregistered anytime.

frontend/express/public/javascripts/countly/vue/core.js
itsiprikshit: actually last i checked formatTimeAgo was not being referenced anywhere, so nothing should break.

frontend/express/public/javascripts/countly/vue/components/nav.js
itsiprikshit: actually here the idea is when the route is present we want the backbone routing to trigger, so passing true as second parameter gives us the desired result.

frontend/express/public/javascripts/countly/vue/components/sidebar.js
```diff
@@ -0,0 +1,425 @@
...
+                if (currMenu.length) {
+                    this.$store.dispatch("countlySidebar/updateSelectedMenuItem", {menu: "management", item: currMenu[0]});
+                }
+            }
+        });
+
+        var SidebarView = countlyVue.views.create({
+            template: CV.T('/javascripts/countly/vue/templates/sidebar/sidebar.html'),
+            mixins: [
+                countlyVue.container.dataMixin({
```
itsiprikshit, line 271: @can-keklik

PR #1957, opened 2021-08-16
Added missing action button item which "Pluggin Settings"

pnrgenc:
Added missing action button item which "Pluggin Settings"


can-keklik:
Could you please resolve the conflict? 🙂

PR #1964, opened 2021-08-18
[systemlogs] new ui

ar2rsawseen:


plugins/systemlogs/frontend/public/javascripts/countly.views.js
```diff
@@ -1,539 +1,414 @@
...
-                            CountlyHelpers.reopenRows(self.dtable, self.expandTable, self);
-                        }
-                    });
-                },
-                "fnServerParams": function(aoData) {
-                    if (self._query) {
-                        aoData.push({ "name": "query", "value": JSON.stringify(self._query) });
+                return false;
+            },
+            renderField: function(key, field, sub) {
```
can-keklik, line 20: I think it would be better to move renderField fn to methods, rather than computed, as you "call" it with some args in various places.

plugins/systemlogs/frontend/public/javascripts/countly.views.js
```diff
@@ -1,539 +1,414 @@
...
+                loaded: true,
+                selectAction: this.$route.params.query.a || "all",
+                selectUser: this.$route.params.query.user_id || "all",
+                detailData: {},
+                back: this.$route.params.back || false,
+                tableStore: tableStore,
+                allActions: {"all": jQuery.i18n.map["systemlogs.all-actions"]},
+                allUsers: {"all": jQuery.i18n.map["systemlogs.all-users"]},
+                remoteTableDataSource: countlyVue.vuex.getServerDataSource(tableStore, "systemLogsTable"),
+                tablePersistKey: "systemLogsTable_localTable_" + countlyCommon.ACTIVE_APP_ID,
```
can-keklik, line 147: This might be irrelevant, as the view uses only a remote table?

plugins/systemlogs/frontend/public/javascripts/countly.views.js
can-keklik: Also we could move this logic to template itself, since it is mostly about creating visual output.

plugins/systemlogs/frontend/public/javascripts/countly.views.js
```diff
@@ -1,539 +1,414 @@
...
+                remoteTablePersistKey: "systemLogsTable_remoteTable_" + countlyCommon.ACTIVE_APP_ID,
```
can-keklik, line 148: This one also not mentioned in the template.

can-keklik: (remoteTablePersistKey)

plugins/systemlogs/frontend/public/javascripts/countly.views.js
can-keklik: (tablePersistKey)

plugins/systemlogs/frontend/public/javascripts/countly.views.js
```diff
@@ -1,539 +1,363 @@
...
-        try {
-            query = JSON.parse(query);
-        }
-        catch (ex) {
-            query = null;
-        }
-        this.systemLogsView._query = query;
-        this.renderWhenReady(this.systemLogsView);
+    //register views
+    app.systemLogsView = new countlyVue.views.BackboneWrapper({
```
itsiprikshit, line 152: https://github.com/Countly/countly-server/pull/1968/files#r692696143

itsiprikshit (approved):

PR #1968, opened 2021-08-19
[errorlogs] new ui

ar2rsawseen:


plugins/errorlogs/frontend/public/javascripts/countly.views.js
```diff
@@ -1,83 +1,82 @@
...
+            }
         }
-    },
-    refresh: function() {
-    }
-});
+    });
 
+    //register views
+    app.errorLogsView = new countlyVue.views.BackboneWrapper({
```
itsiprikshit, line 61: i don't think we need to assign the returned value to app object.
you can simply create a variable here.
or imo the best practice is to create a function that returns the backbone view and call that function only when the route is triggered.
https://github.com/Countly/countly-server/blob/9a72eba5057bc960d078df5d32dbcdfc27de9938/plugins/vue-example/frontend/public/javascripts/countly.views.js#L737-L768

itsiprikshit: what do you think @can-keklik

can-keklik: Yeah, I've been thinking about this. It apparently works this way, too. But our practical knowledge is limited about whether reusing the BackboneWrapper causes any trouble in terms of Vue instance's lifecycle, etc. 

I think it is better to create the BackboneWrapper in a lazy fashion (and not mount to `app`) for the time being, as Prikshit suggested.

itsiprikshit (approved):

PR #1970, opened 2021-08-21
[ui/cly-multi-select] added @updated custom event to component.

frknbasaran:
- added @updated custom event for allowing track unsaved changes in one of the select boxes.


can-keklik:
Well actually I had added a "change" event before you did :upside_down_face:

PR #1979, opened 2021-08-24
Feature/server 2303 validate message media, title and content

LumbardhAgaj:
Implement validation to message media URL fields, title and content.
Replace all tag elements with empty string for message title and content.
Use requirement specifications to validate message image and video media.

How to test it?
lumbardh.count.ly server has the latest changes

BONUS:
- Move all commonly used components under countly common components file.


plugins/push/frontend/public/javascripts/countly.models.js
```diff
@@ -347,13 +414,32 @@
                 dataType: "json"
             });
         },
+        fetchMediaMetadata: function(url) {
+            var self = this;
+            return new Promise(function(resolve, reject) {
```
can-keklik, line 419: You didn't need to wrap the CV.$.ajax with Promise, as it is already done in the CV.$.ajax itself. You might use then/catch instead of success/error, as well.

can-keklik (approved):

plugins/push/frontend/public/javascripts/countly.models.js
LumbardhAgaj: Yes, that's what I have for cases when I do not map the data to model. Unfortunately, I will have to keep them at least I am afraid for the upcoming release.

PR #1988, opened 2021-08-28
close select-x within cly-multi-select

kanwarujjaval:


can-keklik:
So that's it? I'd expect data manager's source to be longer, a couple of thousand lines may be 😄

itsiprikshit:
I think he lost all his code and this is what is left.

kanwarujjaval:
Updated the title 😄

PR #1993, opened 2021-08-31
Feature/server 2204 implement times of day

LumbardhAgaj:
Implement times of day user loyalty page.
Add user loyalty pages as dynamic container tabs.

How to test it?
use lumbardh.count.ly server
BONUS: refactor fetchMixin module to abstract the loader and instead accept it as parameter only for relevant actions


plugins/times-of-day/frontend/public/javascripts/countly.models.js
```diff
@@ -1,508 +1,172 @@
...
+                weekDaySerieItem.forEach(function(hourSerieValue) {
+                    if (maxValue < hourSerieValue) {
+                        maxValue = hourSerieValue;
+                    }
+                });
+            });
+            return maxValue;
+        },
+        mapTimesOfDayDtoToModel: function(dto) {
+            this.findMaxSeriesValue(dto);
```
itsiprikshit, line 72: maybe this can be removed.

LumbardhAgaj: Absolutely. Great catch Prikshit!.

plugins/times-of-day/frontend/public/javascripts/countly.views.js
```diff
@@ -1,225 +1,118 @@
...
-                    }
-                },
-                {
-                    "mData": "friday",
-                    "sType": "numeric",
-                    "sTitle": jQuery.i18n.map['times-of-day.friday'],
-                    "mRender": function(d) {
-                        return countlyCommon.formatNumber(d);
+                xAxis: {
+                    type: 'category',
```
itsiprikshit, line 43: In the BaseChart class which every chart inherits has xAxis by default 'category' btw.

plugins/times-of-day/frontend/public/stylesheets/main.scss
```diff
@@ -0,0 +1,28 @@
...
+    &__scatter-chart-tooltip-text {
+        font-size: typography.$text-sm-size;
+        line-height: typography.$text-button-line-height;
+    }
+
+    &__scatter-chart-tooltip-total-users-value {
+        font-weight: typography.$font-weight-primary;
+        font-size: typography.$text-lg-size;
+        line-height: typography.$heading-5-line-height;
+    }
```
itsiprikshit, line 20: we also have classes - **text-small** and **text-big**
maybe we can use them instead of creating new classes?

plugins/times-of-day/frontend/public/javascripts/countly.views.js
```diff
@@ -264,436 +157,3 @@ app.addPageScript("/custom#", function() {
         });
     }
 });
-
-$(document).ready(function() {
```
itsiprikshit, line 268: lets keep the following code, since this is required by custom dashboard.
removing this will break the times of day widget on custom dashboard.

plugins/times-of-day/frontend/public/templates/widget.html
```diff
@@ -1,51 +0,0 @@
-<div class="grid-stack-item-content cly-widget">
```
itsiprikshit, line 1: this file will also be required in order for the times of day widget to work on custom dashboards.

plugins/times-of-day/frontend/public/stylesheets/main.scss
LumbardhAgaj: I tried to re-use them but unfortunately the text property values did not match. I am sticking to design specifications first as that's how it willbe reviewed. Lack of design consistency is the culprit here I am afraid.

itsiprikshit (approved):

can-keklik:
you like large PRs 👀

PR #2006, opened 2021-09-06
Minor Fixes for events

abhisheksingla02:
1. Checks added and a few changes on overview page


frontend/express/public/core/events/javascripts/countly.details.models.js
```diff
@@ -526,7 +526,7 @@
                 allEventsData: {},
                 allEventsGroupData: [],
                 selectedEventsData: {},
-                selectedDatePeriod: "30days",
+                selectedDatePeriod: countlyCommon.getPeriod(),
```
itsiprikshit, line 529: any specific reason for having proxy for the period obj?
if it is to react to period changes you can use **dateChanged** method
like its being done here - https://github.com/Countly/countly-server/blob/7303b953e9f8f1fbea391f025c8b5cf811805d7e/frontend/express/public/core/session-frequency/javascripts/countly.views.js#L45-L47

itsiprikshit: also you should get the period from root getters - countlyCommon/period
i believe you are already using the same getter in nps and ias to get the period before making server request.
you can use the same here.

frontend/express/public/core/events/javascripts/countly.overview.views.js
```diff
@@ -220,6 +220,7 @@
                     return this.$store.getters["countlyEventsOverview/selectedDatePeriod"];
                 },
                 set: function(value) {
+                    countlyCommon.setPeriod(value);
```
itsiprikshit, line 223: you can get rid of this.
and move this.$store.dispatch('countlyEventsOverview/fetchMonitorEvents'); to dateChanged method.

PR #2008, opened 2021-09-07
Changes to cly-select-x to show selected count.(If set in settings that we need it)

Cookiezaurs:


frontend/express/public/javascripts/countly/vue/components/input.js
```diff
@@ -530,6 +531,14 @@
                     "cly-vue-select-x__pop--has-single-option": this.hasSingleOption
                 };
             },
+            showSelectedCount: function() {
```
can-keklik, line 534: This could be simplified as explained here: https://stackoverflow.com/a/49208677

frontend/express/public/javascripts/countly/vue/components/input.js
```diff
@@ -552,8 +561,10 @@
             innerValue: {
                 get: function() {
                     if (this.uncommittedValue && this.uncommittedValue !== this.value) {
+                        this.selectedCountText = CV.i18n('export.export-columns-selected-count', this.uncommittedValue.length, this.options.length);
```
can-keklik, line 564: You can define a computed selectedCountText property for this, instead of adding a side effect to getter of innerValue

can-keklik (changes requested):

PR #2009, opened 2021-09-07
cly-vue-radio-block : option to put them horizontally

Cookiezaurs:


frontend/express/public/javascripts/countly/vue/components/input.js
```diff
@@ -710,7 +710,8 @@
                 }
             },
             skin: { default: "main", type: String},
-            disabled: {type: Boolean, default: false}
+            disabled: {type: Boolean, default: false},
+            direction: { default: "vertical", type: String}
```
can-keklik, line 714: `direction` is referenced nowhere. Do I miss something?

frontend/express/public/javascripts/countly/vue/components/input.js
```diff
@@ -725,6 +726,44 @@
                     classes.push("disabled");
                 }
                 return classes;
+            },
+            wrapperClasses: function() {
+                var direction = "vertical";
+                if (this.$attrs && this.$attrs.radioDirection) {
```
can-keklik, line 732: I believe that using props for `radioDirection` would be a better practice here.

So you could define a `radioDirection` prop and pass a default value to it

can-keklik: That would also help you to eliminate local `direction` value

can-keklik (changes requested):

PR #2017, opened 2021-09-08
fixed visual bugs

pnrgenc:
fixed visual bugs

### DIFF
--- a/plugins/browser/frontend/public/stylesheets/browser.scss
+++ b/plugins/browser/frontend/public/stylesheets/browser.scss
@@ -50,29 +50,11 @@
         } 
     }
 
- 
-
     .cly-vue-metric-card {
         width: fit-content !important;
-        border-top: 1px solid #ececec;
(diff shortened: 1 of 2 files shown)


plugins/browser/frontend/public/stylesheets/browser.scss
```diff
@@ -50,39 +50,32 @@
...
 
-    h5 {
-        font-family: Inter;
-        font-style: normal;
-        font-weight: 500;
-        font-size: 14px;
-        line-height: 24px;
-        color: #333C48;
-        margin-bottom:20px;
+    .cly-vue-metric-card {
```
can-keklik, line 55: Would you think that it is possible to make this generic? If it is so we should better move it to core styling, rather than keeping it in a plugin :slightly_smiling_face:

pnrgenc: I removed these lines. I used "bulma" for some margin and background properties 🙂

PR #2043, opened 2021-09-24
Feature/server 2304 hook in push notification client side with server

LumbardhAgaj:
Connect push notification client with the server.

How to test it?
[lumbardh.count.ly](https://lumbardh.count.ly/) server has the latest changes.

BREAK:
 - Push notification message settings are not yet hooked in. They will be implemented when the new API supports settings separately for each platform.

WARNING:
- introduces a Vue warning that complains about overwriting the module state by the submodule itself. At the moment, the root cause is not known.


frontend/express/public/javascripts/countly/vue/imports.js
```diff
@@ -33,6 +33,18 @@
...
+            try {
+                JSON.parse(value);
+                return true;
+            }
+            catch (error) {
+                return false;
+            }
+        }
+    });
+
```
can-keklik, line 47: Pretty smart one :D

plugins/push/frontend/public/javascripts/countly.views.component.common.js
```diff
@@ -152,7 +152,7 @@
     countlyPushNotificationComponent.ReviewSectionRow = countlyVue.views.create({
         props: {
             value: {
-                type: String,
+                type: String | Number,
```
can-keklik, line 155: The correct way of doing the intended thing is [String, Number], as the current expression evaluates to 0.

can-keklik: https://stackoverflow.com/a/53608245

plugins/push/frontend/public/javascripts/countly.views.component.common.js
```diff
@@ -636,37 +636,87 @@
         },
     });
 
+    countlyPushNotification.DetailsTabRow = countlyVue.views.create({
+        template: '#details-tab-row',
+        props: {
+            value: {
+                type: String | Number,
```
can-keklik, line 643: "|"

plugins/push/frontend/public/javascripts/countly.models.js
```diff
@@ -99,271 +214,718 @@
...
-        mapStatus: function(status, error) {
-            if (StatusFinderHelper.isSending(status)) {
-                if (error) {
-                    return {value: 'sending-errors', label: CV.i18n('push-notification.status-sending-errors')};
+    };
+
+    countlyPushNotification.mapper = {
+        incoming: {
+            getUserPropertyElement: function(index, userProperty) {
+                var newElement = document.createElement("span");
```
can-keklik, line 262: I cannot get why createElement is needed here (similar to latter occurrences). It looks a bit hacky I can tell you that, but more importantly, It is placed in models.js

can-keklik: If it is related to string unescaping as in one of your other tickets, I don't think that it is a good practice. Because it allows all types of unescaping, which might be dangerous.

can-keklik (changes requested):

plugins/push/frontend/public/javascripts/countly.views.component.common.js
LumbardhAgaj: I saw your note earlier, yet I had no idea I was doing it. I will use arrays.

plugins/push/frontend/public/javascripts/countly.models.js
LumbardhAgaj: I should have added a comment perhaps explaining the purpose of the mappers as well as helper methods. This is an incoming mapper (it maps the server response). The reason is because the message title and content needs to be constructed given the user properties that it might have, e.g. at index position 5, insert a specific user property in message. I chose to do it in the model so that I don't have to construct the message everytime I use it. The incoming escaped content is simply processed.

PR #2044, opened 2021-09-24
UI improvement and logout redirection on same page

shaharyar-shamshi:


frontend/express/public/javascripts/countly/vue/templates/sidebar/users-menu.html
```diff
@@ -165,7 +165,7 @@
                         <i class="ion-android-exit"></i>
                     </div>
                     <div class="bu-level-item">
-                        <a href="#" target="_blank">
+                        <a>
```
itsiprikshit, line 168: why even have an anchar tag if it does nothing?
maybe we can remove it.

PR #2045, opened 2021-09-27
[um-style-changes] some style fixes.

frknbasaran:


frontend/express/public/localization/dashboard/dashboard.properties
```diff
@@ -17,17 +17,11 @@ common.session-duration = Session Duration
 common.table.no-data = No data
 common.search.no-match-found = No match found
 common.table.total-users = Total Users
-common.table.total-users-desc = Number of times your application is opened by a new or returning user from a unique device, in the selected time period.
```
Cookiezaurs, line 20: Please don't delete/overwrite localization strings !!!!

Cookiezaurs, line 20: Please don't delete/overwrite localization strings !!!!

Cookiezaurs (changes requested):

frontend/express/public/localization/dashboard/dashboard.properties
```diff
@@ -219,10 +213,6 @@ common.confirm = Confirm
 common.remove = Remove
 common.next = Next
 common.previous = Previous
-common.select-at-least = At least {0} items should be selected. 
-common.select-at-most = At most {0} items can be selected.
-common.and = and
```
can-keklik, line 224: Hmm, I agree with Anna. It is weird though... I don't think that you explicitly removed these fields, so why are they deleted?

frontend/express/public/localization/dashboard/dashboard.properties
frknbasaran: okay, please be calm.

frontend/express/public/localization/dashboard/dashboard.properties
frknbasaran: not sure, solving rn.

PR #2066, opened 2021-10-08
[vue-ratings] ratings segmentation.

frknbasaran:


itsiprikshit:
So I had a discussion with @ArtursKadikis about this yesterday.
So we might have to rethink about this @frknbasaran.

After the discussion it was concluded that we should make all feedback types - ratings, nps and surveys homogeneous in terms of api requests and popup urls (sdk side).

Also we have to return all feedback types in the same api call /o/sdk?method=feedback. This is also something furkan suggested initially but slightly with a different server side approach of doing it.

We might have to find a solution for this where the core dispatches calls to nps/surveys and ratings simultaneously and when both of them return the responses it is assembled in core and returned to the sdk maybe.
So before we merge this pr.
Lets go through all the cases.
I will document it and share the document.

Because we also have to think about manual reporting of ratings.

PR #2102, opened 2021-10-27
added hide-top prop to cly-datatable-n

frknbasaran:


can-keklik:
hide-header would be more consistent with the slots. Could you change it?

frknbasaran:
> `hide-header` name would be more consistent with the slots. Could you change it?

nope. because `hide-header` is not a good naming at all. :D  **header** and **top** are different things in datatable context, **header** is the row that contains column names and we have `show-header` prop for hiding **header** row that comes from element ui.

Screen Shot 2021-11-03 at 11 48 26

`hide-top` that I added with this PR isn't hiding header row, it's hiding the white section at the **top** of the table that contains search input, export buttons etc. I added an image below that explains the differences.

Screen Shot 2021-11-03 at 11 54 25

can-keklik:
you win.

PR #2107, opened 2021-10-28
[dbviewer-vue] first step of dbviewer -> vue transition.

frknbasaran:


can-keklik:
Eslint fails, seems ok except that.

PR #2111, opened 2021-11-01
vue transition of "Data Populator" plugin

pnrgenc:


pnrgenc:
Hello @can-keklik ,
We've talked about the things I need to fix about this PR. Regarding this,
- I changed faker.js to faker.min.js
- and another issue was to make a component of something I used. I talked to the design team for this. The component I use is not generic for now. It is only used in populator. We can reconsider this when they generalize it.

plugins/populator/frontend/public/templates/template_drawer.html
```diff
@@ -0,0 +1,112 @@
...
+        <cly-form-step id="create-template-step">
+            <div class="cly-vue-drawer-step__section">
+                <div class="text-button font-weight-bold bu-mb-2">{{ i18n('populator.template-name') }}</div>
+                <validation-provider name="name" rules="required" v-slot="v">
+                    <el-input class="bu-is-shadowless" :class="{'is-error': v.errors.length > 0}" :placeholder="i18n('populator.template-name-placeholder')" v-model="drawerScope.editedObject.name"></el-input>
+                </validation-provider>
+            </div>
+            <div class="cly-vue-drawer-step__section">
+                <h4 class="bu-mb-4">{{ i18n('populator.custom-user-props') }}</h4>
+                <div :key="index" class="populator-template-drawer__main-container bu-p-2" v-for="(upItem, index) in drawerScope.editedObject.up" >
```
can-keklik, line 12: Here is something confusing:

index is loop variable for editedObject.up, but you use it as editedObject.events[index] too. Is this intentional?

PR #2129, opened 2021-11-09
[user-profile-feedback-tab] added feedback tab to user profile detail.

frknbasaran:


can-keklik:
Does the exported component react to changes in user details page? If there are any expected changes there of course (e.g. date change)

PR #2134, opened 2021-11-11
css changes for cly-empty-home

pnrgenc:
css changes for cly-empty-home


pnrgenc:
According to the design in Figma, I made some changes about the places of the elements and css. 
(here is ref: [https://www.figma.com/file/PUjfGUiaVQpSBeS5EWMKYu/countly-components?node-id=20921%3A120431]
I can take it back if you don't find it proper 🙂 
fyi: @can-keklik

PR #2135, opened 2021-11-15
SERVER-2253 Remote config vue transition

abhisheksingla02:


plugins/remote-config/api/api.js
```diff
@@ -338,7 +338,7 @@ plugins.setConfigs("remote-config", {
         var maximumParametersAllowed = plugins.getConfig("remote-config").maximum_allowed_parameters;
         var maximumConditionsAllowed = plugins.getConfig("remote-config").conditions_per_paramaeters;
         var collectionName = "remoteconfig_parameters" + appId;
-
+        parameter.ts = Date.now();
```
itsiprikshit, line 341: Is this ts used as the created timestamp?

plugins/remote-config/frontend/public/templates/conditions-drawer.html
```diff
@@ -1,23 +1,85 @@
...
+    <cly-form-step id="conditions-drawer">
+      <div class="cly-vue-drawer-step__section">
+        <div class="bu-pb-4">
+          <div class="font-weight-bold text-smallish cly-vue-remote-config-conditions-drawer__margin-bottom">
+            {{i18n('remote-config.condition-name')}}
+          </div>
+          <div class="color-cool-gray-50 text-small cly-vue-remote-config-conditions-drawer__margin-bottom">
+            {{i18n('placeholder.condition-name')}}
+          </div>
+          <validation-provider name="condition_name" rules="required" v-slot="v">
```
itsiprikshit, line 17: please refrain from using validation-provider directly.
use `cly-form-field` instead.

plugins/remote-config/api/api.js
abhisheksingla02: yes, showing "-" for those

plugins/remote-config/frontend/public/templates/conditions-drawer.html
abhisheksingla02: sure:)

itsiprikshit: Incase you are worried that `cly-form-field` does not provide an option to add the description below the label.
you can add it to the component itself.
description can be used as a prop in the `cly-form-field` component then.

If you do this, you can also get rid of the class - `cly-vue-remote-config-conditions-drawer__margin-bottom`, and use bu-mb-2 or bu-mb-1 instead. Although i agree they don't equal 6px as mentioned in the design. We can discuss it with them.

plugins/remote-config/frontend/public/javascripts/countly.views.js
```diff
@@ -13,51 +18,247 @@
...
                 var action = "countlyRemoteConfig/parameters/create";
                 if (doc._id) {
                     action = "countlyRemoteConfig/parameters/update";
                 }
-
+                if (!doc.description || !this.showDescription) {
+                    doc.description = "-";
+                }
+                if (doc.status === "Expired" && doc.expiry_dttm > Date.now()) {
+                    doc.status = "Running";
```
itsiprikshit, line 54: Do you save the statuses in the database?
If so why?

plugins/remote-config/frontend/public/javascripts/countly.views.js
```diff
@@ -13,51 +18,247 @@
...
+        template: '<div class="bu-is-flex bu-is-flex-wrap-wrap bu-is-align-items-center">\
+                                <div class="cly-vue-remote-config-conditions-drawer__color-tag-wrapper"  v-for="(tag,idx) in tags" :data-color="tag.key">{{ tag.label }}\
+                                <div v-if="tag.key == selectedColorKey" @click="click" class="cly-vue-remote-config-conditions-drawer__color-tag cly-vue-remote-config-conditions-drawer__color-tag__selected bu-is-flex bu-is-align-items-center bu-is-justify-content-center" :style="{backgroundColor: tag.colorCode}" :data-color="tag.key">\
+                                    <i class="ion-checkmark cly-vue-remote-config-conditions-drawer__checkmark"></i>\
+                                </div>\
+                                <div v-else @click="click" class="cly-vue-remote-config-conditions-drawer__color-tag bu-is-flex bu-is-align-items-center bu-is-justify-content-center" :style="{backgroundColor: tag.colorCode}" :data-color="tag.key"></div>\
+                                </div>\
+                    </div>'
+    });
+    var JsonEditor = countlyVue.views.create({
```
itsiprikshit, line 128: This should be made a global component and kept in .../vue/components/helpers.js
Please move its template to core as well in .../vue/components/templates

itsiprikshit: Also please modify this component a bit.
json editor should be an independent component that can be used within cly-dialog or any other component.

Currently, json editor is a part of cly-dialog, so if someone wants to use json editor without the cly-dialog, they cannot do so.

plugins/remote-config/frontend/public/javascripts/countly.models.js
```diff
@@ -124,23 +131,33 @@
         var parametersResource = countlyVue.vuex.Module("parameters", {
             state: function() {
                 return {
-                    all: []
+                    all: [],
+                    showJsonEditor: false
```
itsiprikshit, line 135: this can be a local state within the component i think.

plugins/remote-config/frontend/public/javascripts/countly.views.js
```diff
@@ -13,51 +18,247 @@
...
 
+    var ColorTag = countlyVue.views.BaseView.extend({
+        data: function() {
+            return {
+                selectedColorKey: this.defaultColorKey
+            };
+        },
+        methods: {
+            click: function(event) {
+                this.selectedColorKey = event.target.getAttribute('data-color');
```
itsiprikshit, line 103: rather than accessing the data-color from event.target using the getAttribute,
you can pass the data-color value as argument to the click method.

itsiprikshit: [this](https://github.com/Countly/countly-server/pull/2135/files?authenticity_token=vYyp91wk%2BqYqB42yV3zUT0B64NoqmNshpVSsQqPt%2BER9kAGFeXeaaYSrEqXsOEhYYzrO7L9dAKs5HjQ7WawFIQ%3D%3D&file-filters%5B%5D=.html&file-filters%5B%5D=.js&file-filters%5B%5D=.properties&file-filters%5B%5D=.scss&hide-deleted-files=true#diff-35a50fbf8b26bca06763311058cfc63614f4383b1e0e8b9bfa7995533959b53cR121) should be changed to 
<div v-if="tag.key == selectedColorKey" @click="click(tag)"

itsiprikshit: Also move this component to helpers.js in core.

plugins/remote-config/frontend/public/javascripts/countly.views.js
```diff
@@ -13,51 +18,247 @@
...
+            "color-tag": ColorTag
+        },
         props: {
             controls: {
                 type: Object
             }
         },
+        computed: {
+            showSeedValue: function() {
+                if (!_.isEmpty(this.managedPropertySegmentation.queryText) && this.managedPropertySegmentation.queryText.includes("Random Percentile")) {
```
itsiprikshit, line 186: please use some id check in the query object instead of checking for the Random percentile in the queryText. It won't work as expected if the string is localized.
I am sure random percentile must have some id or something.

plugins/remote-config/frontend/public/javascripts/countly.views.js
```diff
@@ -13,51 +18,247 @@
...
+                    id: "disallowOperator",
+                    params: {
+                        selector: function(operator) {
+                            return ["cly.=", "cly.!=", "cly.contains", "cly.between", "cly.isset"].includes(operator.id);
+                        }
+                    }
+                })]
+            })];
+            additionalProperties.push(new countlyQueryBuilder.Property({
+                id: 'condition.randomPercentile',
```
itsiprikshit, line 209: please use the same id for random percentile as its on master - `up.random_percentile` i think to avoid compatibility issues.

plugins/remote-config/frontend/public/javascripts/countly.views.js
```diff
@@ -85,20 +308,55 @@
...
         },
         computed: {
             tableRows: function() {
                 return this.$store.getters["countlyRemoteConfig/parameters/all"];
             }
         },
         methods: {
+            getNumberOfConditionsText: function(conditions) {
+                if (conditions.length === 0) {
+                    return "1 condition";
```
itsiprikshit, line 323: if conditions length is 0, shouldn't we return 0 condition?

plugins/remote-config/frontend/public/javascripts/countly.views.js
```diff
@@ -85,20 +308,55 @@
...
+                }
+                return conditions.length + 1 + "conditions";
```
itsiprikshit, line 325: why are you adding 1 to the conditions.length?

plugins/remote-config/frontend/public/templates/parameters.html
```diff
@@ -1,36 +1,88 @@
...
+        <el-table-column type="expand" min-width="50">
+          <template v-slot="props">
+            <cly-section>
+              <el-tabs type="border-card" value="default_value">
+                <el-tab-pane name="default_value" :label="i18n('remote-config.default-value')">
+                  <div>
+                    <pre>{{ props.row.default_value }}</pre>
+                  </div>
+                </el-tab-pane>
+                <el-tab-pane name="conditions" :label="i18n('remote-config.conditions')">
```
itsiprikshit, line 27: this tab should only be shown if drill is enabled. please add the check for that.

plugins/remote-config/frontend/public/javascripts/countly.models.js
kanwarujjaval: Agreed, You can just pass this as a prop to the component

plugins/remote-config/frontend/public/javascripts/countly.views.js
abhisheksingla02: This is for default value, which is a mandatory field

plugins/remote-config/frontend/public/javascripts/countly.views.js
abhisheksingla02: This is for default value too

plugins/remote-config/frontend/public/javascripts/countly.views.js
kanwarujjaval: Yes, We should try to use vue abstractions as much as possible and not interact with dom elements directly.

plugins/remote-config/frontend/public/templates/parameters.html
abhisheksingla02: yes right, will add it here:)

plugins/remote-config/frontend/public/javascripts/countly.views.js
kanwarujjaval: but if conditions are 0, why should the default value be 1 condition? @itsiprikshit

abhisheksingla02: default value is being stored seperately in db

itsiprikshit: afaik there is no default value for conditions.
if it was addd later, please point me to the use case and db schema for that. @abhisheksingla02

plugins/remote-config/frontend/public/javascripts/countly.views.js
itsiprikshit: default value is a mandatory field i agree.
but why is it counted in the conditions?
default value of a parameter has nothing to do with the conditions or its values.

plugins/remote-config/frontend/public/javascripts/countly.views.js
abhisheksingla02: @itsiprikshit  this is not default value for condition but default value field for parameter itself. aacording to design we needed to show both. https://www.figma.com/file/PUjfGUiaVQpSBeS5EWMKYu/countly-components?node-id=11374%3A77951

plugins/remote-config/frontend/public/javascripts/countly.views.js
abhisheksingla02: This was according to design. I think we can discuss with Duygu and remove it:) 

https://www.figma.com/file/PUjfGUiaVQpSBeS5EWMKYu/countly-components?node-id=11374%3A77951

plugins/remote-config/frontend/public/javascripts/countly.views.js
abhisheksingla02: Good catch. Yes that would be the ideal way :)

plugins/remote-config/frontend/public/javascripts/countly.views.js
abhisheksingla02: It is not possile with the current object. But can did give the solution for this yestaerday, havent commited it yet. Will update the pr :)

plugins/remote-config/frontend/public/javascripts/countly.views.js
abhisheksingla02: Yeah , sure.

frontend/express/public/stylesheets/vue/clyvue.scss
```diff
@@ -23,6 +23,26 @@
     }
 }
```
itsiprikshit, line 25: this should be added to _helpers.scss in `frontend/express/public/stylesheets/styles/blocks/_helpers.scss`

frontend/express/public/stylesheets/vue/clyvue.scss
```diff
@@ -23,6 +23,26 @@
...
 
+.cly-vue-color-tag {
```
itsiprikshit, line 26: this should be added to _helpers.scss in frontend/express/public/stylesheets/styles/blocks/_helpers.scss

kanwarujjaval (approved):

plugins/remote-config/frontend/public/javascripts/countly.models.js
itsiprikshit: as discussed on slack.
this is to be done in subsequent pr's.
just making a note.

PR #2165, opened 2021-11-30
Feature/server 2323 update push plugin to use new api

LumbardhAgaj:
Adapt new push API in the client side.

Features:
  - Implement mappers for message settings.
  - Implement support for silent message type.

Note: some information (createdBy, isSilent, reschedule) is not yet supported from the server.


plugins/push/frontend/public/javascripts/countly.views.component.common.js
```diff
@@ -503,6 +497,7 @@
                 defaultLocalizationValidationErrors: [],
                 selectionRange: null,
                 mutationObserver: null,
+                userPropertyEvents: new Map(),
```
can-keklik, line 500: I hope this works fine in IE11. Map itself is supported apparently, but may be limited in terms of functionality.

LumbardhAgaj: yes, the support is limited. however, I've only used the green functionalities marked from our caniuse reference guide. It's worth giving a try.

PR #2166, opened 2021-11-30
Feature/email report attachments

waiterZen:


plugins/reports/api/reports.js
```diff
@@ -640,30 +640,36 @@ var metricProps = {
...
-                if (mail.sendPoolMail) {
-                    mail.sendPoolMail(msg);
-                }
-                else {
-                    mail.sendMail(msg);
-                }
+                var options = { "directory": "/tmp", "width": "1028px", height: "1000px" };
+                pdf.create(msg.html, options).toFile('/tmp/email_report_' + new Date().getTime() + '.pdf', function(err, res) {
+                    if (err) {
+                        return console.log(err);
```
itsiprikshit, line 655: instead of console.log please use the logger module.
The error message should be displayed on debug `log.d` only.

PR #2191, opened 2021-12-14
Feature/reports ui update

waiterZen:


plugins/reports/frontend/public/javascripts/countly.views.js
```diff
@@ -33,6 +32,9 @@
             };
         },
         methods: {
+            createReport: function() {
+                $("#create-report-button").click();
```
itsiprikshit, line 36: why are we using still jquery in reports? 😄

plugins/reports/frontend/public/templates/vue-main.html
```diff
@@ -19,16 +19,43 @@ <h2> {{i18n('reports.title')}}
...
+        <div   style="width: 400px;">
+            <h3 class="color-cool-gray-100 bu-mt-5 bu-mb-4 bu-has-text-centered">...hmm, seems it’s empty here</h3>
+            <div class="color-cool-gray-50 text-medium bu-has-text-centered">
+                Create something to Lorem ipsum dolor sit amet,
+                consectetur adipiscing elit, 
+            </div>
+            <div  v-on:click="createReport()" class="button bu-has-text-centered bu-mt-5 color-blue-100 pointer">+ Create New Report</div>
+        </div>
+    </div>
+</div>
```
itsiprikshit, line 36: I believe you are creating a global component for the empty state. Right? @waiterZen 
Will you add that to this pr only ? Or should I merge this pr and later you will change this code and use the global component here ?

plugins/reports/frontend/public/javascripts/countly.views.js
waiterZen: fixed

plugins/reports/frontend/public/templates/vue-main.html
waiterZen: global component also here now.

plugins/alerts/frontend/public/templates/vue-main.html
```diff
@@ -49,9 +62,9 @@ <h2> {{i18n('alert.plugin-title')}}
                 </div>
             </template> 
             <template v-slot="scope">
-                <el-table-column fixed="left" width="100" label="">
+                <el-table-column fixed="left" width="108" label="">
                     <template v-slot="rowScope">
-                        <el-switch :value="rowScope.row.enabled"
+                        <el-switch :value="rowScope.row.enabled"  class="switch-column"
```
itsiprikshit, line 67: instead of using a custom class of switch column.
you can do this - https://countly.slack.com/archives/C018XS3SA0G/p1631607559006400
in the el-table-column, set type to switch

PR #2199, opened 2021-12-15
Feature/server 2342 save push notification drafts

LumbardhAgaj:
Implement saving push notification drafts with the rest of user actions such as editing, duplicating, deleting push notifications  and editing a draft push notification.

Refactor:
 - Separate data access communication to api object from push notification service.
 - Replace the list of locale values with a list of locale items in the model.


plugins/push/frontend/public/templates/common-components.html
```diff
@@ -261,20 +261,30 @@ <h5>{{ category }}</h5>
...
-                    </template>
-                </details-tab-row>
-                <details-tab-row :label="i18n('push-notification-details.message-content')">
-                    <template v-for="(component) in previewMessageContent">
-                        <keep-alive>
-                            <component v-bind:is="component.name" :value="component.value"></component>
-                        </keep-alive>
-                    </template>
-                </details-tab-row>
+                <div class="bu-level bu-ml-5 bu-mb-4 bu-is-justify-content-flex-start bu-is-align-items-flex-start">
```
LumbardhAgaj, line 264: unfortunately the default slot of details-tab-row component was being assigned falsy value even though there were components being passed. As a result, I ended up removing details-tab-row for this specific case only.

plugins/push/frontend/public/javascripts/countly.models.js
```diff
@@ -192,6 +195,113 @@
                 periods: {daily: []},
             };
         },
+        getInitialBaseModel: function() {
```
LumbardhAgaj, line 198: I have moved initial model state from view in helper object and ended up creating helper methods for each notificaiton type.

plugins/push/frontend/public/javascripts/countly.models.js
```diff
@@ -302,12 +418,148 @@
         },
     };
 
+    //NOTE: api object will reside temporarily in countlyPushNotification until countlyApi object is created;
```
LumbardhAgaj, line 421: Separated the data access communication from the module service. Once, I can put in the countlyApi global object then I will move the interface inside it.

plugins/push/frontend/public/javascripts/countly.models.js
```diff
@@ -1548,12 +1659,153 @@
                     });
             });
         },
+        fetchAll: function(type) {
```
LumbardhAgaj, line 1662: The service will use the api object to make the calls. It's responsbility is only to know what to do with data transfer objects.

plugins/push/frontend/public/javascripts/countly.models.js
```diff
@@ -1677,6 +1926,11 @@
             platformFilter: countlyPushNotification.service.PlatformEnum.ALL,
             totalPushMessagesSent: 0,
             totalUserActionsPerformed: 0,
+            userCommand: {
```
LumbardhAgaj, line 1929: userCommand state with isDrawerOpen will be moved in its own module, so that actions, mutations in reference to the drawer will be re-used for main and details pages.

PR #2218, opened 2021-12-20
[SERVER-2433]: New component: empty state for charts

pnrgenc:


frontend/express/public/javascripts/countly/vue/components/vis.js
```diff
@@ -1109,13 +1130,24 @@
...
+                                    <div class="bu-mt-6">\
+                                        <img src="/images/icons/empty-icon.svg" alt="">\
+                                    </div>\
+                                    <div class="bu-mt-2">\
+                                        <h4>{{i18n("common.emtpy-table-title")}}</h4>\
+                                    </div>\
+                                    <div class="bu-mt-1 bu-mb-6">\
+                                        <span class="text-small color-cool-gray-50">{{i18n("common.emtpy-table-subtitle")}}</span>\
+                                    </div>\
+                                </div>\
```
itsiprikshit, line 1143: You can maybe create a local empty chart data component within vis.js with the above template and use that component in all chart components rather than using this chunk of template everywhere.
In case we want to make changes in the empty state for charts, we will have to make changes everywhere. However, if it were a component, the changes required would be only within the component.

pnrgenc: Fixed :)

PR #2227, opened 2021-12-22
Feature/server 2424 implement notification toast messages in vuex module

LumbardhAgaj:
Implements notification toast messages using global vuex module. By default, notification toasts messages are automatically hidden after 7 seconds. 

Use dispatchNotificationToast() utility method to display a new notification toast message e.g. 
`countlyCommon.dispatchNotificationToast({text: "Test message", color: "light-warning"});`


can-keklik:
It is all good. I think it could be even better if we changed CountlyHelpers.notify's body accordingly.

frontend/express/public/javascripts/countly/vue/core.js
```diff
@@ -372,6 +388,24 @@
         });
     };
 
+    var NotificationToastsView = VueCompositionAPI.defineComponent({
```
can-keklik, line 391: Why composition API? :slightly_smiling_face:

frontend/express/public/javascripts/countly/countly.common.js
```diff
@@ -205,6 +205,31 @@
...
+            }
+        };
+
+        /**
+         * Generates unique id string using unsigned integer array.
+         * @returns {string} unique id
+         */
+        countlyCommon.generateId = function() {
+            var crypto = window.crypto || window.msCrypto;
+            var uint32 = crypto.getRandomValues(new Uint32Array(1))[0];
```
can-keklik, line 229: still good enough, but we could simply use a counter too :slightly_smiling_face:

LumbardhAgaj: that's what I initially had. My thinking was that I wanted to make the ids more robust while also not polluting the common vuex module with counter state

frontend/express/public/javascripts/countly/vue/core.js
LumbardhAgaj: It's difficult to explain but my understanding was that I needed to render a vue component inside the core component (wrapper), meaning that any countlyVue element has not yet been initiated. I can tell that using composition API is on overkill but I really did not know what else would be the best approach to do it, plus I already saw that we were using the same interface with the dummy component. any suggestions?

PR #2229, opened 2021-12-22
[alerts] update calc logic & notification

waiterZen:


plugins/alerts/api/api.js
```diff
@@ -75,6 +75,12 @@ const FEATURE_NAME = 'alerts';
         }, 2000);
     });
 
+
+    plugins.register("/alerts/addAlertCount", function(ob) {
+        console.log("/alerts/addAlertCount", ob);
```
itsiprikshit, line 80: please remove console.log
use logger instead.

PR #2282, opened 2022-01-07
Reports alerts hooks updates

waiterZen:


plugins/hooks/frontend/public/templates/vue-drawer.html
```diff
@@ -16,9 +17,16 @@
                     </el-input>
                 </cly-form-field>
 
-                <cly-form-field name="description" :label="i18n('hooks.description')">
-                    <div class="cly-vue-hook-drawer__help-text">
-                        {{i18n('hooks.optional')}}
+                <cly-form-field name="description">
```
itsiprikshit, line 20: cly-form-field accepts label and optional as prop. So you did not have to write your custom template for it unless its different.

waiterZen: great, updated

PR #2310, opened 2022-01-12
Feature/server 1846 add push to custom dashboard

LumbardhAgaj:
Add push notification as a plugin to custom dashboard.

NOTE: at the moment the server does not support adding push notification as plugin. However, the implementation of the push notification widget is done assuming that the API will not change as far as data goes. It has been tested using fake data objects.


plugins/dashboards/frontend/public/javascripts/countly.helpers.js
```diff
@@ -134,40 +134,39 @@
...
         }
     });
 
     var AppCountComponent = countlyVue.views.create({
         template: CV.T('/dashboards/templates/helpers/drawer/app-count.html'),
         props: {
-            apps: {
-                type: Array,
-                default: []
+            value: {
```
LumbardhAgaj, line 152: @itsiprikshit I used a custom v-model to app count component because the sourceapps component was not reacting to user app count selection, e.g. whenever user changed app count selection, the source apps selection multiplicity remained the same. As a matter of fact, user was able to select one application only. Please let me know if you have any thoughts on this.

plugins/dashboards/frontend/public/javascripts/countly.helpers.js
```diff
@@ -134,40 +134,39 @@
             },
             types: function() {
                 var self = this;
-                var result = this.allTypes.filter(function(item) {
+                if (this.enabledTypes && !this.enabledTypes.length) {
```
LumbardhAgaj, line 137: fixes a small bug for data type that was not showing any data type options when enabledTypes was empty.

can-keklik:
Should we merge with CD changes? @itsiprikshit

itsiprikshit (approved):

itsiprikshit:
@can-keklik Yes we can merge it.

plugins/dashboards/frontend/public/javascripts/countly.helpers.js
itsiprikshit: Yeah this is fine i suppose.
The component was incomplete earlier, thanks for making the change :)

PR #2319, opened 2022-01-13
Feature/server 2503 dynamic back link

LumbardhAgaj:
Implement goTo() and getBacklinkRoute() helper methods that store and return the backlink url.

getBacklinkRoute() method must be called only inside the view data initialization because it deletes the backlink route after reading it.

### DIFF
--- a/frontend/express/public/javascripts/countly/countly.helpers.js
+++ b/frontend/express/public/javascripts/countly/countly.helpers.js
@@ -322,14 +322,14 @@
     };
 
     CountlyHelpers.goTo = function(options) {
-        app.backlinkRoute = options.from;
+        app.backlinkUrl = options.from;
         window.location.hash = options.url;
     };
 
-    CountlyHelpers.getBacklinkRoute = function() {
-        var route = app.backlinkRoute;
-        app.backlinkRoute = null;
-        return route;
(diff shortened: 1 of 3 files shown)


frontend/express/public/javascripts/countly/countly.helpers.js
```diff
@@ -321,6 +321,16 @@
...
     };
 
+    CountlyHelpers.goTo = function(options) {
+        app.backlinkRoute = options.from;
+        window.location.hash = options.url;
+    };
+
+    CountlyHelpers.getBacklinkRoute = function() {
+        var route = app.backlinkRoute;
+        app.backlinkRoute = null;
```
LumbardhAgaj, line 331: Unfortunately, due to view re-renders I was not able to keep the backlink route reactive (add it in the global store) since we remove it after we read it. Hence, I decided to save it outside of the store.

can-keklik:
I think "cly-back-link" should be modified in order to check the redirection. So that, we can add it to the header component, which would display all the redirections in a unified way. What do you guys think about that? @LumbardhAgaj @itsiprikshit

can-keklik:
Yeah, that would work too. Can you do that?

LumbardhAgaj:
Sure, I will include it here.

PR #2406, opened 2022-01-27
Feature/server 2531 delete push notification keys

LumbardhAgaj:
User explicitly deletes a push notification key. Upon doing so, he must confirm his action using the dialog.

fix:
 - updating rate overwrite error in the server-side


plugins/push/api/api-push.js
```diff
@@ -181,11 +181,11 @@ module.exports.onAppPluginsUpdate = async({params, app, config}) => {
         }
 
         if (config.rate.period) {
-            update.$set = {'plugins.push.rate.period': config.rate.period};
+            update.$set = Object.assign(update.$set, {'plugins.push.rate.period': config.rate.period});
```
iartem, line 184: good catch

plugins/push/api/api-push.js
```diff
@@ -181,11 +181,11 @@ module.exports.onAppPluginsUpdate = async({params, app, config}) => {
...
             pushcfg.rate.period = config.rate.period;
         }
         else {
-            update.$unset = {'plugins.push.rate.period': 1};
+            update.$unset = Object.assign(update.$unset, {'plugins.push.rate.period': 1});
```
iartem, line 188: bad catch XD 

`$unset` might not exist here

plugins/push/api/api-push.js
iartem: ok, it's a bad catch as well lol as `$set` might not be there either

plugins/push/frontend/public/templates/push-notification-app-level-config.html
```diff
@@ -59,4 +70,14 @@
...
+    <cly-confirm-dialog 
+        title="Delete key"
+        dialogType="danger"
+        saveButtonLabel="I understand, delete this key" 
+        cancelButtonLabel="Cancel" 
+        :visible.sync="isDeleteKeyDialogVisible"
+        @cancel="onDeleteKeyCancel" 
+        @confirm="onDeleteKeyConfirm">
+    </cly-confirm-dialog>
+</cly-confirm-dialog >
```
can-keklik, line 82: You can use CountlyHelpers.confirm unless you have a specific reason not to do so :slightly_smiling_face:

can-keklik (approved):
I left just one comment. Other than that, it is up to @iartem :slightly_smiling_face:

plugins/push/frontend/public/templates/push-notification-app-level-config.html
LumbardhAgaj: yes I will. It slipped my mind.

PR #2425, opened 2022-01-31
crashes design review fix

FoysalOsmany:


plugins/crashes/frontend/public/javascripts/countly.models.js
```diff
@@ -550,7 +550,7 @@
                         data: mapKeys
                     },
                     series: [
-                        {data: Object.values(mapObj)}
+                        {data: Object.values(mapObj), color: "#F96300", name: `${jQuery.i18n.prop("crashes.fatal_crash_count")} by ${chartBy}`}
```
can-keklik, line 553: We don't have string interpolation on UI. Could you do this with concatenation? Also we have `CV.i18n` instead of `jQuery.i18n.prop` :slightly_smiling_face:

can-keklik (changes requested):

plugins/crashes/frontend/public/javascripts/countly.models.js
FoysalOsmany: fixed

PR #2451, opened 2022-02-02
Feature/server 2510 implement location picker

LumbardhAgaj:
Implement map location picker component. Support single location selection.


frontend/express/public/javascripts/countly/vue/components/vis.js
```diff
@@ -1594,6 +1594,111 @@
...
+                },
+                default: null
+            }
+        },
+        data: function() {
+            return {
+                markerCoordinates: null,
+                defaultCenterCoordinates: {lat: 48.66194284607008, lng: 8.964843750000002}, // Note: arbitrary coordinates used for map center when user location is not found
+                userCenterCoordinates: null,
+                tileFeed: 'https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
```
can-keklik, line 1619: This might be a problem for offline users. :slightly_smiling_face:

frontend/express/public/javascripts/countly/vue/components/vis.js
```diff
@@ -1594,6 +1594,111 @@
...
+                return 6;
+            },
+        },
+        methods: {
+            setMarkerCoordinates: function(coordinates) {
+                this.markerCoordinates = coordinates;
+            },
+            onLocationClick: function(event) {
+                this.setMarkerCoordinates(event.latlng);
+                this.$emit('select', event.latlng);
```
can-keklik, line 1661: We can emit the value, it is ok. How do we get the initial value tho? So I would expect a :value/@input pair here, but may be there is something I'm missing :slightly_smiling_face:

frontend/express/public/javascripts/countly/vue/components/vis.js
LumbardhAgaj: Well, one way it so move that information to our server. Not sure how much is there though. In saying that, the implementation is the same as what we had, so not sure how the offline users were supported.

frontend/express/public/javascripts/countly/vue/components/vis.js
LumbardhAgaj: You're right.I initially thought that I did not need it but it turns out we are also displaying a specific location that was created using the map.

frontend/express/public/javascripts/countly/vue/components/vis.js
```diff
@@ -1593,6 +1593,138 @@
...
+            'l-marker': Vue2Leaflet.LMarker,
+            'l-circle': Vue2Leaflet.LCircle,
+        },
+        props: {
+            value: {
+                type: [Object],
+                required: false,
+                validator: function(item) {
+                    return item && item.lng && item.lat;
+                },
```
can-keklik, line 1609: This validator might cause errors (plus, there is this `required: false` and `default: null`). For async binding occassions, there will be time spans in which the value will be falsy making validator fn fail.

LumbardhAgaj: I am not sure I follow in terms of async binding. When required is set to false, validator with false values passes.

PR #2468, opened 2022-02-04
Fixed actions of the form on the applications page and implemented the expected component

hilalozturk1:


frontend/express/public/core/app-management/templates/app-management.html
```diff
@@ -235,9 +276,9 @@ <h2 class="bu-mb-2" v-if="!newApp" >{{i18n('management-applications.plugins')}}
                         <cly-diff-helper :diff="changeKeys" @discard="onDiscard" @save="saveSettings" :disabled="v.invalid" :emitSaveWhenDisabled="true"></cly-diff-helper>
                     </validation-observer>
                 </cly-section>
-                <el-card class="box-card" v-if="!newApp" >
-                    <h3>{{i18n('common.integrate-sdks')}}</h3>
-                    <p>{{i18n('common.integrate-sdks-text')}}</p>
+                <el-card class="box-card color-cool-gray-100" v-if="!newApp" >
+                    <p>{{i18n('common.integrate-sdks')}}</p>
+                    <p><a href='https://code.count.ly' target='_blank'>Countly Code Generator </a>{{i18n('common.integrate-sdks-text')}}</p>
```
can-keklik, line 281: Does this hurt whitelabeling @ar2rsawseen?

ar2rsawseen: There was a specific setting inside Settings -> Frontend -> Show Code Generator for SDK integration
so this part should only be displayed when that setting is true and should not be displayed when setting is false :)

can-keklik: @hilalozturk1

ar2rsawseen: @hilalozturk1 just to speed things app, the setting should be in `countlyGlobal.config.code` so you can expose it to template and put in v-if

frontend/express/public/core/app-management/stylesheets/_main.scss
```diff
@@ -10,8 +76,89 @@
...
+            }
+          }  
+        }
+      }
+      #push-notification-app-level-config:first-child {
+        font-size: 16px;
+        line-height: 24px;
+        font-weight: 500;
+        color: #333C48;
+      }
```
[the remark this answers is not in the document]
LumbardhAgaj: If the style is specific to push notifications plugin (or any other plugin as a matter of fact) only then I agree. When this is the case the template is located in respective plugin so I would expect to find the css there as well.

can-keklik:
This is broken, replaced by #2516

PR #2494, opened 2022-02-08
epdata action fix and ui improvement

shaharyar-shamshi:


plugins/compliance-hub/frontend/public/javascripts/countly.views.js
```diff
@@ -396,7 +397,16 @@
 
         },
         methods: {
+            formatTableNumber: function(data) {
```
itsiprikshit, line 400: are you using this method?

PR #2588, opened 2022-02-17
Popup closable issue and some CSS problems fixed

pnrgenc:


pnrgenc:
Hello @hilalozturk1 ,
I made a small css change to cly-notification component. Could you check too if that's ok 🙂 Thanks!

hilalozturk1 (approved):

PR #2676, opened 2022-02-23
Fix/active app error on create app

LumbardhAgaj:
Remove broadcast() method when communicating state changes between plugins.

broadcast() method is unreliable because if the component which listents to the event is re-mounted than it will not be able to pick up that event.


frontend/express/public/core/app-management/javascripts/countly.models.js
```diff
@@ -0,0 +1,28 @@
+/*global countlyVue */
```
LumbardhAgaj, line 1: State change should be communicated using vuex modules only.

frontend/express/public/core/app-management/javascripts/countly.views.js
```diff
@@ -11,7 +11,7 @@
                 set: function(value) {
                     this.newApp = false;
                     this.selectedApp = value;
-                    this.broadcast('AppSettingsContainerObservable', 'selectedApp', value);
```
LumbardhAgaj, line 14: when a new application is created, it re-renders the plugins collapse view. As a consequence, no view that is subscribed to "selectedApp" event will not pick it up. Ultimately, this leads to stale data in those views as they still think that the active app id is the one before the new app was created.

PR #2730, opened 2022-02-25
Fix/times of day dashboard properties,and height error

LumbardhAgaj:
Use backward compatible period values for times of day widget.
Remove height property from times of day widget scatter chart.

### DIFF
--- a/plugins/times-of-day/frontend/public/templates/times-of-day-widget.html
+++ b/plugins/times-of-day/frontend/public/templates/times-of-day-widget.html
@@ -13,6 +13,6 @@
     <clyd-primary-legend :apps="data.apps" :show-period="false"></clyd-primary-legend>
 
     <div class="clyd-widget__content">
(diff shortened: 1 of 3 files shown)


plugins/times-of-day/frontend/public/templates/times-of-day-widget.html
```diff
@@ -13,6 +13,6 @@
     <clyd-primary-legend :apps="data.apps" :show-period="false"></clyd-primary-legend>
 
     <div class="clyd-widget__content">
-        <times-of-day-scatter-chart :series="series" :maxSeriesValue="maxSeriesValue" :show-zoom="false" :show-toggle="false" :showDownload="false" height="auto" skin="full"></times-of-day-scatter-chart>
+        <times-of-day-scatter-chart :series="series" :maxSeriesValue="maxSeriesValue" :show-zoom="false" :show-toggle="false" :showDownload="false" skin="full"></times-of-day-scatter-chart>
```
itsiprikshit, line 16: why did you remove height: auto?

LumbardhAgaj: It seems like we are stretching the charts to parent height only. I will revert it.

PR #2779, opened 2022-02-28
Empty pie chart with value 0. (Showing grey circle)

Cookiezaurs:


frontend/express/public/javascripts/countly/vue/components/vis.js
```diff
@@ -754,9 +754,23 @@
             };
         },
         computed: {
+            hasAllEmptyValues: function() {
+                var options = _merge({}, this.baseOptions, this.mixinOptions, this.option);
```
itsiprikshit, line 758: instead of doing the merge here again.
can't we just do 
`var options = this.mergedOptions` ?

itsiprikshit (approved):

PR #2793, opened 2022-02-28
Added user permissions to crashes plugin.

pnrgenc:


plugins/crashes/frontend/public/templates/overview.html
```diff
@@ -34,13 +34,13 @@ <h2>
...
-                            <el-button @click="setSelectedAs('unresolved')" icon="fas fa-circle crash-icon-color-orange"> {{i18n("crashes.action-unresolved")}}</el-button>
-                            <el-button @click="setSelectedAs('delete')" icon="el-icon-delete">{{i18n("crashes.action-delete")}}</el-button>
-                            <el-button @click="setSelectedAs('hide')" icon="fas fa-eye-slash crash-icon-color-gray"> {{i18n("crashes.action-hide")}}</el-button>
-                            <el-button @click="setSelectedAs('show')" icon="fas fa-eye crash-icon-color-gray"> {{i18n("crashes.action-view")}}</el-button>
-                            <el-button @click="setSelectedAs('resolved')" icon="fas fa-circle crash-icon-color-green"> {{i18n("crashes.action-resolved")}}</el-button>
-                            <el-button @click="setSelectedAs('resolving')" icon="fas fa-circle crash-icon-color-blue"> {{i18n("crashes.action-resolving")}}</el-button>
+                        <div class="crashes-action-bar" v-if="selectedCrashgroups.length > 0 && (canUserCreate || canUserDelete || canUserUpdate)">
```
ar2rsawseen, line 37: why do we need canUserCreate check here if used actions are only update and delete? :)

pnrgenc: yes, it was unnecessary. I fixed :)

PR #2842, opened 2022-03-03
last column width not fixed

shaharyar-shamshi:


plugins/compliance-hub/frontend/public/templates/userConsentHistory.html
```diff
@@ -81,7 +81,7 @@
                     </template>
 
                 </el-table-column>
-                <el-table-column width="230" sortable="custom" prop="ts" label="TIME">
```
kanwarujjaval, line 84: Instead of removing you could make this min-width @shaharyar-shamshi

itsiprikshit: yeah that would be better.

shaharyar-shamshi: done

shaharyar-shamshi: @itsiprikshit  @kanwarujjaval

PR #2845, opened 2022-03-03
fix: No push credentials platform error message

LumbardhAgaj:


plugins/push/api/api-message.js
```diff
@@ -3,6 +3,21 @@ const { Message, Creds, State, Status, platforms, Audience, ValidationError, Tri
     common = require('../../../api/utils/common'),
     log = common.log('push:api:message');
 
+
+/**
+ * 
+ * @param {string} key mobile platform short key 
+ * @returns {string} full platform name
+ */
+function getPlatformName(key) {
```
iartem, line 12: Can you take it from `/dashboard`?

LumbardhAgaj: sure, please let me know where do you want me to put it.

iartem: Sorry I meant taking logic from dashboard endpoint. That is using `PLATFORM_TITLES` from [here](https://github.com/Countly/countly-server/blob/feature/new-ui/plugins/push/api/api-dashboard.js#L5)

PR #2855, opened 2022-03-04
[feature-enable-hooks] added mechanism that set all flags false when …

frknbasaran:
…a new feature enabled.


frknbasaran:
goodbye my little PR... 🥲

PR #2866, opened 2022-03-04
feature to purge and export of user data implemented

shaharyar-shamshi:


plugins/compliance-hub/frontend/public/javascripts/countly.views.js
```diff
@@ -1,14 +1,84 @@
-/*global $, CV, app, countlyVue, countlyConsentManager, countlyCommon, countlyConsentManager */
+/*global $, CV, app, countlyVue, countlyConsentManager, countlyCommon, countlyConsentManager, CountlyHelpers, countlyGlobal */
 (function() {
+    var FEATURE_NAME = "compliance-hub";
```
itsiprikshit, line 3: Feature name for compliance hub is `compliance_hub`. Please update it.

plugins/compliance-hub/frontend/public/templates/user.html
```diff
@@ -38,6 +38,34 @@
 
                     </template>
                 </el-table-column>
+                <el-table-column>
```
itsiprikshit, line 41: pass the following prop to the column -
`type="options"`

plugins/compliance-hub/frontend/public/templates/user.html
```diff
@@ -38,6 +38,34 @@
...
+                    <template v-slot="rowScope">
+                        <cly-more-options  size="small">
```
itsiprikshit, line 43: Since we need to show the action button on hover.
You need to a if condition here - 
`v-if="rowScope.row.hover"`

PR #2884, opened 2022-03-07
removed icons for the items in the dropdown

hilalozturk1:


can-keklik:
I think all these items are valuable, so we should just comment them out rather than hard deleting.

PR #2932, opened 2022-03-10
Feature/add url to sent and actioned users

LumbardhAgaj:


plugins/push/frontend/public/javascripts/countly.views.js
```diff
@@ -1677,9 +1686,63 @@
             getRemainingStackBar: function(value) {
                 return {data: [100 - value], itemStyle: {color: "#E2E4E8"}, silent: true};
             },
-
             onDrawerClose: function() {
                 this.$store.dispatch('countlyPushNotification/details/onSetIsDrawerOpen', false);
+            },
+            onGoToSent: function() {
+                var queryData = {message: {$in: ["609526bd23c8ba7ef0b1e099"]}};
```
can-keklik, line 1693: What is this harcoded thing?

plugins/push/frontend/public/javascripts/countly.views.js
```diff
@@ -1677,9 +1686,63 @@
...
+                };
+                CountlyHelpers.goTo({
+                    url: '/users/request/' + JSON.stringify(queryData),
+                    from: "#/" + countlyCommon.ACTIVE_APP_ID + "/messaging/details/" + this.pushNotification._id,
+                    title: CV.i18n("push-notification.back-to-push-notification-details")
+                });
+            },
+            onGoToErrored: function() {
+                var self = this;
+                var queryData = {message: {"$nin": ["609526bd23c8ba7ef0b1e099"]}};
```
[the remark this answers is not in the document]
iartem: wow you actually review them @can-keklik :D

plugins/push/frontend/public/javascripts/countly.views.js
LumbardhAgaj: I fixed it for both cases. Forgot to replace the actual one.

plugins/push/frontend/public/javascripts/countly.views.js
```diff
@@ -1677,9 +1686,63 @@
...
+                    queryData.geo = {"$in": this.pushNotification.locations};
+                }
+                if (this.pushNotification.cohorts && this.pushNotification.cohorts.length) {
+                    queryData.chr = {"$in": this.pushNotification.cohorts};
+                }
+                this.pushNotification.platforms.forEach(function(platform) {
+                    if (self.PlatformEnum.ANDROID === platform) {
+                        $in.push('tkap');
+                    }
+                    if (self.PlatformEnum.IOS === platform) {
```
iartem, line 1735: This is wrong. How many times I asked you to take platforms and fields from `/dashboard` 😭

plugins/push/frontend/public/javascripts/countly.views.js
```diff
@@ -1677,9 +1686,63 @@
...
+                        $in.push('tkip');
+                    }
+                });
+                queryData.push = {};
+                queryData.push.$in = $in;
```
iartem, line 1740: What is `{push: {$in: [...]}`?

plugins/push/frontend/public/javascripts/countly.views.js
iartem: it should be smith like ```queryData.$or = dashboardFields.filter(field => messagePlatforms.indexOf(field[2]) !== -1 ).map(field => ({[field]: 1}))```

iartem (changes requested):

plugins/push/frontend/public/javascripts/countly.views.js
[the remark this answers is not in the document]
can-keklik: of course I do :D

plugins/push/frontend/public/javascripts/countly.views.js
LumbardhAgaj: hey Artem! how hard can it be? let's be precise. the reference code that you sent me to build the query used only token prop names, not values. Now, if the reference code is incorrect please let me to which 'fields' are you referring to? 
And, a follow up question to better understand the /dashboard request. Why do you have to filter on dashboard platforms from the platforms of a push notification message?

iartem: Ok, if you want it this way. Original code has [5 fields](https://github.com/Countly/countly-server/blob/master/plugins/push/frontend/public/javascripts/component.push.view.js#L266-L275), you decreased this to 2 without even asking. It should have 4, that's why it's wrong.

> the reference code that you sent me to build the query used only token prop names, not values

Which values are you referring to? `$in` array should have array of field names like `{... push: {$in: ['tkip', 'tkap']}, ...}`. That's what I refer to as "fields" because those are fields of `app_usersXXX` document. A field or token type is well.. token type. It can be iOS Development token, iOS Ad Hoc token, Android token, etc. It's encoded as `"tk" + platform + plarform-specific-token-subtype"`. 

> Why do you have to filter on dashboard platforms from the platforms of a push notification message?

If you look a bit closer you don't filter platforms here. You filter fields to be of those platforms which are set in the message. The logic behind this query is "filter users and return only the ones who have these properties set". You filter users limiting them to only the ones who have at least one of the tokens of mentioned types. The token type list is formed by all possible tokens for given message platforms list. ```messagePlatforms.indexOf(field[2])``` is an expression which checks 3rd symbol in a field name like `tkip` which is a platform.

> Now, if the reference code is incorrect please let me to which 'fields' are you referring to?

The reference code is correct as far as I can tell. I just omitted the fact that main purpose of `tokens` from `/dashboard` is to list token titles supported by plugin. I though that's obvious.

```
var dashboardResponse = ...;
...

var message = ...,
  dashboardFields = Object.keys(dashboardResponse.tokens);
...
queryData.push = {$in: dashboardFields.filter(field => message.platforms.indexOf(field[2]) !== -1 )};
```

LumbardhAgaj: In all fairness, I'm not sure how you can assume that I should know the third char on a token name indicates the platform which it belongs to. 

Anyways, I guess  we want to include all token names for those platforms that the message has. Now, I know that in order to do that I have to check if the third char on a token name is the same as the platform that the message has, e.g.
'tkip', 'tkid' should be added when message platform includes ios ('i' which indicates ios platform is the third char).

Please let me know if there any other concers about the tokens that you are adding when the test property is true (from the code reference that you shared). As you are aware we do not have the testing property anymore.

iartem: There's nothing else

PR #3079, opened 2022-04-15
Feature/client logger

LumbardhAgaj:
Create logger service that will be used by any plugin to log any information which is deemed important.
Create indexed db service that uses localforage library to store information in the client side.
All client logs are saved per user session. A session implies the duration until the user refreshes the browser page.
Update push notification plugin to use logger service.


ar2rsawseen:
Seems to be a duplicate of https://github.com/Countly/countly-server/pull/3080

PR #3081, opened 2022-04-15
Feature/countly client api

LumbardhAgaj:
Introduce countly client api object that is used as a data communication service. As a consequence, all requests for each plugin are stored in the given object.

Note: this branch picks up changes of client logger pull request to avoid any conflicts on the push notification model changes.


ar2rsawseen:
For united request client we cannot store plugin endpoints in the core

PR #3082, opened 2022-04-15
Feature/use countly api in push notification

LumbardhAgaj:
Remove countly api object from push notification model. Instead, use the global client api object to make server requests.


ar2rsawseen:
So PR actually contains a bit more from other PRs
but main point in this one is from: frontend/express/public/javascripts/countly/countly.api.js
that combines API calls in single service/object

I think we still need to decide on how we want to approach that, mostly because plugins are extendable and their code should not be in the core

PR #3161, opened 2022-05-10
Tests for active users.

pnrgenc:


plugins/active_users/tests.js
```diff
@@ -0,0 +1,85 @@
...
+var should = require('should');
+var testUtils = require("../../test/testUtils");
+request = request.agent(testUtils.url);
+
+var APP_KEY = "";
+var API_KEY_ADMIN = "";
+var APP_ID = "";
+
+describe('Testing Active Users', function() {
+    describe('Reset app', function() {
```
Cookiezaurs, line 11: Resetting should be done as almost last one.
You call reset. And then in the next step make sure it is reseted

plugins/active_users/tests.js
```diff
@@ -0,0 +1,85 @@
...
+        it('should reset data', function(done) {
+            API_KEY_ADMIN = testUtils.get("API_KEY_ADMIN");
```
Cookiezaurs, line 13: You don;t need to do it for each test. Just do it once on the first one.
(Reassigining API_KEY_ADMIN + APP_ID values)

PR #3220, opened 2022-05-27
Ser 16 custom variables on push provided by api

iartem:


plugins/push/frontend/public/javascripts/countly.models.js
```diff
@@ -841,6 +846,20 @@
             isAdjacentIndex: function(previousIndex, currentIndex) {
                 return (parseInt(currentIndex) - 1) === parseInt(previousIndex);
             },
+            decodeHtml: function(str) {
```
iartem, line 849: @ar2rsawseen how secure is that?

iartem (commented):
@ar2rsawseen please check that

ar2rsawseen (approved):

PR #3228, opened 2022-05-31
Dashboards changes

kanwarujjaval:
[SER-75]
[SER-74]
[SER-76]

[SER-75]: https://countly.atlassian.net/browse/SER-75?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SER-74]: https://countly.atlassian.net/browse/SER-74?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SER-76]: https://countly.atlassian.net/browse/SER-76?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/dashboards/frontend/public/javascripts/countly.views.js
```diff
@@ -1340,91 +1118,6 @@
                     this.updateWidgetGeography(wId, {size: size, position: position});
                 }
             },
-            autoPosition: function(allWidgets) {
```
osoner, line 1343: @kanwarujjaval do you think we'll need something to handle customers who are already using the new UI?

kanwarujjaval: I don't think there is a direct way of doing that since,
In the 4 column grid system, total sum of x coordinates of all widgets can be maximum 6, but now we are reverting to 12 columns, so there is not definite way of telling if it is old ui, or the new UI.
Will test some cases and get back to you

ar2rsawseen (approved):

PR #3240, opened 2022-06-03
revenue widget api level changes to fetch the data

shaharyar-shamshi:


ar2rsawseen:
Hello,
let's not put plugin-specific code in the core, but keep it in the plugin. I think you can just reuse the same code inside the plugin and that's it :)

PR #3241, opened 2022-06-03
[SER-131] Extra day on graphs for the selected date range

pnrgenc:


frontend/express/public/javascripts/countly/vue/components/date.js
```diff
@@ -935,11 +945,24 @@
                 );
             },
             handleConfirmClick: function() {
+                if (this.rangeMode === 'inBetween') {
+                    var _minDate = new Date(this.minDate).getTime();
+                    var _maxDate = new Date(this.maxDate);
```
Cookiezaurs, line 950: Shouldn't it also have .getTime()?
I tried on your server - it is object. So check in line 952 will not work.

Cookiezaurs (approved):

PR #3268, opened 2022-06-16
[SER-183] [core] Update select-x

widatama:
Update the condition for showing search field in select-x component

### DIFF
--- a/frontend/express/public/javascripts/countly/vue/components/input.js
+++ b/frontend/express/public/javascripts/countly/vue/components/input.js
@@ -457,7 +457,8 @@
         },
         data: function() {
             return {
-                activeTabId: null
+                activeTabId: null,
+                initialOptionsCount: 0
             };
         },
         computed: {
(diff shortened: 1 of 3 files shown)


frontend/express/public/javascripts/countly/vue/components/input.js
```diff
@@ -811,6 +813,25 @@
                     return true;
                 }
                 return Array.isArray(this.innerValue) && this.innerValue.length >= this.minItems && this.innerValue.length <= this.maxItems;
+            },
+            isSearchShown: function() {
+                if (this.searchable) {
+                    if (this.remote) {
```
kanwarujjaval, line 819: We can wait for remote data to resolve, and then mark searchable as true/false.

widatama: That is actually already reflected in options/flatoptions.
The problem here is that we need to keep the initial amount of options. I'll see what I can do with that

widatama: Updated

PR #3272, opened 2022-06-17
[SER-179] Dashboard widget drawer select

widatama:


plugins/times-of-day/frontend/public/javascripts/countly.views.js
```diff
@@ -136,7 +136,7 @@
                     isPluginWidget: true,
                     apps: [],
                     data_type: "",
-                    events: [],
+                    events: "",
```
kanwarujjaval, line 139: Should this be string in empty state?

widatama: This warning came out if events is an array
Screenshot-20220620144605-1920x1052

PR #3286, opened 2022-06-24
[SER-118]fetched the data of the push messages table directly from server side

hilalozturk1:


Cookiezaurs (changes requested):
Needed changes described in slack

Cookiezaurs (approved):

PR #3330, opened 2022-07-07
[SER-30] record operation id in long runni

fatihtrkmen90:


api/utils/taskmanager.js
```diff
@@ -64,10 +67,92 @@ taskmanager.longtask = function(options) {
...
+        catch (err) {
+            console.log('error saving mongo operation id');
+        }
+    }
+
     /** switching to long task */
     function switchToLongTask() {
         timeout = null;
         exceeds = true;
+        options.comment_id = crypto.randomUUID();
```
Cookiezaurs, line 154: We are getting comment_id from drill. Why do you need this line?

frontend/express/public/core/report-manager/javascripts/countly.views.js
```diff
@@ -264,6 +268,14 @@
                     return false;
                 }
             },
+            isStopable: function(row) {
```
Cookiezaurs, line 271: Can we make this isStopable based on if we have operation ID?
like 
```
if(row.hasOperationID){
return true;
}
else {
return false;
}

```

api/utils/requestProcessor.js
```diff
@@ -697,6 +697,17 @@ const processRequest = (params) => {
                         });
                     });
                     break;
+                case 'stop':
+                    validateUserForWrite(params, () => {
+                        taskmanager.stopTask({
+                            db: common.db,
```
Cookiezaurs, line 703: You are passing there task_id and operation id.
And as I check taskmaager.stopTask() you are killing operation based on operation id. In that moment you are not checking if it is connected to any task.

So it means if in some evil way I get global admin API key  - I can kill any operation as long as I guess operation id.

I would suggest rewriting inside stopTask a bit.
Can we do it similar like with rerun Task?

1)Check if task exists
https://github.com/Countly/countly-server/blob/master/api/utils/taskmanager.js#L827
2)And only then kill something. And kill based on operation id you read from mongo. Not the one you recieved.

PR #3454, opened 2022-08-24
[SER-219] Continious data population

pnrgenc:


bin/scripts/continious-data-population.js
```diff
@@ -0,0 +1,1183 @@
...
+        if (users[index].hasSession) {
+            if (users[index].eventsInSession > 9) {
+                users[index].eventsInSession = 0;
+                users[index].startSession(template);
+            }
+            else {
+                if (getRandomInt(1, 3) === 1) {
+                    users[index].addEvent(template, getRandomInt(1, template.events.length - 1));
+                    users[index].eventsInSession++;
+                }
```
Cookiezaurs, line 250: We have chosen  this user to have something. No need to filter out:
So I think we can remove that  if (getRandomInt(1, 3) === 1) {

bin/scripts/continious-data-population.js
```diff
@@ -0,0 +1,1183 @@
...
+ *  Path: countly dir
+ *  Command: node continious-data-population.js
+ */
+
+// var date = new Date();
+// Please set below parameters properly before running script.
+var INTERVAL_AS_SECOND = 1; // 60 * 60 * 24 * 30;
+var APP_ID = ""; // Enter your app Id
+var POPULATOR_TEMPLATE_ID = "defaultBanking"; // Enter the custom populator template ID you created or default template _id "{defaultGaming}"
+var SERVER_URL = "https://master.count.ly";
```
Cookiezaurs, line 13: Please put localhost as default here.
"http://localhost";

bin/scripts/continious-data-population.js
```diff
@@ -0,0 +1,1183 @@
...
+    catch (error) {
+        clearInterval(interval);
+        writeMsg('There was an error while checking db parameters', error);
+    }
+}
+
+readDbParameters();
+
+var interval = setInterval(async function() {
+    users = [];
```
Cookiezaurs, line 216: You call this function on each interval. Why are you resetting users on each call?

bin/scripts/continious-data-population.js
```diff
@@ -0,0 +1,1861 @@
...
+        var upperBoundary = Math.floor(userCount / 10);
+
+        var indexes = [];
+        for (var i0 = 0; i0 < upperBoundary; i0++) {
+            indexes.push(getRandomInt(0, users[inputs[z].APP_ID].length - 1));
+        }
+
+        indexes.forEach(function(index) {
+            if (users[inputs[z].APP_ID][index].hasSession) {
+                if (users[inputs[z].APP_ID][index].eventsInSession >= 1) {
```
Cookiezaurs, line 254: Lets keep this like you had:
(users[inputs[z].APP_ID][index].eventsInSession >=9)
So it would mean  at least 9 events in session

bin/scripts/continious-data-population.js
```diff
@@ -0,0 +1,1861 @@
...
+        return [event];
+    };
+
+    this.addEvent = function(template, app, templateId) {
+        var req = {};
+        var events;
+
+        if (!this.isRegistered) {
+            this.isRegistered = true;
+            events = this.getEvent("[CLY]_view", template && template.events && template.events["[CLY]_view"], app.key, templateId).concat(this.getEvent("[CLY]_orientation", template && template.events && template.events["[CLY]_orientation"], app.key, templateId), this.getEvents(4, template && template.events, app.key, templateId));
```
Cookiezaurs, line 1073: I thought having one event here.
this.getEvents(1, template && template.events, app.key, templateId)

bin/scripts/continious-data-population.js
```diff
@@ -0,0 +1,1861 @@
...
+            req.events = req.events.concat(this.getScrollmapEvents(app.key, templateId));
+            req.events = req.events[Math.floor(Math.random() * req.events.length)];
+        }
+        else {
+            events = this.getEvent("[CLY]_view", template && template.events && template.events["[CLY]_view"], app.key, templateId).concat(this.getEvent("[CLY]_orientation", template && template.events && template.events["[CLY]_orientation"], app.key, templateId), this.getEvents(4, template && template.events, app.key, templateId));
+            // events = events.sort(() => 0.5 - Math.random()).slice(eventChance);
+            events = events[Math.floor(Math.random() * events.length)];
+            req = {timestamp: this.ts, events: events};
+        }
+        this.request(req, app.key);
```
Cookiezaurs, line 1087: req.events.length === events count you are adding to session. 
So:
 users[inputs[z].APP_ID][index].eventsInSession+=req.events.length;
 (You could return that value and then sum OR pass  users[inputs[z].APP_ID][index] and add there.
 (Objects are passed by reference)

bin/scripts/continious-data-population.js
```diff
@@ -0,0 +1,1861 @@
...
+
+        indexes.forEach(function(index) {
+            if (users[inputs[z].APP_ID][index].hasSession) {
+                if (users[inputs[z].APP_ID][index].eventsInSession >= 1) {
+                    users[inputs[z].APP_ID][index].eventsInSession = 0;
+                    users[inputs[z].APP_ID][index].startSession(template[inputs[z].APP_ID], app[inputs[z].APP_ID], inputs[z].POPULATOR_TEMPLATE_ID);
+                }
+                else {
+                    users[inputs[z].APP_ID][index].addEvent(template[inputs[z].APP_ID], app[inputs[z].APP_ID], inputs[z].POPULATOR_TEMPLATE_ID);
+                    users[inputs[z].APP_ID][index].eventsInSession++;
```
Cookiezaurs, line 260: How many events do we add here?
Just one or more?

bin/scripts/continious-data-population.js
```diff
@@ -0,0 +1,1861 @@
...
+                }
+            }
+            else {
+                users[inputs[z].APP_ID][index].startSession(template[inputs[z].APP_ID], app[inputs[z].APP_ID], inputs[z].POPULATOR_TEMPLATE_ID);
+            }
+        });
+    }
+    if (intervalCount === 5) { // it should run one time each interval with some users.
```
Cookiezaurs, line 268: Why when intervalCount === 5?
Why can't we  do this at first and after we do that  setInterval()?

pnrgenc: In the populator, this code is called just before the populate is stopped. 
Current User and Today's Change fields in the cohort table are shown as 0 because the user is not created when I call it at the beginning.
I didn't want to do this after setInterval because it is not clear how long the script will run and the user will not be able to see the cohort as long as the script is running.
I thought it would be good to call this function after creating some users to show some numbers there.
But calling this function in the beginning would be fine also, after a certain time the computation for the cohort will work.

bin/scripts/continious-data-population.js
pnrgenc: it changes depending on the condition [here.](https://github.com/Countly/countly-server/pull/3454/files#diff-0c04b3f9535f8f004b58ca5b95264c0e117bbf0152040336f9151559ccd6d3f3R1074) 
If other events are concats to the Events object, more than 1 event can be added.

Cookiezaurs (approved):

PR #3496, opened 2022-09-15
[SUP-16] Fix for 2 factor auth

kanwarujjaval:


ar2rsawseen (changes requested):
Hello,
what would happen if the user does not have update permission for the current app?
Then I think the endpoint would not work.
Instead, I think we need to change the validation on the API side to only validate if the user exists, because then we query the same authenticated user and do modifications for it, so should be enough to just do if user exists in db
the validation method was validateUser I think, which does not require app id

kanwarujjaval:
> 
validateUser makes more sense

PR #3566, opened 2022-10-07
[SER-331] note in dashboards does escaping

John-Weak:


John-Weak:
https://github.com/Countly/countly-server/pull/3582

PR #3576, opened 2022-10-11
SER-280 format duration properly in events overview

John-Weak:


frontend/express/public/javascripts/countly/countly.common.js
```diff
@@ -4091,9 +4091,11 @@
         * Parse second to standard time format
         * @memberof countlyCommon
         * @param {number} second  number
-        * @returns {string} return format "HH:MM:SS"
+        * @param {number} [trimTo=3]  number [1,3]
+        * @returns {string} return format "HH:MM:SS", if trimTo is specified the length of the result is trimmed
+        * @example trimTo = 2, HH:MM:SS result will be trimmed to HH:MM
         */
-        countlyCommon.formatSecond = function(second) {
+        countlyCommon.formatSecond = function(second, trimTo = 3) {
```
ar2rsawseen, line 4098: I think using default as 3 kind of breaks the default behavior and other places that expect full output would not get it anymore?

frontend/express/public/core/events/javascripts/countly.overview.views.js
```diff
@@ -353,7 +370,7 @@
             return {
                 description: CV.i18n('events.overview.title.new'),
                 disabledDatePicker: '1months',
-                monitorEventsLegend: {"show": false},
+                monitorEventsLegend: { "show": false },
```
ar2rsawseen, line 373: Let's try to sync on formatting styles and use eslint in the repo and not do unnecessary changes

ar2rsawseen (changes requested):

frontend/express/public/javascripts/countly/countly.common.js
John-Weak: No, the default 3 will make the function behave as it was doing before the trimTo argument was added.

frontend/express/public/core/events/javascripts/countly.overview.views.js
John-Weak: hmm, strange I do run eslinteslint . --ext .js --fix before commit, and it even passed eslint check of github.
will change it back nonetheless

frontend/express/public/javascripts/countly/countly.common.js
ar2rsawseen: hmm, correct me if I am wrong, but by default, there are possible 5 tiers (year, day, hour, minute, second) so if there are some cases where it expects to return all 5, it would not only get 3 and thus break old behavior?

frontend/express/public/javascripts/countly/countly.common.js
```diff
@@ -4121,7 +4123,14 @@
...
             }
             else {
-                return resultStrings.join(" ");
+                let res = resultStrings[0];
+                if (trimTo > 3 || trimTo < 1) {
+                    trimTo = 3;
+                }
+                for (let j = 1; j < trimTo && j < resultStrings.length; j++) {
+                    res += ' ' + resultStrings[j];
+                }
```
ar2rsawseen, line 4132: I think this loop can be replaced by array.slice which will be more error prone and more understandable, sorry :)

frontend/express/public/javascripts/countly/countly.common.js
John-Weak: yes you are correct, the default should have been 5, I did not consider YY:DD:HH:MM:SS, I will rectify it

kanwarujjaval (approved):

abhisheksingla02 (approved):

kanwarujjaval (approved):

kanwarujjaval (approved):

ar2rsawseen (approved):

PR #3582, opened 2022-10-13
[SER-331] Note in dashboards does escaping fixed

John-Weak:


plugins/dashboards/frontend/public/templates/helpers/widget/title.html
```diff
@@ -1,4 +1,4 @@
 <div class="bu-is-flex bu-is-flex-shrink-1" style="min-width: 0">
-    <h4 class="bu-is-flex-shrink-1 has-ellipsis" v-if="title">{{title}}</h4>
+    <h4 class="bu-is-flex-shrink-1 has-ellipsis" v-if="title" v-html="title">{{title}}</h4>
```
ar2rsawseen, line 2: I think if we use v-html we need to remove the value from the tag itself, in the next versions VueJs will give out an error of "VueCompilerError: v-html will override element children"

ar2rsawseen (changes requested):

ar2rsawseen (approved):

PR #3592, opened 2022-10-17
[SER-330] high-rate-on-0-rating

John-Weak:


plugins/star-rating/frontend/public/javascripts/countly.views.js
```diff
@@ -963,7 +963,7 @@
                 ];
             },
             ratingRate: function() {
-                var timesShown = this.widget.timesShown === 0 ? 1 : this.widget.timesShown;
+                var timesShown = this.widget.timesShown === 0 ? undefined : this.widget.timesShown;
```
kanwarujjaval, line 966: I dont think this key should be undefined.
@abhisheksingla02 can you please suggest changes?

abhisheksingla02: Here instead of making this key undefined , we should check why ratingsCount is not correct.

John-Weak: by the formula logic, I chose to make it undefined so that the default value (0) can be returned by the formula,
but yes ofcourse timesShown cannot be zero if ratingCount is not zero.
Will look into the populator script.

kanwarujjaval (approved):

PR #3599, opened 2022-10-20
[SER-359] Ser 359 do not use convert to capped

coskunaydinoglu:


plugins/logger/install.js
```diff
@@ -14,9 +14,40 @@ pluginManager.dbConnection().then((countlyDb) => {
             function cb() {
                 done();
             }
-            countlyDb.command({"convertToCapped": 'logs' + app._id, size: 10000000, max: 1000}, function(err) {
+
+            countlyDb.command({"listCollections": 1, "filter": {"name": "logs" + app._id }}, function(err, res) {
+                if (err) {
+                    countlyDb.close();
+                    return;
```
ar2rsawseen, line 21: So if we just return and close the database, I think async.js which iterates through apps in the bottom of the script would be stuck. I think we just need to call return cb(); so the loop can still continue trying to process next app
But also we can log the error via console.log(err); to notify user about the error

plugins/logger/install.js
```diff
@@ -14,9 +14,40 @@ pluginManager.dbConnection().then((countlyDb) => {
...
+                else {
+                    //check if collection capped
+                    if (res && res.cursor && res.cursor.firstBatch) {
+                        //collection exists
+                        if (res.cursor.firstBatch > 0) {
+                            if (!res.cursor.firstBatch[0].options.capped) {
+                                countlyDb.command({ "convertToCapped": 'logs' + app._id, size: 10000000, max: 1000 }, function(err) {
+                                    if (err) {
+                                        countlyDb.close();
+                                        return;
```
ar2rsawseen, line 32: Same goes here

ar2rsawseen (changes requested):

PR #3632, opened 2022-11-07
License backend

iartem:


frontend/express/app.js
```diff
@@ -916,6 +931,8 @@ Promise.all([plugins.dbConnection(countlyConfig), plugins.dbConnection("countly_
                 path: countlyConfig.path || "",
                 cdn: countlyConfig.cdn || "",
                 message: req.flash("message"),
+                notify,
+                license,
```
iartem, line 935: @kanwarujjaval license is an error message for frontend. It's empty if we're all good, `license.err.restricted` if instance is locked (only license manager for global admin case) and `license.err.locked` if instance is restricted (prevent login for non global admins case).

plugins/server-stats/api/jobs/stats.js
```diff
@@ -141,6 +141,17 @@ class StatsJob extends job.Job {
                                 });
 
                                 Countly.userData.save();
+
+                                Countly.fetch_remote_config(null, null, (e, conf) => {
+                                    if (err) {
+                                        log.e('Failed to fetch remote config', e);
+                                    }
+                                    else {
+                                        db.collection('plugins').updateOne({_id: 'plugins'}, {$set: {remoteConfig: conf || {}}}).catch(dbe => {
```
iartem, line 150: @kanwarujjaval remote config is fetched during regular stats ping and stored in db, let me know if it won't work

iartem (commented):
Check comments above please

iartem (commented):
Check comments above please

kanwarujjaval (approved):

PR #3671, opened 2022-11-17
Graph notes changes

pnrgenc:
This commit includes the following changes.
- When clicking the close icon in the graph notes tooltip, the tooltip in the main chart does not close fixed.
- Html entity decode problem fixed in "manage note" and drawers
- Add new note feature has been added to the page in the analytics tab.
- Interactive close icon in Tooltip.
- Centering the labels inside the points
- Points in the widget do not fit depending on height is fixed.
- The add notes button on the widgets has been removed and this feature has been added to more options.

### DIFF
--- a/frontend/express/public/core/notes/javascripts/countly.views.js
+++ b/frontend/express/public/core/notes/javascripts/countly.views.js
@@ -58,6 +58,20 @@
                     });
                 }
             },
+            decodeHtml: function(str) {
+                var map =
+                {
+                    '&amp;': '&',
+                    '&lt;': '<',
+                    '&gt;': '>',
+                    '&quot;': '"',
+                    '&#039;': "'",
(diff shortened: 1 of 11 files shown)


frontend/express/public/core/notes/javascripts/countly.views.js
```diff
@@ -68,12 +82,22 @@
                 case 'edit':
                     data.color = {value: data.color};
                     this.drawerSettings.isEditMode = true;
+                    data.note = this.decodeHtml(data.note);
```
Cookiezaurs, line 85: Can you call this here:
https://github.com/Countly/countly-server/blob/master/frontend/express/public/javascripts/countly/countly.common.js#L294
+if something missing - make additions in that common function.

Cookiezaurs (approved):

PR #3708, opened 2022-11-28
[core] Missing logout call

iartem:


kanwarujjaval (approved):

ar2rsawseen:
As a result of pentest, we should not logout on get, but only on the post.
That is why it is removed from there.

Do we use logout from get somewhere in the code?

PR #3736, opened 2022-12-02
member utility prevent modify immutable field

FoysalOsmany:
Fix for create member error cognito Performing an update on the path '_id' would modify the immutable field '_id' MongoServerError: Performing an update on the path '_id' would modify the immutable field '_id'

### DIFF
--- a/frontend/express/libs/members.js
+++ b/frontend/express/libs/members.js
@@ -1067,8 +1067,14 @@ membersUtility.findMembers = async function(query = {}) {
 */
 membersUtility.updateMember = async function(query = {}, data = {}, upsert = true) {
     return new Promise((resolve, reject) => {
-        delete data._id; // update on the path '_id' would modify the immutable field '_id'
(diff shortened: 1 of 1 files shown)


frontend/express/libs/members.js
```diff
@@ -1067,6 +1067,7 @@ membersUtility.findMembers = async function(query = {}) {
 */
 membersUtility.updateMember = async function(query = {}, data = {}, upsert = true) {
     return new Promise((resolve, reject) => {
+        delete data._id; // update on the path '_id' would modify the immutable field '_id'
```
ar2rsawseen, line 1070: Deleting _id would remove this property from the originally passed object too, and the code might rely on having that value.
So instead of simply deleting, we need to do the deep copy and then remove that property from the copy of the object

ar2rsawseen (changes requested):

ar2rsawseen (approved):

PR #3856, opened 2023-01-17
[SER-433] Date property filtering - "in the last" operator

pnrgenc:


api/lib/countly.common.js
```diff
@@ -97,6 +97,8 @@ function getPeriodObject() {
         reqZeroDbDateIds: []
     };
 
+    _period = period || _period;
```
Cookiezaurs, line 100: You can't do it like that.
In _period you have current set period. You will override if if you do that and it could lead to errors.

Instead you should do:
var my_period = period || _period.

And then change code below in this function where it uses _period.

Cookiezaurs (approved):

PR #3890, opened 2023-02-01
[SER-386] update javascript js in codemirror

coskunaydinoglu:


Gruntfile.js
```diff
@@ -79,7 +79,7 @@ module.exports = function(grunt) {
                     'frontend/express/public/javascripts/countly/countly.common.js',
                     'frontend/express/public/javascripts/utils/simpleUpload.min.js',
                     'frontend/express/public/javascripts/utils/jsoneditor/codemirror.js',
-                    'frontend/express/public/javascripts/utils/jsoneditor/javascript.js',
+                    'frontend/express/public/javascripts/utils/jsoneditor/javascript.min.js',
```
ar2rsawseen, line 82: I see there is also one used here:
https://github.com/Countly/countly-server/blob/master/frontend/express/views/dashboard.html#L128

ar2rsawseen (changes requested):

ar2rsawseen (approved):

PR #3935, opened 2023-02-13
SER-491: Error handling for invalid input in dbviewer

ayasayadi1:


plugins/dbviewer/api/api.js
```diff
@@ -276,6 +280,10 @@ var spawn = require('child_process').spawn,
             if (params.qstring.iDisplayLength) {
                 aggregation.push({ "$limit": parseInt(params.qstring.iDisplayLength) });
             }
+            if (!Array.isArray(aggregation)) {
+                common.returnMessage(params, 500, "The aggregation pipeline must be of the type array");
```
Cookiezaurs, line 284: If you do not use **return** it will continue execution.(Do all lines that is after this if.(){.....}.
So you will still have error.

Cookiezaurs (approved):

PR #3963, opened 2023-02-20
Fix: supervisord.conf

kaanklky:


api/config.sample.js
```diff
@@ -85,9 +85,9 @@ var countlyConfig = {
     },
     /**
     * Default proxy settings, if provided then countly uses ip address from the right side of x-forwarded-for header ignoring list of provided proxy ip addresses
-    * @type {array=} 
+    * @type {array=}
     */
-    ignoreProxies: [/*"127.0.0.1"*/],
```
ar2rsawseen, line 90: I think here we still need IP address and not hostname, because we expect it in IP address header in nginx

api/utils/common.js
```diff
@@ -1590,7 +1590,7 @@ common.getIpAddress = function(req) {
                 return ips[i].startsWith(elem);
             });
         }
-        if (ips[i] !== "127.0.0.1" && (!countlyConfig.ignoreProxies || !masks)) {
+        if (ips[i] !== "localhost" && (!countlyConfig.ignoreProxies || !masks)) {
```
ar2rsawseen, line 1593: Similar here we compare it with nginx IP address header

ar2rsawseen (changes requested):

PR #4001, opened 2023-02-27
Add sudo to shellchecks wget command for non-sudo install

Hakandede:


kaanklky:
We found out we never `cd` into `countly` dir so all operations like `wget` would work under current directory; you can call `bash /opt/countly/bin/countly.install.sh` from `/root` thus script tries to download to `/root`. We `cd` into `countly dir` with this PR and we can be sure it will do all operations under a directory which we have all required permissions.

PR #4022, opened 2023-03-08
fix of rhel devtoolset breaks down sudo

Hakandede:


kaanklky:
@Hakandede fix shellcheck.

kaanklky (approved):

PR #4023, opened 2023-03-08
[SER-373] Extended-json-support-for-dbviewer

coskunaydinoglu:


ar2rsawseen (approved):

ar2rsawseen (changes requested):
Hello, I think we should add bson as a separate dependency, in case it changes as subdependency of MongoDB driver

PR #4110, opened 2023-03-30
fix for populator

abhisheksingla02:


ar2rsawseen:
Already merged here: https://github.com/Countly/countly-server/pull/4113

PR #4136, opened 2023-04-03
[SER-567] Showing Omitted Segments in Events > All Events

ayasayadi1:


frontend/express/public/core/events/templates/allEvents.html
```diff
@@ -39,6 +39,21 @@ <h2> {{i18n('events.all.title.new')}}
 			  <el-select v-model="selectedSegment" class="cly-vue-events-all--margin">
 				<el-option v-for="item in availableSegments" :key="item.value" :value="item.value" :label="item.label">
 				</el-option>
+				<el-option-group
```
Cookiezaurs, line 42: We should not show ommited group if there is no ommited segment.
As you add 'ommited' as first - then you always have at least one item

Cookiezaurs (changes requested):

Cookiezaurs (approved):

PR #4140, opened 2023-04-05
[SER-632] Update user and group permission when an app is deleted

widatama:
When an app is deleted, that app id should be removed from user and group permission


ArtursKadikis:
copilot:summary

kanwarujjaval (approved):

PR #4207, opened 2023-05-05
[SER-692] Add rebuild member permission script

widatama:
This pr adds a script that will rebuild member permission.

Previously when a group is deleted or updated, that group member permissions are not updated correctly. This script can be used to repair those incorrect permission.


kanwarujjaval (approved):

ar2rsawseen:
Should this script be added also to the next major upgrade too?
If it needs to be rerun for some users?

PR #4248, opened 2023-05-24
[SER-718] [crashes] Cleanup crashgroup custom field

widatama:
Crashgroup is saving custom field sent by client. When there are too many unique custom fields saved, crashgroup document size gets too big and problems start coming.

To mitigate the problem this pr adds a script that can be run to cleanup custom fields from crashgroup documents. The script will keep 100 custom field key with the most count for each crashgroup document (something like `{ custom: { key1: { id1: 100, id2: 200, …, id100: 20} }, { key2: { id1: 40, id2: 10, …, id100: 5 }}}`).

This pr also adds a background job that runs the script operation every 10 minutes.


widatama:
Moved to next https://github.com/Countly/countly-server/pull/4265

PR #4263, opened 2023-05-30
[SER-689] Delete dashboard widgets if a related Formula is deleted

pnrgenc:
Method that checks for deleted records in widgets has been moved to a common place


pnrgenc:
it has been already included in SER-690 branch.

PR #4293, opened 2023-06-08
[SER-727] rejectUnauthorized option is fixed

coskunaydinoglu:


test/unit-tests/api.utils.countly-request.js
```diff
@@ -0,0 +1,70 @@
+var request = require('../index.js');
```
ar2rsawseen, line 1: I think this would not work, needs to load countly-request here

ar2rsawseen (approved):

PR #4358, opened 2023-07-07
[DEVOPS-86] Add tests for Countly CE 

Hakandede:
Add tests from EE workflow and seperated slack notify job from the build as in EE in mail.yml workflow
Add new script 'prepare.ce.test.sh' to run the prepopulator script and link node to nodejs
Removed 'Copy plugins' && 'Install plugins' steps. Not needed anymore with the absence of the EE plugins.

MongoDB connection issue in run.sh script is fixed by changing the connection string now in the file but in the job step, after container is up

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -97,14 +97,14 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Prepare tests
+        shell: bash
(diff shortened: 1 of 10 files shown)


kaanklky (approved):

kaanklky:
@ar2rsawseen any comments about failing 749 tests?

ar2rsawseen:
Thank you,
let me play with this branch now :D

PR #4372, opened 2023-07-14
Update callPromisedAppMethod to work with new plugin structure

John-Weak:
[fix] callPromisedAppMethod


kanwarujjaval (approved):

plugins/pluginManager.js
```diff
@@ -881,7 +881,7 @@ var pluginManager = function pluginManager() {
             if (methodPromiseCache[method].length > 0) {
                 for (let i = 0; i < methodPromiseCache[method].length; i++) {
                     try {
-                        let ret = methodPromiseCache[method][i][method](params);
+                        let ret = methodPromiseCache[method][i].plugin[method](params);
```
Cookiezaurs, line 884: **Can you please do the same as in this.callMethod function.?**

They both use the same methodPromise Cache.
On pushing:
 methodCache[method].push(plugs[i].plugin);

So there would be no change.

Cookiezaurs (changes requested):
Left comment

kanwarujjaval (approved):

PR #4373, opened 2023-07-14
[SER-687] Updated countly.common.js to include formatTime function

abhisheksingla02:


api/lib/countly.common.js
```diff
@@ -2660,5 +2660,22 @@ countlyCommon.stringIncrement = function(str) {
...
+* @example
+* //outputs Thu, 23 Jun 2022 19:20:25
+* countlyCommon.formatTime(1658673625);
+*/
+countlyCommon.formatTime = function(timestamp) {
+    if (Math.round(timestamp).toString().length === 10) {
+        timestamp *= 1000;
+    }
+    var target = new Date(timestamp);
+    var tooltip = moment(target).format("ddd, D MMM YYYY HH:mm:ss");
```
widatama, line 2676: I think this time format "ddd, D MMM YYYY HH:mm:ss" is too specific to be in common. Is it being used somewhere else?

Maybe this function can be made more generic by enabling it to receive format string and this time format "ddd, D MMM YYYY HH:mm:ss" can be used as default.

kanwarujjaval (approved):

PR #4378, opened 2023-07-17
[SER-740] CMS API calls for data retrieval

ayasayadi1:


api/parts/mgmt/cms.js
```diff
@@ -0,0 +1,204 @@
...
+* @param {params} params - params object
+**/
+cmsApi.clearCache = function(params) {
+    var query = {};
+    if (params.qstring._id) {
+        query = {'_id': {'$regex': `^${params.qstring._id}`}};
+    }
+    common.db.collection('cms_cache').deleteMany(query, function(err) {
+        if (err) {
+            common.returnMessage(params, 500, 'An error occured while clearing CMS cache: ' + err);
```
Cookiezaurs, line 198: You have to put in return or  do if....else.
It does not stop execution.
On error it will run line 198
And then line 200
Then you will have error. (Because you can't return output twice)

api/parts/mgmt/cms.js
```diff
@@ -0,0 +1,204 @@
...
+**/
+cmsApi.getEntries = function(params) {
+
+    if (!params.qstring._id || AVAILABLE_API_IDS.indexOf(params.qstring._id) === -1) {
+        common.returnMessage(params, 400, 'Missing or incorrect API _id parameter');
+        return false;
+    }
+
+    //WIP
+    var query = {'$or': [ {'_id': {'$regex': `^${params.qstring._id}`}}, {'_id': `${params.qstring._id}_meta`} ]};
```
Cookiezaurs, line 143: query can be (If you don't have any additional query)
query  = {"_id":{"$regex":"^"+params.qstring._id+"_"}}
because  {Some_ID}_meta also is matched by criteria.

If you have additional query, then it is:

query  = {"_id":{"$regex":"^"+params.qstring._id+"_"},{"$or":[{Passed rules in params.qstring.query}, {'_id': `${params.qstring._id}_meta`} ]}

api/parts/mgmt/cms.js
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+* Transform and store CMS entries in DB
+* @param {params} params - params object
+* @param {array} data - data array
+* @param {function} callback - callback function
+**/
+function transformAndStoreData(params, data, callback) {
+    let lu = Date.now();
+    if (!data || data.length === 0) {
```
Cookiezaurs, line 73: Can we do this deletion in a bit different way?
Do at first bulk update for each:
https://github.com/Countly/countly-enterprise-plugins/blob/master/plugins/drill/api/parts/data/drill.js#L2791-L2797

Then delete  where {'_id': {'$regex': `^${params.qstring._id}`} && lu <current value.

Cookiezaurs (approved):

PR #4460, opened 2023-08-21
[SER-746] Countly tracking

widatama:
- Check consent before initializing SDK
- Replace device_id with hostname
- Update app key for stats.count.ly


api/jobs/ping.js
```diff
@@ -57,7 +57,7 @@ class PingJob extends job.Job {
                                     timeout: 4E3,
                                     qs: {
                                         device_id: member.email,
```
ar2rsawseen, line 59: Device id should not be email here, but the same as in frontend

api/jobs/ping.js
```diff
@@ -57,7 +57,7 @@ class PingJob extends job.Job {
                                     timeout: 4E3,
```
ar2rsawseen, line 57: On the https://github.com/Countly/countly-server/pull/4460/files#diff-093d3ee8e3c1f466c82d6f8f2a0fa1ff3b4b507a6e3f123b2d6d6cc7ae790cd2R50 We need to check if the frontend country_tracking setting is enabled and not check config file anymore

frontend/express/libs/members.js
```diff
@@ -364,7 +364,7 @@ membersUtility.updateStats = function(member) {
                 timeout: 4E3,
                 qs: {
                     device_id: member.email,
```
ar2rsawseen, line 366: Member email should not be used, we should use the same as on frontend

frontend/express/views/dashboard.html
```diff
@@ -1952,74 +1940,88 @@ <h4><a href="#/analytics/events/key/{{encodeURIComponent this.name}}">{{this.nam
...
             };
 
-            //track pageviews automatically
-            Countly.track_pageview(Countly.getViewName());
+            if (countlyGlobal.countly_tracking) {
+                //initializing countly with params
+                Countly.init({
+                    app_key: "<%= frontend_app || "e70ec21cbe19e799472dfaee0adb9223516d238f" %>",
+                    url: "<%= frontend_server || "https://stats.count.ly/" %>",
+                    device_id: window.location.hostname,
```
ar2rsawseen, line 1948: We should not use location hostname, because it can be different (multiple domain use, etc)
We should use api.domain setting
and if we detect that it has changed, we should do user merging

frontend/express/libs/members.js
```diff
@@ -364,7 +364,7 @@ membersUtility.updateStats = function(member) {
                 timeout: 4E3,
```
ar2rsawseen, line 364: Check on https://github.com/Countly/countly-server/pull/4460/files#diff-a402ac28f1f8ba694491fae7b927df7679142198ce79470e986fba04ae8260b9R349 should include offline check and frontend.country_tracking check, no need for config check anymore

frontend/express/views/dashboard.html
```diff
@@ -1952,74 +1940,88 @@ <h4><a href="#/analytics/events/key/{{encodeURIComponent this.name}}">{{this.nam
...
+                            }
+                        },1);
+                    });
+                }
+
+                Countly.user_details({
+                    "name": countlyGlobal["member"]["full_name"] || "",
+                    "username": countlyGlobal["member"]["username"] || "",
+                    "email": countlyGlobal["member"]["email"]
+                });
```
ar2rsawseen, line 2005: We should not collect these

frontend/express/views/dashboard.html
```diff
@@ -1952,74 +1940,88 @@ <h4><a href="#/analytics/events/key/{{encodeURIComponent this.name}}">{{this.nam
...
-                key:"INSTALL"
-            });
-            <% } %>
+                Countly.userData.set("lastServer", window.location.hostname);
+                Countly.userData.set("lastVersion", "<%= countlyVersion %>");
+                Countly.userData.set("lastEdition", "<%= countlyTypeTrack %>");
+                Countly.userData.set("isTrial", <%= countlyTrial %>);
+                <% if (!offline_mode) { %>
+                    Countly.userData.set("cpus", <%= cpus %>);
+                <% } %>
```
ar2rsawseen, line 2013: We can remove offline check from here

plugins/server-stats/frontend/app.js
```diff
@@ -54,7 +54,7 @@ const { getUserApps } = require('../../../api/utils/rights');
                             timeout: 4E3,
```
ar2rsawseen, line 54: Check on https://github.com/Countly/countly-server/pull/4460/files#diff-aefa6b585d0d6fde5db7e44f9b8dac0a96e995b5c68d9306e33eb084954d9764R12 should be fixed to check offline mode and frontend.countly_tracking setting

plugins/server-stats/frontend/app.js
```diff
@@ -54,7 +54,7 @@ const { getUserApps } = require('../../../api/utils/rights');
...
                             qs: {
                                 device_id: member.email,
```
ar2rsawseen, line 56: device id should be api.domain setting

ar2rsawseen (changes requested):

frontend/express/views/dashboard.html
widatama: How do we detect if the domain has changed?
And how do we do user merging?

widatama:
@ar2rsawseen 
I've updated the domain for device id and the config checks.
How to detect the domain change and do the merging?

ar2rsawseen:
@widatama I think it is more related to SDK team, we can check existing device id is different from what we have in setting, etc so maybe @turtledreams can help
I will create a separate thread for this

widatama:
@ar2rsawseen 
Updated the domain change check.
Also added another pr here https://github.com/Countly/countly-enterprise-plugins/pull/711

ar2rsawseen (approved):

PR #4471, opened 2023-08-23
add getAllEventsSegmentsForApp.js file

LaNinja06:


bin/scripts/export-data/getAllEventsSegmentsForApp.js
```diff
@@ -0,0 +1,46 @@
...
+ * Command: node getallEventsSegmentsForApp.js
+ */
+
+var plugins = require("../../../plugins/pluginManager.js");
+
+plugins.dbConnection("countly").then(function(db) {
+    db.collection("events").find().toArray(function(err, events) {
+        db.collection("apps").find().toArray(function(err, names) {
+
+            if (err) {
```
ar2rsawseen, line 14: Here we need to check for both errors from events and app, so the names or arguments in those callbacks should be different and in the if you would check both with || operator

bin/scripts/export-data/getAllEventsSegmentsForApp.js
```diff
@@ -0,0 +1,46 @@
...
+ * Server: countly
+ * Path: bin/scripts/export-data
+ * Command: node getallEventsSegmentsForApp.js
+ */
+
+var plugins = require("../../../plugins/pluginManager.js");
+
+plugins.dbConnection("countly").then(function(db) {
+    db.collection("events").find().toArray(function(err, events) {
+        db.collection("apps").find().toArray(function(err, names) {
```
ar2rsawseen, line 12: Here from app documents we only use name property, so let's provide it in projection, so we read from disk only that property and not whole document and not consume too much RAM to store whole app object there

ar2rsawseen (changes requested):

bin/scripts/export-data/getAllEventsSegmentsForApp.js
LaNinja06: I modified the code to reflect these changes! :)

bin/scripts/export-data/getAllEventsSegmentsForApp.js
LaNinja06: I've modified the code to specify only the name property! :)

ar2rsawseen (approved):

PR #4499, opened 2023-09-05
[SER-142] Populator need to have examples with symbolication

John-Weak:
image


plugins/populator/frontend/public/javascripts/countly.models.js
```diff
@@ -436,13 +481,11 @@
         };
 
         this.getError = function() {
-            var errors = [];
+            var errors = "";
```
widatama, line 484: Why is this variable changed into a string?
It's no longer used for web platform, but it is still used in other platforms as an array, so it could be confusing if the type is changed.

widatama (approved):

abhisheksingla02 (approved):

PR #4511, opened 2023-09-07
[SER-812] Script to recheck merges

ayasayadi1:


bin/scripts/fix-data/recheck_merges.js
```diff
@@ -0,0 +1,89 @@
...
+        if (err) {
+            return close(err);
+        }
+        //get all drill collections
+        drillDb.collections(function(err, collections) {
+            if (err) {
+                return close(err);
+            }
+            //for each app
+            asyncjs.eachOf(apps, function(app, index, done) {
```
Cookiezaurs, line 26: We need this script to not be the fastest, but slowly work not overwhelming system.
Based on documentation it means that it will run in parallel. We want it to be in series. New app should be processed only when first one is finished:

http://caolan.github.io/async/v3/docs.html#eachOf - states it works like each
http://caolan.github.io/async/v3/docs.html#each states it works in paralel

bin/scripts/fix-data/recheck_merges.js
```diff
@@ -0,0 +1,89 @@
...
+                var app_id = app._id;
+                if (err) {
+                    return done();
+                }
+                //get all users with merges
+                countlyDb.collection('app_users' + app_id).find({merges: {$gt: 0}}).toArray(function(err, users) {
```
Cookiezaurs, line 32: Do not use to array. Use cursor. If they have 10 million users this will read them all in memory

Cookiezaurs (changes requested):
Comments in here and in slack

PR #4518, opened 2023-09-11
[SER-802] [events] Replace Total Duration with Average Duration for charts 

ayasayadi1:


frontend/express/public/localization/dashboard/dashboard.properties
```diff
@@ -1104,6 +1104,7 @@ events.overview.count = Count
 events.overview.sum = Sum
 events.overview.duration = Duration
 events.overview.dur = Dur
+events.overview.avg.duration = Avg. Duration
```
Cookiezaurs, line 1107: Please don't create new one, but use existing one:
https://github.com/Countly/countly-server/blob/master/frontend/express/public/localization/dashboard/dashboard.properties#L635

Cookiezaurs (approved):

PR #4543, opened 2023-09-15
mongodb installation process updated

ihaardik:


kaanklky:
@ihaardik install pipeline has been failed since it says
> mongodb is already installed

so cannot proceed with plugin installations:
https://github.com/Countly/countly-server/actions/runs/6201155467/job/16837262674?pr=4543#step:4:922

kaanklky:
@ihaardik why no IPv6 for RHEL? And I would recommend calling `sudo bash "$DIR/scripts/mongodb.install.sh" configure` instead of handling IPv6 support manually.

ihaardik:
> @ihaardik why no IPv6 for RHEL? And I would recommend calling `sudo bash "$DIR/scripts/mongodb.install.sh" configure` instead of handling IPv6 support manually.

@kaanklky was trying to figure out the failure of pipeline even though mongo was running on ubuntu 20.04, so only performed the changes in ubuntu script and tested the working.
Now that we know the details will test the required changes locally and create a new PR once all tested!

PR #4569, opened 2023-09-26
Update/mongodb install process

ihaardik:


bin/countly.install_rhel.sh
```diff
@@ -94,7 +94,21 @@ npm config set prefix "$DIR/../.local/"
...
 #install mongodb
-sudo bash "$DIR/scripts/mongodb.install.sh"
+if ! command -v mongod &> /dev/null; then
+    echo "mongod not found, installing MongoDB"
+    sudo bash "$DIR/scripts/mongodb.install.sh"
+else
+    echo "MongoDB is already installed" 
+    # Check if the MongoDB service is active or not
+    if sudo systemctl is-active --quiet mongod; then
+        echo "MongoDB service is already running, no action needed."
```
kaanklky, line 104: So if service is already running, you didn't configure for IPv6?

ihaardik: If service is already running, I am assuming they have already installed service using our scripts and IPv6 config would have been handled.

CI was failing because in ubuntu 20.04 (currently used in our CI) it has mongo packages by default, but post ubuntu 22.04 they have discontinued mongodb service in it and it will perfrom fresh install.

can go for installing ipv6 for running service and restart mongo!

ihaardik: Hi @kaanklky 
have Configured mandatory ipv6 check irrespective of it's service running state.
could you help in reviewing the changes

kaanklky (approved):

PR #4598, opened 2023-10-09
[License] Check license at render 

John-Weak:
Check license at render instead of only login.

### DIFF
--- a/frontend/express/app.js
+++ b/frontend/express/app.js
@@ -1049,113 +1049,119 @@ Promise.all([plugins.dbConnection(countlyConfig), plugins.dbConnection("countly_
         else {
             countlyDb.collection('members').findOne({"_id": countlyDb.ObjectID(req.session.uid + "")}, function(err, member) {
                 if (member) {
-                    req.session.cookie.maxAge = countlyConfig.cookie.maxAgeLogin;
-                    var adminOfApps = [],
-                        userOfApps = [],
-                        countlyGlobalApps = {},
-                        countlyGlobalAdminApps = {};
+                    plugins.callPromisedAppMethod('checkMemberLicense', { req, member }).then(licenseCheck => {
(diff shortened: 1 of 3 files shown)


frontend/express/app.js
```diff
@@ -1049,113 +1049,119 @@ Promise.all([plugins.dbConnection(countlyConfig), plugins.dbConnection("countly_
...
-                    if (member.global_admin) {
-                        countlyDb.collection('apps').find({}).toArray(function(err2, apps) {
-                            adminOfApps = apps;
-                            userOfApps = apps;
-
-                            for (let i = 0; i < apps.length; i++) {
-                                if (apps[i].checksum_salt) {
-                                    apps[i].salt = apps[i].salt || apps[i].checksum_salt;
+                    plugins.callPromisedAppMethod('checkMemberLicense', { req, member }).then(licenseCheck => {
+                        common.licenseAssign(req, licenseCheck);
```
widatama, line 1053: Since `licenseCheck` is stored in session, it will persists even when the license is updated. So this `licenseCheck` has to be removed from session if there's no problem with license.

Steps to reproduce the problem
- Add `ret.notify.push([WARNINGS.EXPIRED]);` before returning here https://github.com/Countly/countly-enterprise-plugins/blob/master/plugins/drill/api/parts/data/license.js#L207 (this is to simulate that the license is expired)
- Restart server, refresh dashboard page, there will be notification on the page
- Now remove `ret.notify.push([WARNINGS.EXPIRED]);` (this is to simulate that the license has been renewed)
- Restart server, refresh dashboard page, the notification is still there (it should've been gone)

widatama (changes requested):

kanwarujjaval (approved):

abhisheksingla02 (approved):

widatama (approved):

PR #4721, opened 2023-11-20
[SER-959] Fix localization for systemlogs.preventIPTracking

emreustere:
in <countly>/plugins/plugins/frontend/public/templates/configuration.html at line 50, getLabelName(key) looks up systemlogs.preventIPTracking as the key is preventIPTracking.


Cookiezaurs (changes requested):
discussion in slack

PR #4724, opened 2023-11-22
[SER-955] just_schema optional param for no widgets request

emreustere:


Cookiezaurs (changes requested):
Waiting  for you to replace anync loop with simple loop as discussed.

Cookiezaurs (approved):

PR #4804, opened 2023-12-27
[SER-1053] Basic outline of popover and content ready

emreustere:


emreustere:
Check my server when you have the time please. Also sorting by events doesn't work right now.

Cookiezaurs (changes requested):
Requested changes in slack

Cookiezaurs (approved):

PR #4910, opened 2024-02-19
[SER-1126] updateMember update to ensure permanent _id field in members collection

ekin-burak:


coskunaydinoglu (approved):
As api_key has unique index upsert was causing to create duplicate member

PR #4968, opened 2024-03-08
[SER-1155] [rights] Optimize dbLoadEventsData

widatama:


Cookiezaurs (approved):

kanwarujjaval (approved):

api/utils/rights.js
```diff
@@ -496,50 +495,74 @@ function wrapCallback(params, callback, callbackParam, func) {
...
     **/
-    function getEvents(app, cb) {
+    function getEvents(appColl, cb) {
         var result = {};
-        common.db.collection('events').findOne({'_id': common.db.ObjectID(app._id + "")}, function(err, events) {
-            if (!err && events && events.list) {
-                for (let i = 0; i < events.list.length; i++) {
-                    result[crypto.createHash('sha1').update(events.list[i] + app._id + "").digest('hex')] = "(" + app.name + ": " + events.list[i] + ")";
+        common.db.collection('events').find({'_id': { $in: appColl.appIds }}).toArray(function(err, events) {
+            appColl.appIds.forEach((appId) => {
```
Cookiezaurs, line 515: Lets say you are getting events for 5 apps.
Then in events you have list of 5 documents - one for each app. Lets assume each app has 5 events.

Then here as you do for each APP - it will process whole array for each app  - creating  records for all events(not just ones for this app)

So you would get 5*(5*5) = 125 , not 25 records as you should.

api/utils/rights.js
```diff
@@ -496,50 +495,74 @@ function wrapCallback(params, callback, callbackParam, func) {
...
     **/
-    function getViews(app, cb) {
+    function getViews(appColl, cb) {
         var result = {};
-        common.db.collection('views').findOne({'_id': common.db.ObjectID(app._id + "")}, function(err, viewDoc) {
-            if (!err && viewDoc && viewDoc.segments) {
-                for (var segkey in viewDoc.segments) {
-                    result["app_viewdata" + crypto.createHash('sha1').update(segkey + app._id).digest('hex')] = "(" + app.name + ": " + segkey + ")";
+        common.db.collection('views').find({'_id': { $in: appColl.appIds }}).toArray(function(err, viewDocs) {
+            appColl.appIds.forEach((appId) => {
```
Cookiezaurs, line 552: The same issue as for events

api/utils/rights.js
```diff
@@ -548,15 +571,15 @@ function dbLoadEventsData(params, apps, callback) {
         callback(null, params.member.eventList, params.member.viewList);
     }
     else {
-        async.map(apps, getEvents, function(err, events) {
+        getEvents({ appIds, appNamesById }, function(err, events) {
             var eventList = {};
             for (let i = 0; i < events.length; i++) {
```
Cookiezaurs, line 576: I don''t think we need this  for cycle anymore. Before it was async map, so you got:
[result array1, result array2......]
But now all results are in single array at root. 
The same for views

Cookiezaurs (changes requested):
left some comments

widatama:
Updated

kanwarujjaval (approved):

Cookiezaurs (approved):

PR #4983, opened 2024-03-14
[SER-969] Remove internal sdk methods from heatmap js

widatama:


turtledreams (changes requested):
also havent seen additional logs except 1 place. it would be best to add some logs for future debugging.

widatama:
> also havent seen additional logs except 1 place. it would be best to add some logs for future debugging.

I couldn't find any other failure spot like http request that needs logging. Any suggestions?

plugins/views/frontend/public/heatmap.js
```diff
@@ -6,11 +6,147 @@
...
+        VERBOSE: "[VERBOSE] "
+    };
+
+    /**
+     *  Logging stuff, works only when sdk debug mode is true
+     * @param {string} level - log level (error, warning, info, debug, verbose)
+     * @param {string} message - any string message
+     */
+    function log(level, message) {
+        if (Countly && Countly.debug && typeof console !== 'undefined') {
```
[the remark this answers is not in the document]
widatama: Updated

plugins/views/frontend/public/heatmap.js
```diff
@@ -730,35 +866,24 @@
...
+            var method = "POST";
 
-            if (method === "GET") {
-                xhr.open('GET', Countly.passed_data.url + apiPath + "?" + data, true);
-            }
-            else {
-                xhr.open('POST', Countly.passed_data.url + apiPath, true);
-                xhr.setRequestHeader("Content-type", "application/x-www-form-urlencoded");
-            }
-            xhr.setRequestHeader("countly-token", Countly._internals.getToken());
```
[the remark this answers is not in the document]
widatama: Updated

turtledreams:
> > also havent seen additional logs except 1 place. it would be best to add some logs for future debugging.
> 
> I couldn't find any other failure spot like http request that needs logging. Any suggestions?

Fair point but I think we shouldn't only look for failure points to add logs. It would be valuable to get the logs indicating the inner workings of functions or the app state. like the initial state, what is being called, and maybe what is returned. There are plenty of operations happening but when I open the console only thing I see is "[INFO] [Countly] [Heatmap] [sendXmlHttpRequest] Sending XML HTTP request". There can be calculations happening which are non breaking but make things not function as normal. We would ask for a client to return these logs in the future and anything we add now can be useful later. What do you think?

turtledreams:
I checked your test environment and when I select anything other than 'all' I saw sth like this:
image
why do you think that is? do you think that dropdown is misleading? we can have a call with product maybe to discuss that if you think so.

widatama:
> I checked your test environment and when I select anything other than 'all' I saw sth like this: image why do you think that is? do you think that dropdown is misleading? we can have a call with product maybe to discuss that if you think so.

There could be a problem with the token. It should've return empty when there's no heatmap data
Screenshot-20240315164656
Is the token still exists on that page? See if it's in localStorage `cly_token`
Screenshot-20240315165247

kanwarujjaval (approved):

turtledreams (approved):

ArtursKadikis (changes requested):
This is missing logs for all functions to see in the flow

plugins/views/frontend/public/heatmap.js
```diff
@@ -730,35 +883,24 @@
 
     function sendXmlHttpRequest(params, apiPath, callback) {
         try {
-            Countly._internals.log("Sending XML HTTP request");
+            log(logLevelEnums.INFO, "[sendXmlHttpRequest] Sending XML HTTP request");
```
ArtursKadikis, line 886: let us print the params and apiPath here

plugins/views/frontend/public/heatmap.js
```diff
@@ -730,35 +883,24 @@
...
             xhr.onreadystatechange = function() {
                 if (this.readyState == this.HEADERS_RECEIVED) {
                     try {
-                        Countly._internals.setToken(xhr.getResponseHeader("content-language"));
+                        Countly._internals.removeValueFromStorage('cly_token');
+                        Countly._internals.setValueInStorage('cly_token', xhr.getResponseHeader("content-language"));
                     }
                     catch (ex) {
-                        Countly._internals.log("failed, trying fallback header");
+                        log(logLevelEnums.ERROR, "[sendXmlHttpRequest] Failed, trying fallback header");
```
ArtursKadikis, line 903: This logged message doesn't seem to be correctly describing what is happening here.

plugins/views/frontend/public/heatmap.js
```diff
@@ -768,22 +910,18 @@
                     }
                 }
                 else if (this.readyState === 4) {
-                    Countly._internals.log("Failed Server XML HTTP request", this.status);
+                    log(logLevelEnums.ERROR, "[sendXmlHttpRequest] Failed Server XML HTTP request", this.status);
```
ArtursKadikis, line 913: we should also print the response

ArtursKadikis (changes requested):
Following changes for the log calls in "getDocHeight", "getDocWidth",  "getViewportHeight":
* they should be debug level
* they should print the returned value

Functions that don't have any params shouldn't print " [].slice.call(arguments)"

Let us wrap the "drawData" calls in a try catch and print the errors

widatama:
Updated

kanwarujjaval (approved):

ArtursKadikis (approved):

PR #4994, opened 2024-03-19
[SER-1145] visualize omitted views just like we visualize omited events in events section

emreustere:


Cookiezaurs (changes requested):
Changes discussed in slack

emreustere:
Tested on next at https://next.count.ly/dashboard#/64f86dbd9f16bb8ea13262ba/analytics/views 
Deleting branch.

PR #5204, opened 2024-05-13
[SER-1199] - Close the guides modal with ESC key press

LaNinja06:


pnrgenc (changes requested):
Hi Gia,
Great job, I can only suggest this:
When adding an event listener in vue.js, we usually add it mounted and remove this listener before destroying. This way we can prevent possible memory leaks.
You may get better results if you try the following way
       ```
  mounted: function() {
             document.addEventListener('keydown', this.handleEscapeKey);
         },
         destroyed: function() { // or beforeDestroy
             document.removeEventListener('keydown', this.handleEscapeKey);
         },
   methods: {
     handleEscapeKey: function(event) {
                 if (event.keyCode === 27 || event.key === 'Escape') {
                     this.onClose()
                 }
             }
         }
```
   If you're curious, life cycles in vue.js: https://v2.vuejs.org/v2/guide/instance

LaNinja06:
Thank you for the suggestions, Pinar! I modified it as mentioned. Now, it looks more aligned with the existing code. :)

pnrgenc (approved):

PR #5234, opened 2024-05-20
[SER-1481] error in dbviewer for collection name containing $

pnrgenc:


plugins/dbviewer/frontend/public/javascripts/countly.views.js
```diff
@@ -212,9 +214,13 @@
                     if (force) {
                         this.isLoading = true;
                     }
-                    this.tableStore.dispatch("fetchDbviewerTable", {_silent: !force}).then(function() {
-                        self.isLoading = false;
-                    });
+                    if (force || !this.isFetching) {
```
pnrgenc, line 217: I did not change this line (isFetching flag logic) due to it is included in [next branch](https://github.com/Countly/countly-server/blob/next/plugins/dbviewer/frontend/public/javascripts/countly.views.js#L222)
Also without this, it is giving same error recursively

Cookiezaurs (approved):

PR #5239, opened 2024-05-21
[SER-1492] alert edit page special character problem for alert name

ekin-burak:
Special character problem in condition row was solved in another task, but on alert edit page, problem was continuing. It is fixed too with this branch.


ekin-burak:
This PR was closed and moved to task SER-1579 on JIRA.

PR #5287, opened 2024-05-28
[SER-1521] Removed  break code  from navigateToWidgetsDetailPage method 

can-angun:
https://countly.atlassian.net/browse/SER-1521


savascountly (approved):
Good job

PR #5291, opened 2024-05-29
[SER-1160] make the countly guides icons stand out until first use

emreustere:


emreustere:
I need to ask about sidebar opening bug

emreustere:
POST https://emre.count.ly/i/guides/viewed 400 (Bad Request) warning on console when clicking top guides button, also the api.js file feels wrong even though it works.

Cookiezaurs (approved):

PR #5312, opened 2024-06-06
Update crash app version list script

widatama:


bin/upgrade/24.05/scripts/generate_crashgroup_app_version_list.js
```diff
@@ -3,94 +3,62 @@
...
+                else {
+                    done();
+                }
+            });
         }
-    }
-
-    countlyDb.close();
-    console.log('Crashgroup data update done');
+        asyncjs.eachSeries(apps, update, function() {
```
kanwarujjaval, line 59: We dont need asyncjs,
we can just use native promises or async/await, better performance too

kanwarujjaval (approved):

PR #5321, opened 2024-06-10
[SER-1573] old data remains when the alerts metric is updated

ekin-burak:
the bug which is occured when updating the alerts trigger part is fixed


can-angun:
TESTED 
ready for merge  image

image

Thanks for your effort @ekin-burak image

FYI @coskunaydinoglu

coskunaydinoglu (approved):

PR #5353, opened 2024-06-20
[SER-1591] [server-stats][license] Add monthly data points

widatama:


plugins/server-stats/api/parts/stats.js
```diff
@@ -241,6 +241,35 @@ function fetchDatapoints(db, filter, options, callback) {
             return callback(toReturn);
         }
 
+        if (options.monthlyBreakdown) {
```
kanwarujjaval, line 244: Please add the  params monthlyBreakdown and license_hosting  to jsdoc definition of the method

widatama: Updated

kanwarujjaval (approved):

PR #5397, opened 2024-07-08
Update version.info.js

keithlammers:
Version info in 24.05.3 was still showing as 24.05.2.


keithlammers:
No longer needed as 24.0.5.4 is available now.

PR #5419, opened 2024-07-19
[SER-1555] Hooks insufficient null checks

umtcntp:


coskunaydinoglu:
@umtcntp there is a lint failure

cihadtekin (approved):

PR #5431, opened 2024-07-25
[QT-223] log out than log in with created user after onboarding 

can-angun:
https://countly.atlassian.net/browse/QT-223


savascountly (approved):
Good job!

PR #5442, opened 2024-07-30
[Remote-Config] Tests

John-Weak:


plugins/remote-config/api/api.js
```diff
@@ -370,7 +415,7 @@ plugins.setConfigs("remote-config", {
         parameter.conditions = parameter.conditions || [];
         processParamValue(parameter);
 
-        var maximumParametersAllowed = plugins.getConfig("remote-config").maximum_allowed_parameters;
+        var maximumParametersAllowed = params.qstring.isTest ? 4 : plugins.getConfig("remote-config").maximum_allowed_parameters;
```
widatama, line 418: It's not a good idea to add a parameter just for testing.

The config value can be changed from the test file before the request is made https://github.com/Countly/countly-enterprise-plugins/blob/next/plugins/crashes-jira/tests.js#L62

plugins/remote-config/api/api.js
```diff
@@ -839,9 +910,12 @@ plugins.setConfigs("remote-config", {
 
         var collectionName = "remoteconfig_parameters" + appId;
 
+        //TODO:validations
```
widatama, line 913: Is this related to one of the bug tickets?

plugins/remote-config/tests/parameter-crud/update-parameter.js
```diff
@@ -0,0 +1,183 @@
...
+                        condition_name: 'non_existent_condition',
+                        condition_value: 'some_value'
+                    }],
+                    status: 'Running',
+                    expiry_dttm: null,
+                }),
+            })
+        //.expect(400);
+            .expect(200);
+            //TODO:FIX
```
widatama, line 114: Can this test run successfully? Why does it have to be fixed?

plugins/remote-config/tests/index.js
```diff
@@ -1,4 +1,20 @@
-require('./add-remote-config.js');
-require('./fetch_remote_config.js');
-require('./condition-endpoints.js');
-require('./remote-config-endpoint.js');
\ No newline at end of file
+const plugins = require('../../pluginManager.js');
+
+try {
+    if (plugins.isPluginEnabled('ab-testing')) {
+        require('./ab_rc-method.js');
```
widatama, line 5: Let's just move these tests to ab testing plugin to avoid using conditionals here

There're tests for fetch remote config in ab testing plugin https://github.com/Countly/countly-enterprise-plugins/blob/ab-testing-cycle9/plugins/ab-testing/tests/fetch_remote_config.js

plugins/remote-config/tests/ab_rc-method.js
```diff
@@ -0,0 +1,363 @@
...
+            .expect('Content-Type', /json/).expect(200);
+
+        await request
+            .get('/i/ab-testing/add-experiment')
+            .query({
+                api_key: API_KEY_ADMIN,
+                app_id: APP_ID,
+                app_key: APP_KEY,
+                experiment: JSON.stringify({
+                    name: 'fetch_remote_config_ab_testing',
```
widatama, line 72: Can this experiment be named specific to this test file?
It's to avoid clashing with the test for fetch remote config endpoints

widatama (changes requested):

plugins/remote-config/api/api.js
kanwarujjaval: Yes, and kinda no, Its more of a reminder for circling back to this later

plugins/remote-config/tests/parameter-crud/update-parameter.js
John-Weak: we should ideally be getting 400 but instead we are getting 200.

PR #5467, opened 2024-08-08
feat: adding display mode customization to datatable component

pnrgenc:
Other plugins that use this mod have been updated to be customized with this prop


kanwarujjaval (approved):

pnrgenc:
duplicated this one: https://github.com/Countly/countly-server/pull/5478

PR #5499, opened 2024-08-22
[SER-1547] Helpers. Loop with splice should be rewriten

YavuzYilmazz:


cihadtekin (commented):
could be better with just `array.filter(item => item !== "")` instead of starting from end

cihadtekin (approved):

PR #5500, opened 2024-08-22
[SER-1548] [Deepscan] Token-manager. Loop with splice should be rewriten to preven out of bounds error

YavuzYilmazz:


coskunaydinoglu (commented):
@YavuzYilmazz Wouldn't it cause infinite loop since you are decrementing j while it is incremented in for loop?

YavuzYilmazz:
> @YavuzYilmazz Wouldn't it cause infinite loop since you are decrementing j while it is incremented in for loop?

@coskunaydinoglu I think this wouldn't be the problem here. In fact, the suggestion was to add break, but I thought it wouldn't work because the array size would also change. But I can add break if you think there is a problem

coskunaydinoglu:
In fact this j-- does not have any affect since it is already looped. @YavuzYilmazz I think this is false positive. Array length is already checked. With j-- or without it works correctly

YavuzYilmazz:
> In fact this j-- does not have any affect since it is already looped. @YavuzYilmazz I think this is false positive. Array length is already checked. With j-- or without it works correctly

Yes that's true, it could be a false positive situation. Let me try to check this, I need to look at examples.

cihadtekin (approved):
better to use `array.prototype.filter` next time

PR #5505, opened 2024-08-26
[SER-1558] [Deepscan] Mix 2. Fix insufficient null checks

YavuzYilmazz:


frontend/express/public/core/events/javascripts/countly.details.models.js
cihadtekin, line 1: - inside getAllEventsList, it seems like we cannot discard the possibility to have eventsList empty but groupList with some data. possible fix: instead of the early return, move "var map ...." inside the if block.
- we don't need to null check "context" inside fetchRefreshAllEventsData. every vuex action will have the first argument as "context"

api/utils/taskmanager.js
cihadtekin, line 1: since you already added a default for "options", you don't need to check for its existence again in the if block

cihadtekin (changes requested):
requested some changes per file. please update the files accordingly

cihadtekin (approved):

PR #5506, opened 2024-08-26
[SER-1557] [Deepscan] Mix 3. Fix insufficient null checks

YavuzYilmazz:


plugins/hooks/api/parts/triggers/api_endpoint.js
cihadtekin, line 1: deconstruction for "paths" indicates that "params" need to be an object, not an array

plugins/crashes/api/api.js
cihadtekin, line 1: this may crash when there's no dbAppUser (or not, because of hoisting inside the function scope, depends on the usage of those variables created in if block). the variables created inside if block shouldn't be used out of the scope

cihadtekin (changes requested):
requested some changes per file. please update the files accordingly

cihadtekin (approved):

PR #5550, opened 2024-09-12
Content builder

kanwarujjaval:


frontend/express/public/stylesheets/vue/clyvue.scss
```diff
@@ -4186,6 +4186,8 @@
 .cly-vue-content-builder {
     position: absolute;
     z-index: 2001;
+    // z-index: 2006;
+    // z-index: 9999999;
```
gabrieloliveirapinto, line 4190: If we do not need this code, we should remove it.

Not a blocking change.

api/utils/countlyFs.js
```diff
@@ -316,6 +316,60 @@ countlyFs.gridfs = {};
...
+    /**
+     * Update file fields by ID
+     * @param {string} category - collection where the file is stored
+     * @param {string} id - file id
+     * @param {object} updateFields - fields to update
+     * @param {function} callback - function called when updating was completed or errored, providing error object as first param
+     * @example
+     * countlyFs.gridfs.updateFileById("test", "66d6c2d770434130c03b7dae", { filename: "newname.png", "metadata.tags": "newtag" }, function(err){
+     *   console.log("Update finished", err);
+     * });
```
gabrieloliveirapinto, line 328: If we do not need this code, we should remove it.

Not a blocking change

gabrieloliveirapinto (approved):

PR #5584, opened 2024-09-19
[SER-1922] add audit logs to hook create edit delete actions

umtcntp:


ArtursKadikis (changes requested):
needs to pass linter

ArtursKadikis (changes requested):
We want to use localization for our audit log entries. 

For that to happen properly, you need to specify localization in the following form:
localization in this form: systemlogs.action.{ACTION}

An example of that can be seen here:
https://github.com/Countly/countly-server/blob/master/plugins/views/frontend/public/localization/views.properties#L67-L68

In your case and "action" is smth like "hook_updated". So you would have the localization entry with the following key:
systemlogs.action.hook_updated

This will then allow it to be automatically be picked up by the system.

Let us also add the "_id" for both creation and updating. That is useful debugging information.

We can also remove the audit entry for "hook_list_viewed". This would not be relevant to the audit log and too spammy.

umtcntp:
The audit logs used are kept in params. The params object holds the details of the operation performed on the hook (updated, created, status updated). The example of hook_created log look like;

 "created": {
      "_id": {
        "$oid": "66ebff11cca4c7bb8a60cdcf"
      },
      "name": "Hello",
      "description": "ff",
      "apps": [
        "65f97d5e0abccaceec50f9d8"
      ],
      "trigger": {
        "type": "IncomingDataTrigger",
        "configuration": {
          "event": [
            "65f97d5e0abccaceec50f9d8****"
          ],
          "filter": "{\"dbFilter\":{}}"
        }
      },
      "createdBy": {
        "$oid": "65f97d56559a0a4974db780a"
      },
      "createdByUser": "",
      "effects": [
        {
          "type": "CustomCodeEffect",
          "configuration": {
            "code": "r"
          }
        }
      ],
      "enabled": true,
      "created_at": 1726742289847
    }

ArtursKadikis (changes requested):
Let us change "update", "created". Let us add comments for example values that would be in those variables

PR #5611, opened 2024-09-30
[QT-227] Verify that the Community Edition screens that includes data, can be opened as error-free - VOL 4

can-angun:
https://countly.atlassian.net/browse/QT-227


savascountly (approved):
Good job!

PR #5612, opened 2024-09-30
[SER-1635] alert testing edit columns icon is missing on alert home page

ekin-burak:


cihadtekin (approved):
column ordering is not working. but its not something related to alerts, global component needs to be fixed afaik

PR #5629, opened 2024-10-04
[dashboards] Update code for latest mongo driver

John-Weak:
Before mogo driver update we used to recieve object like this:

image

After update:
image


widatama:
Please add the tests for update and remove widget

PR #5651, opened 2024-10-08
[SER-1556] Frontend mix 1. Fix insufficient null checks

YavuzYilmazz:


api/parts/data/fetch.js
cihadtekin, line 1: there should be a "returnOutput" kind of function ran here even if there was an error (eg: check the body of fetch.fetchJobs). but this check can be completely removed by just removing the first "?" in front of the "params" inside if condition, since params is required in fetchTimeObj and is not optional anyway

frontend/express/public/core/events/javascripts/countly.details.models.js
cihadtekin, line 1: context is not optional here

cihadtekin (changes requested):
please update the pr according to the comments

cihadtekin (approved):

PR #5676, opened 2024-10-14
[QT-232] Verify that the Community Edition screens that includes data, can be opened as error-free - VOL 5

can-angun:
https://countly.atlassian.net/browse/QT-232


savascountly (approved):
You always do amazing job!

PR #5698, opened 2024-10-22
[SER-1550] API common. Fix insuficient null checks

YavuzYilmazz:


api/utils/common.js
cihadtekin, line 1: I think with this update, if params is not passed to the function, request will just hang since there's nothing ending "res".
to me, completely removing the nullcheck for params from the if condition would be the best approach here

cihadtekin (changes requested):
please update according to the comment on file

PR #5732, opened 2024-11-05
[SER-2118] Improve script to fetch event deletion list from countly.events and drill_meta

ayasayadi1:


Cookiezaurs (changes requested):
Clear out duplicates from event list before processing

Cookiezaurs (approved):

PR #5740, opened 2024-11-07
[SER-2116] Show a flex banner on sidebar for CE

ayasayadi1:


Cookiezaurs (changes requested):
Banner goes over sidebar menu covering some options.

Cookiezaurs (approved):

PR #5756, opened 2024-11-11
[QT-235] Fix error: The Chrome Renderer process just crashed.

can-angun:
https://countly.atlassian.net/browse/QT-235


savascountly (approved):
Good job!

PR #5759, opened 2024-11-12
Add session check to formdata

John-Weak:


frontend/express/app.js
```diff
@@ -603,6 +603,10 @@ Promise.all([plugins.dbConnection(countlyConfig), plugins.dbConnection("countly_
     app.use(function(req, res, next) {
         var contentType = req.headers['content-type'];
         if (req.method.toLowerCase() === 'post' && contentType && contentType.indexOf('multipart/form-data') >= 0) {
+            if (!req.session?.uid || Date.now() > req.session?.expires) {
```
kanwarujjaval, line 606: wouldn't it be better to use checkRequestForSession? method also handles frontend redirects and extendSession?

John-Weak: we can use [checkRequestForSession](https://github.com/Countly/countly-server/blob/johnWeak-master/frontend/express/app.js#L772) if we change it to accept a callback function,
but since it is a middleware function I decided not to do that.

Cookiezaurs (approved):

frontend/express/app.js
John-Weak: We explored alternate fix as well (rearranging middlewares),
but it would need more testing since alot of things might break, so the current way is the best way to solve it quickly since it also checks for valid session

ArtursKadikis (approved):

PR #5825, opened 2024-12-02
Enabled remote config to get survey id from remote config

pnrgenc:


pnrgenc:
will be added when initializing inside the plugin

Issue #1556, opened 2020-10-05
No easy way to access the image binary addresses for symbolicating...


spicymatt:
FIrst of all, thanks a lot for Countly and the quality of your product.

### Expected Behavior
We would like to see the binary addresses for the app and packages in the crash report page.
You have the info in the database but it is hard to access.

### Current Behavior
When examining a crash report in the Community edition, there is no easy way (at least we did not find one) to access the image binary load addresses.
As a result, it is very cumbersome to perform symbolication. 
The load addresses seem only accessible using the Tools > DB Viewer. It means that we need to identify the crash report in the DB Viewer which is not easy, and then decipher the JSON that is there.

### Possible Solution
DIsplay the load addresses in a friendly manner using the info that you stored.

### Your Environment

* Community edition

ar2rsawseen:
Hello
I assume this is about the iOS part specifically?
So previously we used our own crash reporting there, but with the latest iOS SDK version, we moved to PLCrashReport.
And now when you go to a specific crash inside Countly dashboard, there are three dots (...) right above the stack trace, it will allow you to download the whole PLCrash report, which will contain lots of different information, including load addresses and binaries.

Hope that helps :)

spicymatt:
hello,
Thanks for the answer. It was a MacOS question.
We looked at PLCrashReport and did the integration. It works fine and solves our issue.
Thanks.

PR #5848, opened 2024-12-06
Update CHANGELOG.md (#1556), [fix] logoType default is not selectable

ArtursKadikis:


ArtursKadikis:
countly-enterprise-plugins#1556

ArtursKadikis:
Countly/countly-enterprise-plugins#1556

PR #5856, opened 2024-12-10
Script to remove 2FA from user(s)

John-Weak:


bin/scripts/member-managament/disable_2fa.js
```diff
@@ -0,0 +1,61 @@
...
+        console.log(JSON.stringify(users));
+        if (!dry_run) {
+            await Promise.all(users.map(async(user) => {
+                let userId = user._id;
+                await countlyDb.collection("members").findAndModify(
+                    {_id: userId},
+                    {},
+                    {
+                        $set: {"two_factor_auth.enabled": false},
+                        $unset: {"two_factor_auth.secret_token": ""}
```
widatama, line 34: Since the update is the same for all users, I think it's better to do it in one run instead of in a loop. The updated document is also not needed afterwards, so it's better to use updateMany instead.

For example, the db operation could be something like this
`db.collection('members').updateMany({_id: {$in: ['id1', 'id2']}}, {$set: {...}})`

widatama (changes requested):

widatama (approved):

PR #5874, opened 2024-12-18
[QT-262] Fixed console error - tag <template> has no matching end tag.

can-angun:
https://countly.atlassian.net/browse/QT-262


savascountly (approved):
Good job!

PR #5895, opened 2025-01-09
feat: [SER-2215] Adding disable feature to dialog box buttons

pnrgenc:


pnrgenc:
https://github.com/Countly/countly-server/pull/5925 Moved to merge with another branch

frontend/express/public/javascripts/countly/vue/components/dialog.js
```diff
@@ -76,6 +78,17 @@
             },
             customClass: function() {
                 return this.alignCenter ? "el-dialog--centered" : "";
+            },
+            isButtonDisabled: function() {
+                return {
+                    save: this.saveButtonDisabled,
+                    cancel: this.cancelButtonDisabled
+                };
+            },
```
gabrieloliveirapinto, line 87: Why do you converted both props into a object?

In my opinion, this can lead to errors in the future, because when seeing `isButtonDisabled` the developer may expect a boolean and in reality is getting an object.

I suggest using the props directly and removing this

frontend/express/public/javascripts/countly/vue/components/dialog.js
```diff
@@ -76,6 +78,17 @@
...
+            buttonStyle: function() {
+                return function(disabled) {
+                    return disabled ? { opacity: 0.5 } : {};
+                };
```
gabrieloliveirapinto, line 91: You are using a computed property to return a method, this is bad practice.

It should be moved to methods, for example:

```
  getButtonStyle: function(disabled) {
    return disabled ? { opacity: 0.5 } : {};
  } 
```

PR #5900, opened 2025-01-10
[SER-2264] Cannot load event data when there's an escaped character in all events list

cihadtekin:


John-Weak (approved):

frontend/express/public/core/events/javascripts/countly.details.models.js
```diff
@@ -886,6 +886,9 @@
                 return countlyAllEvents.service.fetchAllEventsData(context, period)
                     .then(function(res) {
                         if (res) {
+                            if (Array.isArray(res.list)) {
+                                res.list = res.list.map(eventName => _.unescape(eventName).replace(/&#39;/g, "'"));
```
John-Weak, line 890: we can also use `countlyCommon.unescapeHtml` it already contains `(/&#39;/g, "'")`

cihadtekin: done. thanks

John-Weak (approved):

ArtursKadikis (approved):

PR #5933, opened 2025-01-20
feat: [SER-2215] Adding disable feature to dialog box buttons #5925

pnrgenc:


pnrgenc:
will be merged journey+content-main into journey+content-next later, so this is unnecessary

PR #5935, opened 2025-01-21
server close before flush

kanwarujjaval:
* close server before we begin flushing writes, so we stop listening to new requests and potentially new insertions get queued in the async queue
* settimeout 0 for exit to put the exit at the end of i/o queue


gabrieloliveirapinto (approved):

kanwarujjaval:
Will not be part of https://github.com/Countly/countly-server/pull/5912

PR #5936, opened 2025-01-21
SER-1711: Component changes to support version management in journey

pnrgenc:


frontend/express/public/javascripts/countly/vue/templates/content/content-body.html
```diff
@@ -1,6 +1,13 @@
 <div class="cly-vue-content-builder__layout-main__body" :class="{ 'bu-is-justify-content-flex-end': hideLeftSidebar, 'bu-is-justify-content-flex-start': hideRightSidebar, 'cly-vue-content-builder__layout-main__body__dashed-bg': hasDashedBackground }" :style="{'background-color': backgroundColor}">
     <transition :name="toggleTransition" @enter="onViewEntered">
-        <div v-if="!hideLeftSidebar && (collapsible && !isCollapsed)" class="cly-vue-content-builder__layout-main__body__left-sidebar">
+        <div v-if="!hideLeftSidebar" class="cly-vue-content-builder__layout-main__body__left-sidebar" :style="{ 'width': this.leftSidebarWidth + 'px !important'}">
```
gabrieloliveirapinto, line 3: Why do we need the `!important` here?

Just asking becuse inline styles take precendence from the styles in stylesheet so `!important` is redundent

frontend/express/public/javascripts/countly/vue/templates/content/content-body.html
```diff
@@ -1,6 +1,13 @@
...
+            <template v-if="$slots['content-layout-left-sidebar-wrapper']">
+                <vue-scroll :ops="scrollOps">
+                    <slot name="content-layout-left-sidebar-wrapper"></slot>
+                </vue-scroll>
+            </template>
+        </div>
+        <div v-if="!hideLeftSidebar && (collapsible && !isCollapsed)" class="cly-vue-content-builder__layout-main__body__left-sidebar" :style="{ 'width': this.leftSidebarWidth + 'px !important'}">
```
gabrieloliveirapinto, line 10: Same question as above

frontend/express/public/javascripts/countly/vue/templates/content/content-header.html
```diff
@@ -57,39 +58,44 @@
...
-            size="small"
-            @command="onCommand"
-        >
-            <el-dropdown-item
-                v-for="(option, idx) in options"
-                :key="`cly-vue-content-builder-header-option-${idx}`"
-                class="cly-vue-content-builder-header__option"
-                :command="option.value"
-                :disabled="option.disabled"
+        <template v-if="$slots.headerActions">
```
gabrieloliveirapinto, line 61: Here we can use the slot fallback, to have the same result with less HTML.

Meaning all that instead of this:

```
  <template v-if="$slots.headerActions">
    <slot name="headerActions"></slot>
  </template>
  <template v-else>
    <cly-status-badge
      ...
    />
    ...
  </template>
```

we can have this

```
  <slot name="headerActions">
    <cly-status-badge
      ...
    />
    ...
  </slot>
```

It work the same

If the slot is present, it will display whatever we add inside the slot on the parent component.
And if we do not add the slot on the parent component, it will display the elements that are inside of the slot in the `content-header` template.

pnrgenc: good suggestion, thank you

gabrieloliveirapinto (approved):

PR #5944, opened 2025-01-23
[QT-264] Add data test id to content htmls in countly-server

can-angun:
https://countly.atlassian.net/browse/QT-264


savascountly (approved):
Good job!

PR #5970, opened 2025-02-03
[SER-2121] The Internal Name" and "Add user consent" texts are not correct

pnrgenc:


ArtursKadikis (changes requested):
missing changelog entry

edit: nvm, is fine leaving this out

ArtursKadikis (approved):

can-angun (approved):

PR #6130, opened 2025-04-09
[QT-278] Added data test ids to content drawer elements

can-angun:
https://countly.atlassian.net/browse/QT-278


savascountly (approved):
Thanks for your effort!

PR #6134, opened 2025-04-11
[QT-278] Added test id to collapse item in content drawer

can-angun:


savascountly (approved):
Thanks for your effort!

PR #6142, opened 2025-04-15
[QT-278] Add data test id to select device buttons

can-angun:


savascountly (approved):
Thanks for your effort!

PR #6223, opened 2025-05-13
[SER-2423] Publish toogle converted to button

coskunaydinoglu:


frontend/express/public/javascripts/countly/vue/components/content.js
```diff
@@ -136,6 +138,15 @@
                 }
             },
 
+            localValue: {
```
gabrieloliveirapinto, line 141: Why reposition this variable?

Nothing was changed, and I had this alphabetically ordered

It's not a blocking change, I'm just curious

frontend/express/public/javascripts/countly/vue/components/content.js
```diff
@@ -119,11 +119,13 @@
             'tab-change'
         ],
 
-        data: () => ({
-            currentTab: null,
-
-            isReadonlyInput: true
-        }),
+        data() {
```
gabrieloliveirapinto, line 122: Was there a reason to change the `data` definition?

It's not a blocking change, I'm just curious

frontend/express/public/javascripts/countly/vue/components/content.js
```diff
@@ -196,6 +198,11 @@
                 this.$emit('save');
             },
 
+            onPublishButtonClick() {
+                this.toggleLocalValue = !this.toggleLocalValue;
+                //:this.$emit('switch-toggle', this.toggleLocalValue);
```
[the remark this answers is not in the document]
gabrieloliveirapinto: I agree with Copilot that we should remove commented code. But if there is a reason to have this commented, a note should inform other coders who will work on this.

This is just my opinion, but we should alphabetically order variables, computed, and methods. It helps to organize the code and to find it.

frontend/express/public/javascripts/countly/vue/components/content.js
coskunaydinoglu: This also unintended change placing it back

PR #6276, opened 2025-06-03
Usermerge Script Update

John-Weak:
- updated user merge script to include dry run.
- also added option to merge user on the basis of any field.

### DIFF
--- a/bin/scripts/fix-data/user-merge.js
+++ b/bin/scripts/fix-data/user-merge.js
@@ -1,12 +1,20 @@
 /**
- *  Description: This script is used to merge users based on username.
+ *  Description: This script merges users if they match on any fields configured in the script.
+ *  configure - processAllFields function to add or remove fields for merging.
  *  Server: countly
  *  Path: $(countly dir)/bin/scripts/fix-data
- *  Command: node user-merge.js
+ *  Command: node user-merge.js --no-dry-run
  */
 var pluginManager = require("../../../plugins/pluginManager.js");
 var appUsers = require("../../../api/parts/mgmt/app_users.js");
 var common = require("../../../api/utils/common.js");
+
+// Check for dry run flag
+let DRY_RUN = true;
+if (process.argv.includes('--no-dry-run')) {
+    DRY_RUN = false;
+}
+console.log(DRY_RUN ? "Running in DRY RUN mode - no actual merges will be performed" : "Running in LIVE mode - merges will be performed");
 
 console.log("Merging app users");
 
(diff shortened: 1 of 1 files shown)


bin/scripts/fix-data/user-merge.js
```diff
@@ -26,78 +33,109 @@ var RECORD_OVERLOAD_SLEEP = 2000;
...
-        for (var i = 0; i < duplicates.length; i++) {
+        for (const duplicate of duplicates) {
+            const query = { [field]: duplicate._id };
 
-            var mainUser = null;
-            var mergedUsersUIDs = [];
+            const users = await common.db.collection(COLLECTION_NAME)
+                .find(query)
+                .sort({ lac: -1 })
+                .toArray();
```
pnrgenc: I am not sure if this is processed with toArray() 
Is there a possibility of a large data set coming in here? Then I can suggest using slower but safer methods, like cursor or stream
When I checked before, it was processed with cursor :)

Another thing; I am not sure about using "lac"
"ls" last seen is referenced for merge operation before, I think there should be no problem but I am not sure about this. But if we are going to use "lac" let's make sure it's indexed

bin/scripts/fix-data/user-merge.js
```diff
@@ -1,21 +1,28 @@
...
+// Check for dry run flag
+let DRY_RUN = true;
+if (process.argv.includes('--no-dry-run')) {
+    DRY_RUN = false;
+}
+console.log(DRY_RUN ? "Running in DRY RUN mode - no actual merges will be performed" : "Running in LIVE mode - merges will be performed");
+
 console.log("Merging app users");
 
 var APP_ID = "";
```
pnrgenc, line 13: It might be good to write the variables we expect from the user at the top of the script, they might be overlooked below and it would be better if we check, like showing a message if app_id is not entered and do not continue

bin/scripts/fix-data/user-merge.js
```diff
@@ -26,78 +33,109 @@ var RECORD_OVERLOAD_SLEEP = 2000;
...
+            if (users.length < 2) {
+                continue;
+            }
 
-            while (await cursor.hasNext()) {
-                var doc = await cursor.next();
+            const mainUser = users[0]; // Most recent user becomes main user
+            const usersToMerge = users.slice(1); // All other users will be merged
+            //const noFlag = users.every(user => user['name'] === users[0]['name']); //true if all names are same
+            const mergedUIDs = [];
```
pnrgenc: There is a warning in DeepScan for this assignment. It would be great if you can change it with counter

> This array is useless because added elements are never used. Consider using a numeric variable instead of the array if only the 'length' property is needed.

pnrgenc (changes requested):

PR #6390, opened 2025-07-07
 Changes to run queries(not changestreams) for data processing 

Cookiezaurs:


api/utils/common.js
```diff
@@ -3446,17 +3446,29 @@ common.applyUniqueOnModel = function(model, uniqueData, prop) {
...
  * Shifts hourly data (To be in different timezone)
  * @param {*} data  array of data
- * @param {*} offset (integer)
+ * @param {*} offset (integer) - full hours
+ * @param {string} field - field to shift. Default is "_id"
  * @returns {Array} shifted data
  */
-common.shiftHourlyData = function(data, offset) {
+common.shiftHourlyData = function(data, offset, field = "_id") {
+    var dd, iid;
```
kanwarujjaval, line 3454: is it possible to have this method a bit more readable by having nicer variable names and comments describing date operations?

api/utils/requestProcessor.js
```diff
@@ -2656,7 +2656,52 @@ const processRequest = (params) => {
                     }
                     break;
                 case 'get_events':
-                    validateRead(params, 'core', countlyApi.data.fetch.fetchCollection, 'events');
+                    //validateRead(params, 'core', countlyApi.data.fetch.fetchCollection, 'events');
+                    validateRead(params, 'core', function() {
+
+                        common.db.collection("events").findOne({ '_id': params.app_id }, function(err, result) {
+                            if (!result) {
```
kanwarujjaval, line 2663: lets use promises for mongodb

api/parts/data/fetch.js
```diff
@@ -21,9 +21,25 @@ var fetch = {},
     crypto = require('crypto'),
     usage = require('./usage.js'),
     plugins = require('../../../plugins/pluginManager.js');
-
-
-
+const { fetchAggregatedSegmentedEventData } = require('../queries/events.js');
+
+
+fetch.fetchEventMetaData = async function(options) {
```
kanwarujjaval, line 32: needs jsdoc.
also in general named functions are easier to debug and traces are better

api/parts/data/dataBatchReader.js
```diff
@@ -0,0 +1,200 @@
+const common = require("../../utils/common");
+const log = require('../../utils/log.js')("dataBatchReader");
+var plugins = require("../../../plugins/pluginManager.js");
+var {fetchDataForAggregator} = require("../queries/aggregator.js");
+
+/**
+ * Class to read/aggregate data in batches from mongodb and pass for processing.
+ */
+class dataBatchReader {
```
kanwarujjaval: sorry for the nitpick: classes should be named PascalCase

api/parts/data/dataBatchReader.js
```diff
@@ -0,0 +1,200 @@
...
+        this.pipeline = options.pipeline || [];
+        this.match = options.match || {};
+        this.timefield = options.timefield || "cd"; //default time field is cd
+        this.lastToken = null;
+        this.name = options.name || "";
+        this.collection = options.collection || "drill_events";
+        this.options = options.options;
+        this.onData = onData;
+        this.interval = options.interval || 10000;
+        if (plugins.getConfig("aggregator") && plugins.getConfig("aggregator").interval) {
```
kanwarujjaval, line 27: Does this mean that the interval passed to the object will always be overriden by config?

api/parts/data/dataBatchReader.js
```diff
@@ -0,0 +1,200 @@
...
+                match.cd = {$gte: new Date(cd), $lt: cd2};
+            }
+            pipeline.unshift({"$match": match});
+
+            //Reads from mongodb or clickhouse(if code is defined);
+            var data = await fetchDataForAggregator({pipeline: pipeline, name: this.name, cd1: cd, cd2: cd2});
+
+            try {
+                if (data.length > 0) {
+                    await this.onData({"token": "timed", "cd": cd2, "_id": null}, data);
```
kanwarujjaval: we might need to mark the data that we are processing somehow, either logically or actual mark in db, and then complete that mark. 
This will not only allow us to keep processing consistent but will enable parallel aggregation since data is marked (in some way) before processing and mark is complete at the end

api/parts/data/dataBatchReader.js
```diff
@@ -0,0 +1,200 @@
...
+            }
+            catch (err) {
+                log.e(this.name + ": Error processing data for - " + cd2 + " " + JSON.stringify(err));
+                return; //exiting loop.
+            }
+            cd = cd2;
+            cd2 = new Date(cd2.valueOf() + this.interval);
+            now = Date.now().valueOf();
+        }
+        setTimeout(() => {
```
kanwarujjaval: lets clear this timeout in close() too

api/parts/data/dataBatchReader.js
```diff
@@ -0,0 +1,200 @@
...
+ */
+class dataBatchReader {
+    /** 
+    * @param {Object} db - Database object
+    * @param {Object} options - Options object
+    * @param {function} onData - Finction to call when getting new data from stream
+    */
+    constructor(db, options, onData,) {
+        log.e("Creating dataBatchReader", JSON.stringify(options));
+        this.db = db;
```
kanwarujjaval, line 17: We should use this.db everywhere for consistency 
and change the assignment to
this.db = db || common.db;

api/parts/data/dataBatchReader.js
```diff
@@ -0,0 +1,200 @@
...
+     * Processes range of dates
+     * @param {date} cd  - start time
+     */
+    async processNextDateRange() {
+        var lastToken = await common.db.collection("plugins").findOne({"_id": "_changeStreams"}, {projection: {[this.name]: 1}});
+        var cd = lastToken && lastToken[this.name] ? lastToken[this.name].cd : Date.now();
+        var cd2 = cd.valueOf() + this.interval;
+        var now = Date.now().valueOf();
+        while (cd2 < now) {
+            cd2 = new Date(cd2);
```
kanwarujjaval: cd2 is first defined as a number, then assigned to a date object.
Also cd2 is a bad naming, doesnt tell me why we needed second cd or why it is being assigned different type.

for query purposes, lets just convert it to date object for the query and keep the variable always a number

api/parts/data/dataBatchReader.js
```diff
@@ -0,0 +1,200 @@
...
+            var pipeline = JSON.parse(JSON.stringify(this.pipeline)) || [];
```
kanwarujjaval, line 76: parse&stringify will lose date,bson types, do we really need deep clone?
can we structuredClone?

api/parts/data/dataBatchReader.js
kanwarujjaval: maybe we can add a lease document, with the query and if its in progress, basically query state instead of marking all the documents we are processing. so the aggregator can know the lease that is going on for parallel processing, or lease that was never completed for consistency

kanwarujjaval (changes requested):
Left some comments

api/parts/data/dataBatchReader.js
Cookiezaurs: Yes. It is not meant to override config in real life.
Added mostly so I can do tests and data is updated in 100 ms.

PR #6398, opened 2025-07-08
[remote-config] Enable app version older/newer comparison

widatama:


ArtursKadikis (changes requested):
The changelog does not mention the feature addition of app version filters

ArtursKadikis (changes requested):
The changelog changes modifies the changelog for previous released versions, it should not do that

PR #6416, opened 2025-07-16
fix: Delete widgets when a dashboard is removed

pnrgenc:


bin/scripts/fix-data/delete_widgets_of_deleted_dashboards.js
```diff
@@ -0,0 +1,106 @@
...
+                    orphanLongTaskIds.push(...w.drill_report);
+                }
+            }
+        }
+        await widgetCursor.close();
+
+        console.log(`Orphan widgets found: ${orphanWidgetIds.length}`);
+        console.log(`Linked long_tasks to drop: ${orphanLongTaskIds.length}`);
+
+        await deleteByChunks(db, 'widgets', orphanWidgetIds);
```
widatama, line 94: The script will always delete widgets and long tasks here.
Is there no checking for dry run?

widatama (approved):

plugins/dashboards/api/api.js
```diff
@@ -1093,30 +1093,53 @@ plugins.setConfigs("dashboards", {
...
 
-                        common.db.collection("dashboards").remove(
-                            filterCond,
-                            function(error, result) {
-                                if (!error && result) {
-                                    if (result && result.result && result.result.n === 1) {
-                                        plugins.dispatch("/systemlogs", {params: params, action: "dashboard_deleted", data: dashboard});
+                        try {
+                            // Remove the dashboard
+                            const result = await common.db.collection("dashboards").deleteOne(filterCond);
```
John-Weak: Hello,
If we delete the `dashboards` too early, we lose the opportunity to implement a retry mechanism. However, if we perform the deletion of `dashboards` at a later stage and some cleanup fails, we still have the option to retry.

pnrgenc: Thank you for your comment Tushar, you are right. 
It is fixed :)

John-Weak (approved):

PR #6445, opened 2025-07-23
Update app version log and generator

widatama:


ar2rsawseen (approved):

plugins/populator/frontend/public/javascripts/countly.models.js
```diff
@@ -337,9 +337,13 @@
                 d = moment(seed);
             }
         }
-        var year = trimYear ? d.format('YY')[1] : d.format('YY');
+
+        var year = d.format('YY');
```
kanwarujjaval, line 341: lets use moment for this code

widatama: I'm not sure I get this.
You mean to create new moment object from `d`?

ArtursKadikis: resolving due to what Ujjaval said

PR #6508, opened 2025-08-01
fix: Remove revenue plugin and its all dependencies

pnrgenc:


ar2rsawseen:
For now let's close it, we decided it is easier to keep this in new arch then deal with the problems of removing it
but we could reuse it later when we would actually remove it

PR #6732, opened 2025-10-02
[sdk] Experimental section

turtledreams:
image
image


bin/scripts/localization/clean-all-plugins.sh
```diff
@@ -224,7 +224,7 @@ echo "Failed: $FAILED"
 echo "Plugins with unused keys: $WITH_UNUSED_KEYS"
 echo
 
-if [ $FAILED -gt 0 ]; then
+if [ "$FAILED" -gt 0 ]; then
```
turtledreams, line 227: this was failing the lint check for me

PR #6737, opened 2025-10-03
[SER-2520] Persist newly created app after page refresh

gabrieloliveirapinto:


widatama (approved):
Don't forget to add the changelog

PR #6776, opened 2025-10-09
[Fix]: sort events alphabetically

tryshank:
## What
Sorts events alphabetically in the All events list UI.

## Why
[Jira](https://countly.atlassian.net/browse/SER-1518) Makes events easier to find.

## Changes
- Added `.sort()` to `getAllEventsList()` in events UI

## Files
- `core/frontend/express/public/core/events/javascripts/countly.details.models.js`

Total: 6 files, ~10 lines added


countly-community
ar2rsawseen, line 1: Whats this?

tryshank: Not sure, I think it was something that came up during my first build attempts. Fixed now

PR #6797, opened 2025-10-15
Implement notification goto, update changelog

widatama:


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## Version 24.05.X
+Fixes:
+- [core] Implement go to link in notification
```
gabrieloliveirapinto, line 3: Since it is an implement, shouldn't this entry be a feature?

frontend/express/public/javascripts/countly/vue/core.js
```diff
@@ -688,7 +688,7 @@
 
     var NotificationToastsView = {
         template: '<div class="notification-toasts"> \
-                        <cly-notification v-for="(toast) in notificationToasts" :key="toast.id" :id="toast.id" :text="toast.text" :autoHide="toast.autoHide" :color="toast.color" :closable="true" :customWidth="toast.width" @close="onClose" class="notification-toasts__item"></cly-notification>\
+                        <cly-notification v-for="(toast) in notificationToasts" :key="toast.id" :id="toast.id" :text="toast.text" :goTo="toast.goTo" :autoHide="toast.autoHide" :color="toast.color" :closable="true" :customWidth="toast.width" :toast="true" @close="onClose" class="notification-toasts__item"></cly-notification>\
```
gabrieloliveirapinto: It seems to me that the toast variable has the same properties as the component props, and also, we have some props that are always true, so to simplify the HTML, we could have done the following:

```suggestion
    <cly-notification
        v-for="toast in notificationToasts"
        :key="toast.id"
        v-bind="toast"
        class="notification-toasts__item"
        closable
        :customWidth="toast.width"
        toast
        @close="onClose"
    />
```

Just a suggestion, I will leave it up to you to decide whether to add or not.

PR #7011, opened 2025-11-28
fix: sdk ui updates wrap it all

pnrgenc:


pnrgenc:
will be merging "next->newarch"

PR #7033, opened 2025-12-05
fix: compliance-hub issues in `newarchitecture`

pnrgenc:
**Fixes:**

- Missing segmentation caused consent metrics to be non-drillable.

- Metrics were not stored in the `consents` collection, so they didn’t appear in the metrics chart.

- Incorrect filtering logic in the Consent History tab.

- Incorrect opt-in / opt-out results in the Consent History tab.


api/ingestor/requestProcessor.js
```diff
@@ -1004,6 +1005,9 @@ const processRequest = (params) => {
     params.fullPath = paths.join("/");
 
     reloadConfig().then(function() {
+        if (!common.writeBatcher && common.db) {
+            common.writeBatcher = new WriteBatcher(common.db);
```
pnrgenc: added to ensure writeBatcher exists in ingestor process so recordMetric can enqueue writes

plugins/compliance-hub/api/aggregator.js
```diff
@@ -0,0 +1,140 @@
...
+                                "_uid": "$_uid"
+                            }
+                        }
+                    ]
+                }
+            }
+        });
+
+        try {
+            for await (const {events} of eventSource) {
```
pnrgenc: to-do: I will change this with `eventSource.processWithAutoAck`

Cookiezaurs (approved):

PR #7161, opened 2026-01-14
Fix/memory leak

davidecavaliere:
This fixes the "alleged" memory leak which was indeed a memory leak 🤯

### DIFF
--- a/frontend/express/public/core/events/javascripts/countly.overview.models.js
+++ b/frontend/express/public/core/events/javascripts/countly.overview.models.js
@@ -305,12 +305,12 @@
                     "method": "top_events",
                     "period": "30days",
                     "filter": filter,
-                    "limit": limit
+                    "limit": limit,
+                    "preventRequestAbort": true
                 },
                 dataType: "json",
             }, {
                 "disableAutoCatch": true,
-                preventRequestAbort: true,
             });//to be able to see if discarded
         },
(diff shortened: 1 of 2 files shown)


ar2rsawseen (approved):

frontend/express/public/core/events/javascripts/countly.overview.models.js
```diff
@@ -308,7 +308,10 @@
                     "limit": limit
                 },
                 dataType: "json",
-            }, {"disableAutoCatch": true});//to be able to see if discarded
+            }, {
+                "disableAutoCatch": true,
+                preventRequestAbort: true,
```
pnrgenc: sorry I didn't give details before. `preventRequestAbort` is a Countly flag: https://github.com/Countly/countly-server/blob/master/frontend/express/public/javascripts/countly/countly.template.js#L4047
It’s read from data. Can you please add this in data? :)

pnrgenc (changes requested):

frontend/express/public/core/events/javascripts/countly.overview.models.js
davidecavaliere: Sorry I could have seen it looking at the other methods :facepalm:

pnrgenc (approved):

PR #7293, opened 2026-02-11
feat(logging): human-readable console logs and cleaner plugin errors

davidecavaliere:
## Summary

This PR improves the developer experience when working with Countly by making console logs more human-readable during development.

### Changes

**Human-readable console logs**
- Enable `pino-pretty` by default for console output
- Simplify timestamps to show only time (`HH:MM:ss.l`) instead of full datetime
- JSON output can still be enabled via `config.logging.prettyPrint = false` for production/log aggregation

**Cleaner plugin load errors**
- Plugin load failures now show a single concise line instead of dumping the full stack trace
- Full error details available via `DEBUG=countly:plugins` environment variable

### Why these changes?

**Console readability**: The default JSON output from Pino is optimized for machine parsing and log aggregation, but it's difficult for developers to scan during local development. Pretty-printed logs with simplified timestamps make it much easier to follow what's happening when debugging.

**Debug-js for verbose errors**: When a plugin fails to load due to a missing dependency, the full stack trace with "Require stack" clutters the console and obscures other important startup messages. Using `debug-js` follows the Node.js ecosystem convention where verbose output is opt-in via the `DEBUG` environment variable, keeping the default output clean while preserving access to detailed diagnostics when needed.

## Test plan
- [ ] Start API server and verify logs are pretty-printed with time-only timestamps
- [ ] Verify plugin load errors show single-line message
- [ ] Verify `DEBUG=countly:plugins` shows full stack traces
- [ ] Verify `config.logging.prettyPrint = false` disables pretty-printing

### Example of console output from starting the ingestor process
image

### Example of console output when starting api without `DEBUG` set
image
### Example of console output when starting api with `DEBUG=countly:*` 
image

### DIFF
--- a/api/utils/log.js
+++ b/api/utils/log.js
@@ -216,9 +216,12 @@ class LogManager {
         this.#prefs = loadLoggingConfig();
         this.#prefs.default = this.#prefs.default || 'warn';
         this.#deflt = this.#prefs.default || 'error';
-        // Pretty-print enabled by default for human-readable console output
-        // Set config.logging.prettyPrint = false for JSON output (e.g., for log aggregation)
-        this.#prettyPrint = this.#prefs.prettyPrint !== false;
+        // Output format determined by: explicit config > environment detection
+        // - prettyPrint: true  → always pretty (human-readable)
+        // - prettyPrint: false → always JSON (for log aggregation)
+        // - prettyPrint: unset → auto-detect (pretty in terminal/CI, JSON in Docker/PM2/pipes)
+        const isInteractive = process.stdout.isTTY || !!process.env.CI;
+        this.#prettyPrint = this.#prefs.prettyPrint ?? isInteractive;
 
         // Initialize OpenTelemetry metrics if available
         if (metrics) {
--- a/api/utils/log.js
+++ b/api/utils/log.js
@@ -1,5 +1,14 @@
 const pino = require('pino');
 
+// Optional pino-pretty import for human-readable console output
+let pinoPretty;
+try {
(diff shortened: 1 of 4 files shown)


api/utils/log.js
```diff
@@ -207,7 +216,9 @@ class LogManager {
         this.#prefs = loadLoggingConfig();
         this.#prefs.default = this.#prefs.default || 'warn';
         this.#deflt = this.#prefs.default || 'error';
-        this.#prettyPrint = this.#prefs.prettyPrint || false;
+        // Pretty-print enabled by default for human-readable console output
+        // Set config.logging.prettyPrint = false for JSON output (e.g., for log aggregation)
+        this.#prettyPrint = this.#prefs.prettyPrint !== false;
 
         // Initialize OpenTelemetry metrics if available
```
kanwarujjaval, line 226: I like this but please dont enable this by default.
This is only good for dev envs, since most of our logs are now json ( the ones that are not will be converted ) They will be human readable for prod using grafana or other tools.

The other issue with pretty is it works with event emitter and since countly has huge number of subloggers (plugins,modules each have their own) This can cause memory leaks and similar issues

ref: https://github.com/pinojs/pino-pretty/issues/387
ref: https://github.com/pinojs/pino/issues/144

Lets keep this but not enable by default, or by default when NODE_ENV === development or similar?

kanwarujjaval (changes requested):

davidecavaliere:
closing as this is not relevant for now

PR #7618, opened 2026-06-01
security: fix stored XSS in remote-config condition definition

ar2rsawseen:
## Summary

Fixes a **stored XSS** in the `remote-config` plugin, reported via the security program (validated against master).

`condition_definition` is stored verbatim from `POST /i/remote-config/add-condition` / `update-condition` — the handlers validate `condition_name` (alphanumeric) and `condition_color` but never sanitize `condition_definition`. The conditions table then rendered it with **`v-html`**:

```html
<!-- plugins/remote-config/frontend/public/templates/conditions.html:29 -->
<div ... v-html="rowScope.row.condition_definition"></div>
```

`v-html` assigns to `innerHTML`, so a stored payload like `` executes when another dashboard user opens the Remote Config Conditions view for that app.

## Impact

- **Attacker:** any user with the assignable `remote_config` create/update permission on an app.
- **Victim:** any user who opens that app's conditions view — including an app admin or **global admin**, whose session the injected script then runs in (same-origin dashboard/API access). Low-privilege → admin privilege escalation.

## Fix

Render `condition_definition` with `v-text` instead of `v-html`. It's a query-builder **display string** (e.g. `"Device = Asus Nexus 10"`), never intended to be HTML — so text rendering loses no functionality and additionally fixes display of query text containing `<`/`>` comparison operators. It was the only `v-html` sink in the plugin.

## Severity

High (per [SECURITY.md](https://github.com/Countly/countly-server/blob/master/SECURITY.md)) — stored XSS requiring victim action (opening the view), with privilege-escalation potential into an admin session.

## Testing note

The fix is in DOM rendering (Vue template), so the meaningful regression test is a Cypress UI spec under `ui-tests/cypress/.../remoteConfig` asserting an injected payload renders as inert text. Recommended as a follow-up (separate UI-test harness). The fix itself follows directly from Vue semantics: `v-text` sets `textContent` and never parses HTML.

Reported by: external researcher.

### DIFF
--- a/plugins/remote-config/frontend/public/templates/conditions.html
+++ b/plugins/remote-config/frontend/public/templates/conditions.html
@@ -26,7 +26,7 @@
         </el-table-column>
         <el-table-column test-id="condition" prop="condition_definition" :label="i18n('remote-config.definition')">
           <template v-slot:default="rowScope">
-            <div :data-test-id="'datatable-condition-description-' + rowScope.$index" v-html="rowScope.row.condition_definition"></div>
+            <div :data-test-id="'datatable-condition-description-' + rowScope.$index" v-text="rowScope.row.condition_definition"></div>
           </template>
         </el-table-column>
         <el-table-column test-id="condition" prop="seed_value" :label="i18n('remote-config.seed-value')">


ar2rsawseen:
Closing — this change is incorrect. Countly escapes all API output globally via common.returnOutput / returnMessage (escape_html_entities with more=true), and the dashboard intentionally relies on that: v-html renders the pre-escaped entities back to the correct visible characters while keeping any injected markup inert (a payload arrives as `&lt;img onerror&gt;`, an entity, not a tag, so it never executes).

Switching condition_definition to v-text would display the raw escaped entities (e.g. query text `age < 18` would show as `age &lt; 18`) — a display regression with no security benefit, since the value is already inert via output escaping.

The reporter's PoC fed a raw payload directly into a Vue v-html sink, bypassing the API output escaping that the real data path applies, so it does not reproduce end-to-end. No code change needed.

PR #7620, opened 2026-06-01
security: scope data migration export listing to owner/authorized apps

ar2rsawseen:
## Summary

Fixes an information-disclosure / IDOR in the Data Migration **export listing** (reported via the security program, validated on master).

`/o/datamigration/getmyexports` ran:

```js
common.db.collection("data_migrations").find().sort({ts: -1})  // no filter
```

after only a feature-level `data_migration:read` check on the request `app_id`. Export records are global, so **any** user with that permission on a single app could enumerate **every** export on the instance, exposing each record's `apps`, `userid`, `email`, `server_address`, **`server_token`**, and archive availability — cross-tenant metadata + the remote import token.

## Fix

Scope the listing for non-global-admins to exports they own (`userid`) or that target apps they can read (`apps` ∩ their `data_migration`-readable apps). Global admins still see everything. Since the list is now restricted to the caller's own/authorized exports, the `server_token`/`server_address` the UI's re-send action needs are still present for legitimate rows.

Adds a regression test: seeds a victim export (other user, other app) and asserts a scoped user does not receive it.

## Severity

Medium — authenticated; discloses cross-tenant migration metadata including remote `server_token`.

## Scope note

The same report also flagged `/i/datamigration/sendexport` re-sending a victim export — that vector is fixed in **#7611** (per-app `hasCreateRight` check against the export's stored `apps`, plus `exportid` sanitization). This PR covers the distinct listing/enumeration issue.

### DIFF
--- a/plugins/data_migration/api/api.js
+++ b/plugins/data_migration/api/api.js
@@ -37,6 +37,34 @@ function safeDataMigrationId(exportid) {
 }
 
 /**
+ * Whether the member may access an export record. Export records live in the
+ * global data_migrations collection keyed by exportid = SHA1(JSON.stringify(apps)),
+ * which is computable from public app ids - so endpoints that operate on a
+ * record by exportid alone must additionally verify ownership/app access.
+ * @param {object} member - request member
+ * @param {object} exportRecord - data_migrations document
+ * @returns {boolean} true if the member owns it, is a global admin, or has
+ *                    data_migration access to one of the export's apps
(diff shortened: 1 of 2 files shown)


ar2rsawseen:
Closing: the data_migration plugin is being removed from the product entirely, so these per-endpoint hardening fixes are superseded by the plugin removal (separate PR). No longer needed.

PR #7630, opened 2026-06-02
security: restrict member/icon upload to own image or global admin

ar2rsawseen:
## Summary

Found by the Express-route cross-app census (the "(b)" pass over `frontend/express/app.js` + lifecycle hooks).

`POST /member/icon` updates `members.{_id: member_image_id}` with the uploaded avatar, taking `member_image_id` straight from the request body **without checking it is the caller's own id**. The auth gate `validateCreate(params, 'global_upload')` is satisfied by an **app-admin** (via an `app_id` of their own app), so such a user could overwrite **any member's avatar** by supplying that member's `_id` — a cross-user write (missing authorization). The image is rasterized to PNG (no XSS), so impact is defacement, but it's still a real authz gap.

## Fix

Allow the upload only when `member_image_id` is the caller's own id, or the caller is a global admin. The legitimate "upload another member's image" path is the user-management UI, and `/i/users/create|update` are **global-admin only** (`validateUserForGlobalAdmin`), so the global-admin allowance preserves that flow while blocking app-admins from targeting arbitrary members.

## Severity

Low — cross-user avatar overwrite by an app-admin; rasterized image (no script), only the `member_image` field is affected.

## Census result

The rest of the Express routes and the internal `/i/apps/*` / `/i/app_users/*` lifecycle hooks were verified **safe**: file-serving routes use `resolvePathInBase` (no traversal) and serve non-sensitive assets; settings routes scope to `req.session.uid`; `users/check/*` are global-admin only; `/login/token/:token` gates on token purpose; icon uploads are MIME-allowlisted + rasterized; lifecycle hooks receive a pre-authorized `appId` from the apps controller. `member/icon` was the only gap.

### DIFF
--- a/api/parts/data/fetch.js
+++ b/api/parts/data/fetch.js
@@ -521,6 +521,7 @@ fetch.fetchAllApps = function(params) {
     if (params.qstring.filter) {
         try {
             filter = JSON.parse(params.qstring.filter);
+            common.stripUnsafeMongoOperators(filter);
         }
         catch (ex) {
             filter = {};
--- a/api/parts/mgmt/cms.js
+++ b/api/parts/mgmt/cms.js
@@ -200,6 +200,7 @@ cmsApi.getEntriesWithUpdate = function(params) {
 
     try {
         params.qstring.query = JSON.parse(params.qstring.query);
+        common.stripUnsafeMongoOperators(params.qstring.query);
     }
     catch (ex) {
         params.qstring.query = null;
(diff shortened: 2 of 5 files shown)


ar2rsawseen:
Superseded by a clean branch (#7630 accidentally included the unrelated NoSQL-sanitizer commit from #7629 because it was branched off that branch). Reopening the same change on a fresh branch off master.

PR #7632, opened 2026-06-02
security: cast app_key to string before apps lookup (NoSQL operator injection on /o/sdk)

ar2rsawseen:
## Summary

The SDK fetch path (`validateAppForFetchAPI` in `api/utils/requestProcessor.js`) and two plugin app lookups (`logger` `/sdk/cancel`, `views` `/o/actions`) queried the `apps` collection with the raw, user-supplied `app_key`. A non-string value such as `{"$regex":"^a"}` was passed straight through to MongoDB as a query operator.

On `/o/sdk` this produced an unauthenticated **boolean oracle**: HTTP 200 when the regex matched an existing `app_key`, HTTP 400 (`App does not exist`) when it did not. An attacker could extract any app's `app_key` character-by-character, then use it for unauthenticated write ingestion (event/session injection, analytics pollution) on `/i`.

`common.stripUnsafeMongoOperators` only strips `$where/$expr/$function/$accumulator`, so it did not catch `$regex` here.

## Fix

Cast `app_key` to a string at all three lookup sites (`params.qstring.app_key + ""`), matching the already-safe write path `validateAppForWriteAPI`. Operator objects can no longer be honored.

- `api/utils/requestProcessor.js:3630` (`validateAppForFetchAPI`)
- `plugins/logger/api/api.js:316` (`/sdk/cancel`)
- `plugins/views/api/api.js:1569` (`/o/actions`)

## Test

Adds `test/2.api/12.fail.sdk.appkey.injection.js`: sends `app_key[$regex]=^<real key prefix>` to `/o/sdk` fetch and asserts `400 App does not exist` rather than a prefix match.

### DIFF
--- a/api/utils/requestProcessor.js
+++ b/api/utils/requestProcessor.js
@@ -3627,7 +3627,7 @@ const validateAppForFetchAPI = (params, done, try_times) => {
     if (ignorePossibleDevices(params)) {
         return done ? done() : false;
     }
-    common.readBatcher.getOne("apps", {'key': params.qstring.app_key}, (err, app) => {
+    common.readBatcher.getOne("apps", {'key': params.qstring.app_key + ""}, (err, app) => {
         if (!app) {
             common.returnMessage(params, 400, 'App does not exist');
             params.cancelRequest = "App not found or no Database connection";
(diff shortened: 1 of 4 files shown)


Cookiezaurs (approved):

ar2rsawseen:
Superseded by #7633, which centralizes this as request-intake parameter normalization.

PR #420, opened 2018-02-07
Refactor request processing

FoysalOsmany:

PR #7660, opened 2026-06-04
fix(subdirectory): annotation icons, Pace guard, star-rating assets

ar2rsawseen:
## Summary

Port of [Countly/countly-platform#420](https://github.com/Countly/countly-platform/pull/420) — two frontend fixes for subdirectory (`/<subdir>/`) deployments, found while testing the subdirectory-path work on a real subdirectory deploy.

### 1. Annotation/notes icons 404 on subdirectory deploys (`vis.js`)
The chart "notes" dropdown rendered 4 icons via hardcoded **relative** paths (``). Under a subdirectory deploy the browser resolves `../images/...` from the document base `/<subdir>/dashboard` up to `/images/...` → **404** (the asset lives at `/<subdir>/images/...`). On a root deploy `../images/...` happens to resolve to the correct `/images/...`, so it was never caught.

The component already has a `getIconUrl(icon)` helper returning `` `${countlyGlobal.path}/images/annotation/${icon}.svg` `` but the template was never wired to it. This binds the 4 icons to that helper.

### 2. Pace `shouldIgnoreURL` null-guard (`pace.js`)
On data-heavy views, Pace's `shouldIgnoreURL(url)` could be called with an `undefined` url and throw `Cannot read properties of undefined (reading 'indexOf')` repeatedly (non-fatal, but noisy). The shipped `pace.min.js` already carries this guard; the CoffeeScript-compiled `pace.js` source did not — this adds the `url &&` guard to the source to close the source/min drift.

## Notes
- The platform PR's note about *not* removing `pace.min.js` from the `countly.dom.js` concat applies here too — that change was correctly dropped (Pace is loaded only via that bundle in production), and is not part of this PR.
- `node --check` passes on both files; eslint clean. Minified `/min/` bundles are rebuilt by CI.

### DIFF
--- a/frontend/express/public/javascripts/countly/vue/components/vis.js
+++ b/frontend/express/public/javascripts/countly/vue/components/vis.js
@@ -2002,12 +2002,12 @@
             '<div class="chart-type-annotation-wrapper">\
                 <el-dropdown data-test-id="chart-type-annotation-button" trigger="click" @command="graphNotesHandleCommand($event)">\
                     <el-button size="small">\
-                        <img src="../images/annotation/notation-icon.svg" class="chart-type-annotation-wrapper__icon" data-test-id="chart-type-annotation-icon"/>\
+                        <img :src="getIconUrl(\'notation-icon\')" class="chart-type-annotation-wrapper__icon" data-test-id="chart-type-annotation-icon"/>\
                     </el-button>\
                     <el-dropdown-menu slot="dropdown">\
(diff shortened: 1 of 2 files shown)


Cookiezaurs (approved):

ar2rsawseen:
Added the two further changes from countly-platform#420 (the PR grew after the initial port):
- `star-rating/countly.views.js`: prefix the ratings empty-state image with `countlyGlobal.path` (was root-absolute, 404s under subpath).
- `star-rating/ratings.scss`: rating-preview emoji/thumb/star `background-image` URLs made relative (`../images/...`) instead of root-absolute.

This PR now mirrors all 4 files of #420. Minified/compiled assets (incl. ratings.css) are rebuilt by CI.

PR #7690, opened 2026-06-09
dbviewer: harden aggregation and query handling

ar2rsawseen:
Consolidated DB Viewer hardening (supersedes #7685 and #7686, combined into one reviewable PR). Adds two small, unit-tested guard modules and applies them on **both** the global-admin and non-admin code paths.

### Aggregation
- Validate stages at every depth, including `$facet` sub-pipelines (recursive allow-list); require an explicit allow-list match so inherited `Object.prototype` keys (`constructor`/`__proto__`) are never treated as allowed.
- Block joins/unions into the redacted collections (`members`/`auth_tokens`), block write stages (`$out`/`$merge`), and block server-side-JS operators (`$function`/`$accumulator`/`$where`) — all detected by a **full deep walk**, so any (incl. future) nested stage shape is covered.
- Place the `members`/`auth_tokens` redaction as the **first** pipeline stage so no user stage can read the raw fields first.

### Find / document
- Restrict projections to strict include/exclude (`0`/`1`/booleans), dropping expression / field-path alias values; normalize a non-object projection to `{}`.
- Treat the `_id` search term (`sSearch`) as a literal (regex-escaped, no ReDoS).
- Scope single-document lookups to the caller's apps (same base filter as the listing path).
- Redact `members.two_factor_auth` alongside `password`/`api_key`.
- Cap result size (`limit` / `iDisplayLength`) and return generic 500 messages instead of raw MongoDB errors.

### Tests
`test/unit-tests/plugins.dbviewer.aggregation-guard.js` and `plugins.dbviewer.query-guard.js` — 33 cases covering recursive sanitization, prototype-key handling, deep-walk detection of protected-collection joins / write stages / server-side-JS, projection sanitization, regex escaping. All passing.

### DIFF
--- a/plugins/dbviewer/api/api.js
+++ b/plugins/dbviewer/api/api.js
@@ -12,55 +12,14 @@ const { MongoInvalidArgumentError } = require('mongodb');
 const { EJSON } = require('bson');
 
 const FEATURE_NAME = 'dbviewer';
-const whiteListedAggregationStages = {
-    "$addFields": true,
-    "$bucket": true,
-    "$bucketAuto": true,
-    //"$changeStream": false,
-    //"$changeStreamSplitLargeEvents": false,
-    //"$collStats": false,
-    "$count": true,
-    //"$currentOp": false,
-    "$densify": true,
-    //"$documents": false
-    "$facet": true,
-    "$fill": true,
-    "$geoNear": true,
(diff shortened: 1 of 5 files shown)


ar2rsawseen:
Superseded — keeping #7686 instead, which retains the full commit history of both PRs. Closing this squashed duplicate.

PR #7697, opened 2026-06-09
query validation: explicit confirm_delete_all + clear too-deep error

ar2rsawseen:
Follow-up to #7694 / #7695 addressing review comments raised on the 24.05 backport (#7696) that also apply to master.

- **`/i/app_users/delete` near-total guard**: require an explicit `confirm_delete_all` of `true` (or `"true"`). Previously any truthy value (e.g. the string `"false"`) bypassed the guard.
- **Too-deep sentinel leak**: add `common.unsafeQueryError(bad)` which maps the internal `UNSAFE_QUERY_TOO_DEEP` sentinel to a clear *"Query is nested too deeply"* message. `parseUserQuery` and all direct `findUnsafeMongoOperator` callers (dbviewer, exports, remote-config, push ×3) now use it, so `$__nestedTooDeep` is no longer surfaced to logs/clients.

No behavior change for valid queries. Cherry-pick of the same fix applied to the 24.05 backport (#7696).

### DIFF
--- a/api/parts/data/exports.js
+++ b/api/parts/data/exports.js
@@ -513,9 +513,9 @@ exports.fromDatabase = function(options) {
     // query here at the boundary into the dispatch + DB find.
     var badOp = common.findUnsafeMongoOperator(options.query);
     if (badOp) {
-        log.d("Rejected user query" + common.reqInfo(options.params) + ": " + "Query contains disallowed operator: " + badOp);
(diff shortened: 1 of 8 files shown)


ar2rsawseen:
Superseded — folding these review fixes into the already-open #7695 (same branch `fix/reject-unsafe-query-operators` → master) instead of a separate PR.

PR #7811, opened 2026-07-14
Ser 2915 dymanic parameters deeplinks

coskunaydinoglu:


savascountly (approved):
Thanks for your effort!

PR #7816, opened 2026-07-15
fix(security): add missing security headers to default config

ar2rsawseen:
## What

Adds the security headers that were missing from the default `security` config in `api/api.js`:

| Header | dashboard (before → after) | api (before → after) |
|---|---|---|
| `Referrer-Policy: strict-origin-when-cross-origin` | ❌ → ✅ | ❌ → ✅ |
| `Permissions-Policy: camera=(), microphone=(), geolocation=(), payment=()` | ❌ → ✅ | ❌ → ✅ |
| `X-Content-Type-Options: nosniff` | ✅ (unchanged) | ❌ → ✅ |

Existing values (`X-Frame-Options`, `X-XSS-Protection`, `Strict-Transport-Security`, `Access-Control-Allow-Origin`) are left byte-for-byte unchanged.

## Why

`Referrer-Policy` and `Permissions-Policy` had no code default at all, and `X-Content-Type-Options` was set for the dashboard but not for API routes (`/i`, `/o`). This ships a consistent hardened header set on a fresh install across both surfaces, rather than relying on a front-end reverse proxy to add them.

## Scope / compatibility

- **Defaults only.** Applied via the `security` config defaults, they take effect only when an instance has **no stored `security` config**. Existing installs keep their saved values and are unaffected.
- Faithful port of the equivalent countly-platform change; header set and values kept identical between the two repos.

### Note (out of scope)

This repo's **api** default also lacks `Strict-Transport-Security` (the dashboard default has it). I deliberately did **not** add STS to the api default here — it was outside the ported change and the platform repo already had it on the api side. Happy to add it in a follow-up if desired.


ar2rsawseen:
Good catch — addressed in 5e0a458. The `security` defaults are declared in two places (`api/api.js` and `frontend/express/app.js`); I've updated the `frontend/express/app.js` copy so both sites are now byte-for-byte identical for `dashboard_additional_headers` and `api_additional_headers`, removing the startup-order dependency. (Note: for the equivalent countly-platform PR this duplication doesn't exist — there `init_configs.ts` is the single `setConfigs('security')` source.)

Cookiezaurs (approved):

PR #7851, opened 2026-07-27
fix(api): refuse file uploads unless a plugin declared the path

ar2rsawseen:
## Problem

`api/api.js` parses every POST body with formidable **before** the request is routed or authorized, and formidable writes multipart parts and raw `application/octet-stream` bodies to disk. So any POST carrying such a body left a temp file behind — including requests to paths no handler serves, and requests rejected during validation. Nothing removed them, so they accumulated in the OS temp directory.

Cleaning up at each rejection site does not work: `rights.js` alone rejects in **63 places** across seven validators with no shared exit point, so an unauthenticated `POST /i/apps/update` would still leak.

## Change

**Uploads are refused by default.** A plugin declares the paths where it actually reads `params.files`:

```js
plugins.uploadPaths.push({path: "/i/license/upload"});
plugins.uploadPaths.push({path: "/i/crash_symbols/add_symbol", raw: true});
```

Mirrors the existing `plugins.ttlCollections` registry. `init()` requires every plugin's `api.js` well before the server listens, so declarations are always in place. Core declares `/i/apps/{create,update}`; star-rating declares `/i/feedback/{upload,logo}`.

- **Matching is exact**, after the installation subpath is stripped — so declaring `/i` (the SDK write endpoint, which carries the app-user picture) does not reopen the whole `/i` tree. `POST /i/anything` writes nothing.
- **`raw`** declares that an endpoint reads an `application/octet-stream` body. Without it the `octetstream` parser is left out, which is the only way to stop a raw body reaching disk — that parser calls `_newFile()` directly and never consults `filter`.
- Non-file parsers stay enabled throughout, so urlencoded and JSON bodies still populate `params.qstring` unchanged.

**Whatever does get written is removed when the response is done with**, on both `finish` and `close` — an upload aborted mid-stream leaves a partial file that `finish` never sees. Handlers that consume an upload still unlink it themselves, so in the normal case this is a no-op; what it catches is every request rejected before reaching a handler, with no rejection-site enumeration.

Cleanup works off a **snapshot of the paths formidable produced**, taken at parse time before any handler runs — not off `params.files`. That matters because `crash_symbolication` repoints `params.files.symbols.path` at files shipped with the plugin when serving populator data; those must never be deleted, and a snapshot makes that impossible by construction rather than by directory comparison.

No new configuration: `config.sample.js` and `configextender.js` are untouched by this branch.

## Ordering prerequisite

Cleanup at response time is only safe because handlers no longer use their temp file afterwards. `iconUpload` in `api/parts/mgmt/apps.js` was **not awaited** before `common.returnOutput`, and `jimp` opened the file after the response — it now waits, with a rejection handler since `iconUpload` can reject on a bad MIME type. The two equivalents in the enterprise plugins (`addAppUserPicture`, and the cohorts long task) are handled in the companion PR.

## Compatibility

⚠️ **Merges together with Countly/countly-enterprise-plugins#3324**, which declares the enterprise upload paths. Landing this alone would make those endpoints silently receive no file.

Note `init()` requires every plugin directory present, including disabled ones, so a disabled plugin's path is still declared — harmless, since the endpoint will not route and the file is discarded at response time.

No dependency change — `formidable` stays at 2.1.3.

## Verification

`test/unit-tests/api.utils.upload-temp.js` — 21 unit tests: every declared endpoint allowed; undeclared paths under `/i` refused (`/i/apps/delete`, `/i/bulk`, `/i/whatever`); exact matching (`/i` allowed, `/i/anything` refused); empty and absent registries refuse everything; `raw` honoured only where declared; subdirectory installs; near-misses like `/iffy`; and cleanup, including an explicit test that a path a handler substituted after parsing is left alone while the real upload is removed.

Verified end-to-end against a real `formidable@2.1.3` server mirroring the patched handler — 14 checks asserting both halves, that nothing is written for undeclared paths and that nothing survives the response even on declared paths whose handler never consumed the file: multipart and octet-stream to unrouted paths, `/i/apps/delete`, `/i/garbage` and `/o/export` write nothing; octet-stream to `/i` and `/i/apps/update` write nothing; `/i`, `/i/apps/update` and `/i/crash_symbols/*` receive their file and it is gone afterwards; urlencoded fields still parse on a blocked path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Version 25.03.xx
+Fixes:
+- [events] Fix sum chart tooltip displaying the raw floating-point value instead of a rounded number
+
 ## Version 25.03.50
 Fixes:
 - [star-rating] Fix custom widget logo resolving to the wrong path (mis-detected as the global app logo) after editing a widget
--- a/api/api.js
+++ b/api/api.js
@@ -28,12 +28,9 @@ common.processRequest = processRequest;
 // OS temp directory
 const uploadDir = uploadTemp.resolveUploadDir(countlyConfig.api);
 
-// An SDK write rejected for a bad app key or a failed checksum is cancelled
-// before it is dispatched, so any file parsed out of its body can never be
-// claimed by a handler and is safe to remove here
-plugins.register("/sdk/cancel", function(ob) {
-    uploadTemp.discardUploads(ob.params);
-});
+// Core endpoints that read params.files. Plugins declare their own.
+plugins.uploadPaths.push({path: "/i/apps/create"});
+plugins.uploadPaths.push({path: "/i/apps/update"});
 
 if (cluster.isMaster) {
     console.log("Starting Countly", "version", versionInfo.version, "package", pack.version);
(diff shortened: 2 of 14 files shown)


ar2rsawseen:
Merge order update: the companion PR Countly/countly-enterprise-plugins#3324 now initialises `plugins.uploadPaths` defensively before pushing to it, so its declarations are inert against a core that predates the registry.

That means the two no longer have to land in the same window — **merge the enterprise PR first, then this one**. Landing this one first would still leave the twelve enterprise upload endpoints undeclared and therefore refused, so the order matters even though the coupling is no longer simultaneous.

For context on why: with no defensive init, `plugins.uploadPaths.push(...)` threw at plugin load against an older core, and `pluginManager` responds to a throwing plugin by skipping it and marking it disabled in the database — so nine plugins switched off rather than one endpoint misbehaving.

ar2rsawseen:
Correcting my earlier comment on merge order — it had the process backwards.

The right order is **this PR first**, then a new EE test image is built, and then Countly/countly-enterprise-plugins#3324 can pass its tests and be merged. EE CI takes core from the `countly-ee-core:pipelines-*` container image, so the image rebuild after this merges is what makes the enforcing core available to that PR. My earlier comment said to merge the enterprise PR first; please disregard that.

One thing to be aware of in the interim: between this merging and #3324 merging, the twelve enterprise upload paths are undeclared against an enforcing core, so EE master's upload tests will fail and a master+master deployment would refuse those uploads. The defensive initialisation in #3324 covers the opposite direction (enterprise declarations against a core that predates the registry), not this one.

kanwarujjaval (approved):

PR #7861, opened 2026-07-29
security(hooks): scope internal events to the hook's own apps, and validate the trigger on save

ar2rsawseen:
Reported via `security@count.ly`. Assessed against master `321a6645591e21c6e41f9834614d56126b780f45`.

## Context

The June global-event fix (`59e4ec2f9e5`, `53fefeddc0d`) gated five instance-wide event types behind global-admin ownership via `GLOBAL_EVENT_TYPES`. The remaining `InternalEventTrigger` cases were left untouched, and several of them never consult `rule.apps`. A hook created by a non-global member with `hooks` rights on one app therefore received other apps' data.

Rather than extend the allowlist a third time, **every case now either appears in `GLOBAL_EVENT_TYPES` or checks the event's app id against the hook's apps.** That invariant is stated in the comment above the map so the next event added has to pick one.

## Per-event changes

**`/i/apps/create` becomes a global event type.** It cannot be meaningfully app-scoped, because the app does not exist when the hook is configured, so no hook can legitimately name it. The payload is the new app document, carrying the SDK `key`, the immutable `id_key`, the accepted `keys[]` array and the **checksum salt** (`checksum_salt` is an accepted field on `createApp` and nothing strips it before dispatch). It was the only `/i/apps/*` case that deliberately bypassed the sibling app check. Global-admin-owned hooks still receive it.

**`/i/remote-config/*` (six events)** are delivered only when the event's app is one the hook is scoped to. Five of the six dispatch sites carried no app identifier at all, so they now include `appId`, copied onto the dispatched object rather than set on the stored document. A dispatch with no app id is dropped rather than broadcast.

**`/cohort/enter` and `/cohort/exit`** matched on `rule.trigger.configuration.cohortID` alone. That value comes from the hook's own configuration and is not validated on save, and the handler then reads `app_users<cohort.app_id>` directly, so a hook scoped to one app could receive another app's cohort definition and full app-user documents. The cohort's app must now be one the hook is scoped to.

**`/hooks/trigger`** matched on `rule.trigger.configuration.hookID` alone, equally unvalidated. `fetchRules` loads hooks with only `error_logs` excluded, so the forwarded `ob.rule` is the source hook's **entire document** — every effect configuration, including `HTTPEffect` url and headers, `CustomCodeEffect` code and `EmailEffect` recipients — plus the data that fired it. The source hook must now be scoped within the subscribing hook's apps.

**`/alerts/trigger`** fired every subscribed hook on the instance for any app's alert. Its dispatch carried no arguments at all, so it now passes the alert's app id and alert id and requires the app to be one the hook is scoped to. The effect payload stays empty as before; the new fields are used for scoping only and are not forwarded.

The frontend keeps its own copy of the global-event list for hiding options from non-global admins, so `/i/apps/create` is added there too, with a note to keep the two in sync.

## Tests

`plugins/hooks/tests/internal_event_scope.js` exercises `process()` directly rather than over HTTP, because delivery is driven by `plugins.dispatch` from other plugins and "the hook did not fire" is not observable through the API.

**8 of the 14 fail without this change.** The six that pass either way are the positive controls: a global-admin-owned hook still receives `/i/apps/create`, same-app remote-config still delivers, same-app hook chaining still delivers, and `/crashes/new` scoping is unaffected.

## Not changed here

The `/i/apps/*` and `/i/app_users/*` cases compare `rule.apps[0]` rather than testing membership, so a hook targeting two apps only ever fires for the first. That is a narrowness bug rather than a disclosure, and fixing it would start firing hooks that are currently silent, so it belongs in its own change.


ar2rsawseen:
### Scope of this PR widened

[#7866](https://github.com/Countly/countly-server/pull/7866) is now merged into this branch and closed, so the whole area gets one review. The two changes are the same feature seen from both ends:

- **Delivery time** (original scope): when an internal event fires, only deliver it to hooks scoped to the event's app. Global events reach global-admin-owned hooks only.
- **Save time** (from #7866): reject a hook subscribing to an unknown event type, or naming a cohort, source hook or alert that belongs to an app the hook is not scoped to.

The save-time check is what stops the bad configuration existing; the delivery-time check is what makes the ones already in the database harmless. Neither is sufficient alone, which is why they land together.

Also included, since it came out of reviewing this area:

- The `/alerts/trigger` dispatch in `plugins/alerts/api/parts/common-lib.js` now carries `appId`, so alert-driven hooks can be scoped at all. Without it every hook on the instance fired on any app's alert. (The second producer, `concurrent_users`, is enterprise-only and fixed in Countly/countly-enterprise-plugins#3332.)
- Drops are no longer silent. A dispatch carrying no app id **warns** once, naming the event and how many hooks will not fire, because that is a producer bug and it is how a missing payload quietly disabled working hooks. A dispatch for an app the hook is not scoped to logs at **debug**, since that is ordinary filtering and constant.

### Tests

Three suites: `internal_event_scope.js` (13 cases, no server needed, drives the trigger directly), `internal_event_delivery.js` (end to end through the API), `trigger_config_authz.js` (save-time rejection, end to end).

ar2rsawseen:
### This is the review anchor for the whole change set

Everything substantive about internal event hooks is now in this one PR. The other four are ports or single-line consequences, so reviewing this once covers the feature.

| Repo | Branch | PR | What it is |
|---|---|---|---|
| countly-server | `master` | **this one** | the change, with reasoning and tests |
| countly-server | `release.24.05` | #7863 | same change, tests appended to `plugins/hooks/tests.js` instead of a `tests/` dir |
| countly-platform | `main` | Countly/countly-platform#918 | same change, plus both alert producers, since that repo bundles them |
| countly-enterprise-plugins | `master` | Countly/countly-enterprise-plugins#3332 | six lines: the `concurrent_users` alert dispatch |
| countly-enterprise-plugins | `release.24.05` | Countly/countly-enterprise-plugins#3333 | same six lines |

Three previously separate PRs are folded in and closed: #7866 and #7867 (save-time trigger validation) and Countly/countly-platform#920.

### What is in scope here, and why it is one thing

- **Delivery time**: an internal event only reaches hooks scoped to the event's app. Global events reach global-admin-owned hooks only.
- **Save time**: a hook cannot subscribe to an unknown event type, or name a cohort, source hook or alert belonging to an app it is not scoped to. Save-time stops bad configuration being created; delivery-time makes what is already stored harmless. Neither is sufficient alone.
- **Producers**: scoping is only possible if the event carries an app id. `/alerts/trigger` was dispatched with no payload at all, so the `plugins/alerts` producer is fixed here. The `concurrent_users` producer is enterprise-only, hence the two EE PRs.
- **Observability**: a dispatch with no app id warns once, naming the event and how many hooks will not fire. Out-of-scope filtering logs at debug. A target that cannot be resolved logs why.

### The one ordering constraint

If this merges and Countly/countly-enterprise-plugins#3332 does not, Enterprise loses `concurrent_users` alert hooks, because the core will require an app id that EE is not yet sending. The warning above makes that loud rather than silent, but the hooks do stop until #3332 lands.

kanwarujjaval (approved):

PR #7863, opened 2026-07-29
security(hooks): scope internal events to the hook's own apps, and validate the trigger on save (24.05)

ar2rsawseen:
Backport of #7861 to `release.24.05`.

## What changes

Every case in `InternalEventTrigger.process()` now either appears in `GLOBAL_EVENT_TYPES`, and so reaches only hooks owned by a global admin, or checks the event's app id against the hook's own `apps` list. Several cases previously did neither, so a hook belonging to one app could be delivered another app's event data.

- **`/i/apps/create`** becomes a global event type. It cannot be app-scoped, since the app does not exist when the hook is configured, and it was the only `/i/apps/*` case that bypassed the sibling app check.
- **The six `/i/remote-config/*` events** are delivered only for an app the hook is scoped to. Five of the six dispatch sites carried no app identifier, so they now include `appId`, copied onto the dispatched object rather than set on the stored document. A dispatch without an app id is dropped.
- **`/cohort/enter` and `/cohort/exit`** matched on a `cohortID` taken from the hook's own configuration, which is not validated on save. The cohort's app must now be one the hook is scoped to.
- **`/hooks/trigger`** matched on an equally unvalidated `hookID`. The source hook must now be scoped within the subscribing hook's apps.
- **`/alerts/trigger`** had no scoping and its dispatch carried no arguments, so it now passes the alert's app and alert id and requires the app to be in scope. The effect payload stays empty as before; the new fields are used for scoping only and are not forwarded.

The frontend keeps its own copy of the global-event list for hiding options from non-global admins, so `/i/apps/create` is added there too.

## Adapted for this branch

The hooks tests live in a single `plugins/hooks/tests.js` here rather than a `tests/` directory, so the new describe block is appended there instead of added as a separate file.

## Tests

The new cases exercise `process()` directly, since delivery is driven by `plugins.dispatch` from other plugins and a hook not firing is not observable over the API. 13 assertions, verified locally against this branch.

## Not changed

The `/i/apps/*` and `/i/app_users/*` cases compare `rule.apps[0]` rather than testing membership, so a hook targeting two apps only fires for the first. That is a narrowness bug rather than a scoping gap, and fixing it would start firing hooks that are currently silent, so it belongs in its own change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 ## Version 24.05.51
+Fixes:
+- [hooks] Internal event hooks are validated on save: an unknown event type is rejected, and an event that names a cohort, hook or alert must name one belonging to the hook's own apps
+
 Enterprise Fixes:
 - [data-manager] Fixed bug where event and view transformations occasionally failed to apply to incoming data
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 ## Version 24.05.51
+Fixes:
+- [hooks] Internal event hooks are now scoped to the apps the hook belongs to: app creation is a global-admin-only event, and remote-config, cohort, alert and hook-chaining events are only delivered when the event's app is one the hook is scoped to
+
 Enterprise Fixes:
 - [data-manager] Fixed bug where event and view transformations occasionally failed to apply to incoming data
 
--- a/plugins/alerts/api/parts/common-lib.js
+++ b/plugins/alerts/api/parts/common-lib.js
@@ -216,7 +216,12 @@ async function trigger(result, log) {
     // increase counter just by date
     await increaseAlertCounter(app, date);
     if (alert.alertBy === "hook") {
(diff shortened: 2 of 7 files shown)


ar2rsawseen:
### Scope of this PR widened

[#7867](https://github.com/Countly/countly-server/pull/7867) is now merged into this branch and closed, mirroring [#7866](https://github.com/Countly/countly-server/pull/7866) into [#7861](https://github.com/Countly/countly-server/pull/7861) on master. Save-time validation and delivery-time scoping are the same feature from both ends, so they get one review.

Full reasoning in #7861.

### Difference from the master PR

This branch keeps hooks tests in a single `plugins/hooks/tests.js` rather than a `tests/` directory, so both suites are appended to that file. It is now 771 lines: the 234-line original, plus the scoping and delivery suites, plus the trigger configuration suite. Verified as a pure append on both sides, with no base lines changed.

kanwarujjaval (approved):

PR #7864, opened 2026-07-29
security(reports): authorize the apps a report targets, on every path that uses them

ar2rsawseen:
## What changes

`/i/reports/create` and `/i/reports/update` authorized the target apps only when the effective `report_type` was `"core"`. A non-core `report_type` names the plugin that owns the report, and those reports were inserted and updated without any authorization of the object they point at.

The authorization path for this already existed but was unreachable. The dashboards plugin implements `/report/authorize`, which checks view access to the dashboard a report renders, while the reports-side caller that invoked it (`validateNonCoreUser`) had been commented out.

This restores that caller and gates both create and update on its result.

- **Update authorizes the merged stored-plus-payload report**, not the payload alone, so a partial update cannot leave an unauthorized target in place and repointing an existing report at a different target is checked too.
- A copy is passed, so the `authorized` flag — which is only how the dispatch returns its result — never reaches the written document.
- **Fails closed**, matching the original intent of the commented-out code: a report type whose plugin does not answer `/report/authorize` is not authorized. Only `"dashboards"` implements it today, here and in countly-platform, so a plugin adding a new report type needs to implement the event for it.

Also drops `validateCoreUser`, which was commented out and superseded by the inline per-app check that uses the reports feature permission rather than plain app membership.

## Tests

Real end-to-end coverage through the HTTP endpoints with a real non-admin member, so the whole path runs rather than the branch in isolation:

- a member with reports rights but no view access to a private dashboard is refused when scheduling a report against it
- the same member succeeds for a dashboard they own, which proves the authorize dispatch resolves rather than the request merely failing
- the `authorized` flag does not appear on the stored report
- a core report for an app the member has rights on still succeeds, so the restructuring did not change the core path

Not executed locally, since the harness needs `COUNTLY_TEST_API_KEY_ADMIN` and `COUNTLY_TEST_APP_ID`. CI runs them in `test-api-plugins`.

## Second commit: the paths that were still owner-only

`recordUpdateOrDeleteQuery` adds only `user`, so beyond create and update, four paths were scoped by ownership alone. Owning a report is not the same as still being allowed to read the apps it covers. A member who scheduled a report while they held an app, then lost that access while keeping reports rights elsewhere, could:

- re-enable it through `/i/reports/status` and let the schedule resume,
- mail themselves that app's figures on demand through `/i/reports/send`,
- read them straight out of the response through `/i/reports/preview` and `/i/reports/pdf`, which render the report and hand it back. This is the most direct of the four: no email involved, the figures come back in the HTTP response.

Nothing re-checked anything at send time either, so a report enabled before access was revoked kept mailing indefinitely with no further action by anybody. The scheduled job calls `reports.getReport` directly rather than `sendReport`, so both now consult a shared `reports.ownerMayStillSend`; guarding only `sendReport` would have missed the scheduled path entirely, which is the one that matters most.

Deliberately unchanged, with the reasoning recorded in the code:

- **Disabling** stays allowed. It only reduces what the report does, and refusing would leave someone unable to stop mail they no longer want.
- **Delete** stays owner-only. It cannot reach another app's data, and the owner needs a way to clean up.
- `last_sent` is not stamped when a report is skipped, so it resumes on its own if access is restored.

Reports with no apps pass, because those are non-core reports whose target is authorized by their own plugin through `/report/authorize` rather than per app.

A report whose owner no longer exists also stops being sent. A deleted account holds access to nothing, so the same rule applies to it; the `getReport` global-admin fallback stays where it is, because preview and pdf rely on it and both are authorized against the member making the request, but it no longer decides whether scheduled mail keeps going out. This matches alerts, which already stopped in that case.

Also drops a stray `console.log` of the status payload.

### Worth a look separately

That global-admin fallback for an unresolvable owner quietly *raises* the access a report renders with. `/i/users/delete` removes a departing member's reports, so it should only affect orphaned rows, but I left it alone here rather than change send behaviour inside a security fix.

## Tests

Real end-to-end coverage through the HTTP endpoints with a real non-admin member, so the whole path runs rather than the branch in isolation:

- a member with reports rights but no view access to a private dashboard is refused when scheduling a report against it
- the same member succeeds for a dashboard they own, which proves the authorize dispatch resolves rather than the request merely failing
- the `authorized` flag does not appear on the stored report
- a core report for an app the member has rights on still succeeds, so the restructuring did not change the core path

The second commit adds a suite that revokes a member's access to one app mid-test, then covers both directions. The exploit: enable, send and preview are each refused, and the stored report is verified to still be disabled afterwards. The happy paths, which have to keep working: the member can still enable, disable, send and preview reports on the app they do hold, can still switch **off** and delete the revoked-app report, and a global admin is unaffected.

Not executed locally, since the harness needs `COUNTLY_TEST_API_KEY_ADMIN` and `COUNTLY_TEST_APP_ID`. CI runs them in `test-api-plugins`.

## Propagation

Same change is going to `release.24.05` and countly-platform `main`; both carry the identical commented-out caller.


ar2rsawseen:
### Scope widened: a second gap in the same handler

An audit for the class behind #7872 turned up a chain in this handler that the original fix here did **not** close, so it is fixed in this PR rather than a separate one.

The apps list was authorized only when `report_type` was `"core"`, which left a two step route onto any app:

1. **Create** with `report_type: "dashboards"` and `apps: ["<victim>"]`. The apps check never ran. The non-core check this PR originally added authorizes the *dashboard target*, not the apps list, so pointing it at your own dashboard passed.
2. **Update** with `{_id, report_type: "core"}` and no `apps`. `effectiveType` became `"core"`, but the check only looked at the *submitted* apps, which was absent, so the stored list survived unchecked and the report began using it.

Then read it through `/i/reports/preview` or `/i/reports/pdf`, or wait for the scheduled mail.

Precondition was reports create and update on **any single app**, with no prior access to the victim app. I verified both halves against this branch before fixing, so it was live even with the original change.

### The fix

`apps` is authorized whenever a list is present, whatever the report type, and on update the list authorized is the one the report will actually end up with: the submitted list when one is sent, the stored list otherwise.

### Why this does not break dashboards reports

The dashboards drawer sets `showApps = false` and hides the app picker, so a legitimate non-core report carries `apps: []`. An empty or absent list stays permitted, which is also what keeps `report_type: "dashboards"` reports for **shared** dashboards working, including the case where the member has no access to the apps whose widgets the dashboard contains.

The shared helper keeps the legacy allowance the surrounding code already makes, so members with no `permission` object remain authorized through `user_of`.

### Tests

Two cases walking each half of the chain: create with an unauthorized apps list is refused, and converting a stored report to core is refused when its stored apps name an app the member cannot read. A second app the member has no rights on is created as a fixture and cleaned up.

ar2rsawseen:
### Review pass: every suppressed note on this PR

This PR has no inline comments, but the five Copilot reviews carry ten suppressed notes between them. Went through all of them, plus the two inline notes the countly-platform twin (Countly/countly-platform#919) received, since both repositories run the same handler. Fixed in 163d7cd3cb3.

**Two fail-opens, both found through the review notes rather than stated in them**

- **The authorize flag came from the request.** `props` is `params.qstring.args`, and `validateNonCoreUser` read `props.authorized` after the dispatch without clearing it first. The dashboards handler only assigns the flag on the path where it resolves the dashboard — it resolves without touching it when the dashboard is missing or the lookup errors, and for a `report_type` no plugin owns no handler runs at all. On any of those, a caller sending `authorized: true` answered the non-core check for themselves. On update it also reached `$set`, because authorization ran on `mergedReport` while the write used the original `props`, so it persisted and kept answering on later updates. Now stripped from the request in create and update, cleared before the dispatch, and only an explicit `true` from a plugin authorizes.

- **`/i/reports/status` and the sender disagreed about `enabled`.** The authorization matched `true` and `"true"`; the send job tests `report.enabled + "" !== "false"`. So `{"<id>": 1}` switched a report back on with the apps behind it never checked — a bypass of the enable authorization this PR adds. One helper now decides what the request asks for, both the check and the write read it, and a strict boolean is stored.

Also: an `apps` value that is not a list was read as "no apps" and permitted, then stored unchecked, and an update flipping `report_type` to `"core"` started using it. A submitted non-list is now a 400, and `appsArePermitted` / `ownerMayStillReadApps` refuse one from any source, including a stored document.

**Suppressed notes — addressed**

| Note | Outcome |
|---|---|
| `reports.js:186` — `sendLoadedReport` reads `ob.message` on the error path (raised twice) | Fixed. `getReport` reports some failures with the error alone (`return callback(err2)`), so there is no second argument. Guarded here and in `jobs/send.js`, which does the same with `ob.report`. |
| `reports.js:162/164` — `sendReport` returns `(null, null)` when `ownerMayStillSend` is false and the endpoint answers "Success" (raised twice) | Fixed. The skip is reported as a third callback argument and `/i/reports/send` says so; `generate_reports.js` and `process_reports.js` read only `(err, res)` and are unaffected. |
| `api.js:474/654` — send error logged as `err` instead of `err2` | Fixed. |
| `api.js:644` — status handler double-responds on a bulk error / never responds when there are no ops | Fixed: it returns after the error, and answers `400` when there is nothing to act on rather than leaving the request open. |
| `api.js:609` — `for (const id in statusList)` over a possibly-missing `args` | Fixed by construction: `args` is normalized to an object and the ids come from `Object.keys`, so nothing walks a prototype or a string's indices. |
| `tests.js:505/584` — setup step does not assert success | Fixed: the admin update that plants the apps list asserts `200` and propagates. |
| `tests.js:588` — cleanup does not wait for the victim app delete | Fixed: it is chained ahead of `done()`. |

**Suppressed notes — not defects**

- `api.js:640/655` and `api.js:621/644/648` — *"`common.db.ObjectID` throws on a malformed id, so a crafted `args` can crash the handler / should be a 400 not a 500."* It does not throw. The `pluginManager.js` wrapper catches and returns the id unchanged:

  ```js
  countlyDb.ObjectID = function(id) {
      try {
          return mongodb.ObjectId(id);
      }
      catch (ex) {
          logDbRead.i("Incorrect Object ID %j", ex);
          return id;
      }
  };
  ```

  A malformed id therefore becomes a query that matches nothing — the silent no-op the surrounding comment already describes for unknown ids — and neither the `try/catch` 500 nor an unhandled rejection is reachable that way. No id validation added; the real problem in that block was the missing response.

**Tests added** — a forged `authorize` flag on both unprotected paths, a non-list `apps`, enabling with a truthy non-boolean, and the status endpoint answering when there is nothing to act on.

**Verified** — `node --check` and eslint clean on all four changed files. The reports plugin suite was not run here: it needs backing services on the isolated test ports, which are not up in this environment.

Mirrored in countly-platform#919 (which also needed `appsArePermitted` aligned onto the reports read right, so the helper is identical in both repositories again) and in #7865, which additionally needed the missing half of this fix ported — see the note there.

ar2rsawseen:
### Propagated from #7865: the two failing `Testing Reports after access is revoked` cases

Two P1 review notes landed on the 24.05 twin about `test-api-plugins` failures. Both cases fail identically here, and both were already failing before the review changes went on this branch — see the run on `9ef402fe` — so neither is a regression, but they block the PR either way. Fixed in `caab994c2a0`.

**`should not have enabled the report despite the attempts`** — `Cannot read properties of undefined (reading 'should')`. The report is found; the field is not. `/i/reports/create` never writes `enabled`, and every attempt to switch it on in this suite is correctly refused — which is what the assertion is checking — so the document reaches it with no `enabled` at all and `found.enabled.should` dereferences undefined. The passing case was the crashing case. Now compared rather than dereferenced, so an absent field reads as "not enabled" and a real regression fails as an assertion.

**`should still let the member send and preview a report on the app they hold`** — 50s mocha timeout. The shared `newReport` fixture asks for `star-rating` and `performance`, neither of which the reports plugin resolves itself, so `getReport` dispatches `/email/report` for each with a 30 second cancel timeout when nothing answers — which is the case in the plugin test run, hence `cancel report plugin dispatcher:performance` in the log right before the timeout. This case renders twice, to send and to preview, so it waits ~60s against a 50s cap; every other send/preview case here is refused before rendering, which is why only this one times out. The reports in this describe now ask only for `analytics`, which resolves in process.

The 30s-per-unanswered-metric wait itself is left alone — pre-existing renderer behaviour, not something this PR touches, and the `Testing Report CRUD` cases that pay it once still pass inside the cap. Worth its own issue.

Full reasoning: https://github.com/Countly/countly-server/pull/7865#discussion_r3845733691 and https://github.com/Countly/countly-server/pull/7865#discussion_r3845733970.

For the record, every test added in the previous round passed on this run: the forged authorize flag, the non-list `apps`, the truthy non-boolean enable, and the status endpoint answering when there is nothing to act on.

Cookiezaurs (approved):

PR #7865, opened 2026-07-29
security(reports): authorize the apps a report targets, on every path that uses them (24.05)

ar2rsawseen:
Backport of #7864 to `release.24.05`. Applied cleanly; only the CHANGELOG heading needed resolving.

## What changes

`/i/reports/create` and `/i/reports/update` authorized the target apps only when the effective `report_type` was `"core"`. A non-core `report_type` names the plugin that owns the report, and those reports were inserted and updated without any authorization of the object they point at.

The authorization path already existed but was unreachable: the dashboards plugin implements `/report/authorize`, which checks view access to the dashboard a report renders, while the reports-side caller that invoked it (`validateNonCoreUser`) had been commented out.

Restores that caller and gates create and update on its result. Update authorizes the merged stored-plus-payload report, so a partial update cannot leave an unauthorized target in place and repointing an existing report is checked too. A copy is passed so the `authorized` flag never reaches the written document. Fails closed, matching the original intent: a report type whose plugin does not answer the event is not authorized, and only `"dashboards"` implements it today.

Also drops `validateCoreUser`, which was commented out and superseded by the inline per-app check.

## Tests

Real end-to-end coverage through the HTTP endpoints with a real non-admin member: refused for a private dashboard they cannot view, allowed for one they own, the `authorized` flag absent from the stored report, and a core report still succeeding as a control.

Not executed locally, since the harness needs `COUNTLY_TEST_API_KEY_ADMIN` and `COUNTLY_TEST_APP_ID`. CI runs them in `test-api-plugins`.

## Second commit: the paths that were still owner-only

`recordUpdateOrDeleteQuery` adds only `user`, so beyond create and update the remaining paths were scoped by ownership alone. Owning a report is not the same as still being allowed to read the apps it covers. A member who scheduled a report while they held an app, then lost that access while keeping reports rights elsewhere, could:

- re-enable it through `/i/reports/status` and let the schedule resume,
- mail themselves that app's figures on demand through `/i/reports/send`,
- read them straight out of the response through `/i/reports/preview` (and `/i/reports/pdf` where that endpoint exists). This is the most direct of them: no email involved, the figures come back in the HTTP response.

Nothing re-checked anything at send time either, so a report enabled before access was revoked kept mailing indefinitely. The scheduled job calls `reports.getReport` **directly** rather than `sendReport`, so both now consult a shared `reports.ownerMayStillSend`; guarding only `sendReport` would have missed the scheduled path entirely, which is the one that matters most.

Deliberately unchanged, with the reasoning recorded in the code:

- **Disabling** stays allowed. It only reduces what the report does, and refusing would leave someone unable to stop mail they no longer want.
- **Delete** stays owner-only. It cannot reach another app's data, and the owner needs a way to clean up.
- `last_sent` is not stamped when a report is skipped, so it resumes on its own if access is restored.

Reports with no apps pass, because those are non-core reports whose target is authorized by their own plugin through `/report/authorize` rather than per app.

A report whose owner no longer exists also stops being sent. A deleted account holds access to nothing, so the same rule applies to it; the `getReport` global-admin fallback stays where it is, because preview and pdf rely on it and both are authorized against the member making the request, but it no longer decides whether scheduled mail keeps going out. This matches alerts, which already stopped in that case.

Also drops a stray `console.log` of the status payload.

### Worth a look separately

That global-admin fallback for an unresolvable owner quietly *raises* the access a report renders with. `/i/users/delete` removes a departing member's reports, so it should only affect orphaned rows, but I left it alone here rather than change send behaviour inside a security fix.

### Tests for this commit

A suite that revokes a member's access to one app mid-test, covering both directions. The exploit: enable, send and preview are each refused, and the stored report is verified to still be disabled afterwards. The happy paths, which have to keep working: the member can still enable, disable, send and preview reports on the app they do hold, can still switch **off** and delete the revoked-app report, and a global admin is unaffected.

This branch has no `/i/reports/pdf` endpoint, so that guard from #7864 does not apply here. The inline allow-list the create and update paths built is extracted into a shared `appsArePermitted`, matching how master expresses it, rather than repeated three more times.

One test assertion differs from master deliberately: the preview happy path asserts "not 401" rather than a specific status, because this branch renders with `if (params && params.res) { var html = res.message; ... }` and lacks master's `else if (res)` guard, so a report for an app with no data can fail inside rendering. That is unrelated to authorization and is left alone.

### DIFF
--- a/plugins/reports/tests.js
+++ b/plugins/reports/tests.js
@@ -337,6 +337,7 @@ describe('Testing Reports non-core authorization', function() {
     var sharedDashIdForCleanup = "";
     var memberApiKey = "";
     var memberUserId = "";
+    var VICTIM_APP_ID = "";
     var uniq = Date.now();
 
     /**
@@ -404,6 +405,20 @@ describe('Testing Reports non-core authorization', function() {
             });
     });
 
+    it('should create a second app the member has no rights on', function(done) {
+        request.get('/i/apps/create?api_key=' + API_KEY_ADMIN
+            + '&args=' + encodeURIComponent(JSON.stringify({name: "reportsVictim" + uniq, country: "TR", type: "mobile", category: "6", timezone: "Europe/Istanbul"})))
+            .expect(200)
+            .end(function(err, res) {
+                if (err) {
+                    return done(err);
+                }
+                VICTIM_APP_ID = res.body._id;
+                should.exist(VICTIM_APP_ID);
+                done();
+            });
+    });
+
     it('should reject a dashboards report for a dashboard the member cannot view', function(done) {
         request.get('/i/reports/create?api_key=' + memberApiKey + '&app_id=' + APP_ID
             + '&args=' + encodeURIComponent(JSON.stringify(dashboardReport(adminDashboardId))))
@@ -433,6 +448,65 @@ describe('Testing Reports non-core authorization', function() {
                         }
                         r.body.should.have.property('result', 'Success');
                         done();
+                    });
+            });
+    });
+
+    it('should refuse a non-core report naming an app the member cannot read', function(done) {
+        // apps used to be authorized only for core reports, so a non-core report could
+        // carry any apps list at all
+        var sneaky = Object.assign({}, dashboardReport(memberDashboardId), {apps: [VICTIM_APP_ID]});
+        request.get('/i/reports/create?api_key=' + memberApiKey + '&app_id=' + APP_ID
+            + '&args=' + encodeURIComponent(JSON.stringify(sneaky)))
+            .expect(401)
+            .end(function(err) {
+                if (err) {
+                    return done(err);
+                }
+                done();
+            });
+    });
+
+    it('should refuse to convert a stored report onto an app the member cannot read', function(done) {
+        // the second half of the chain: store apps while the type is non-core, then flip
+        // report_type to core with apps omitted so the stored list survives unchecked
+        var report = Object.assign({}, dashboardReport(memberDashboardId), {title: "convert-" + uniq});
+        request.get('/i/reports/create?api_key=' + memberApiKey + '&app_id=' + APP_ID
+            + '&args=' + encodeURIComponent(JSON.stringify(report)))
+            .expect(200)
+            .end(function(err) {
+                if (err) {
+                    return done(err);
+                }
+                request.get('/o/reports/all?api_key=' + memberApiKey + '&app_id=' + APP_ID)
+                    .expect(200)
+                    .end(function(e, r) {
+                        if (e) {
+                            return done(e);
+                        }
+                        var created = (r.body || []).filter(function(x) {
+                            return x.title === "convert-" + uniq;
+                        })[0];
+                        should.exist(created);
+                        // an admin plants the unauthorized apps list directly, standing in
+                        // for the first half of the chain now that create refuses it
+                        request.get('/i/reports/update?api_key=' + API_KEY_ADMIN + '&app_id=' + APP_ID
+                            + '&args=' + encodeURIComponent(JSON.stringify({_id: created._id, apps: [VICTIM_APP_ID]})))
+                            .expect(200)
+                            .end(function(e1) {
+                                if (e1) {
+                                    return done(e1);
+                                }
+                                request.get('/i/reports/update?api_key=' + memberApiKey + '&app_id=' + APP_ID
+                                    + '&args=' + encodeURIComponent(JSON.stringify({_id: created._id, report_type: "core"})))
+                                    .expect(401)
+                                    .end(function(e2) {
+                                        if (e2) {
+                                            return done(e2);
+                                        }
+                                        done();
+                                    });
+                            });
                     });
             });
     });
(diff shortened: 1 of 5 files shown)


ar2rsawseen:
### This backport was missing half the fix

Checking the review notes across this pair turned up something bigger than the notes. Master's `247b2c7f2af security(reports): authorize the apps a report will actually have` was never carried onto this branch, so the two step chain #7864 describes in its scope-widening note was **still open on release.24.05**:

1. **create** with `report_type: "dashboards"` and `apps: ["<victim>"]` — `apps` was authorized only inside the `report_type === "core"` arm, so the list was stored with nothing checking it while the non-core check looked at the dashboard instead;
2. **update** with `{_id, report_type: "core"}` and no `apps` — the update authorized only the *submitted* list, so an absent one left the stored list in place and `effectiveType` became `"core"`, and the report started using it.

Then read it through `/i/reports/preview` or `/i/reports/pdf`, or wait for the scheduled mail. Precondition is reports create and update on any single app, with no prior access to the victim app — the same as on master.

Ported in 97f4cce9cae:

- create authorizes `apps` whatever the report type;
- update authorizes the list the report will actually end up with — the submitted one when there is one, the stored one otherwise;
- both arms' inline allow-lists are replaced by the `appsArePermitted` helper this branch already had for the enable/send/render paths;
- the four tests master covers this with are ported alongside (the second app fixture, the non-core apps refusal, the convert-to-core refusal, and the cleanup for the fixture).

### Review notes

The same commit carries everything from #7864's review pass, which applies unchanged here — the caller-supplied `authorized` flag (a real fail-open: the dashboards handler leaves the flag untouched when the dashboard is missing or the lookup errors, and no handler runs at all for an unrecognised type), the `enabled` value the status endpoint and the sender read differently, the non-list `apps`, the `err`/`err2` log, the skipped-send reported as "Success", the `ob` dereference on `getReport`'s error-only callbacks, the status handler's double and missing responses, and the two test-hygiene notes. Full accounting in https://github.com/Countly/countly-server/pull/7864#issuecomment-5398377132, including the one class of note that turned out not to be a defect (`common.db.ObjectID` does not throw on a malformed id — the wrapper catches and returns the string).

`plugins/reports/tests.js` is now identical to master's apart from the one deliberate difference already carried on this branch: the preview assertion is `should.not.equal(401)` rather than `expect(200)`, because the renderer here can fail inside rendering for an app with no data, which is unrelated to authorization.

**Verified** — `node --check` and eslint clean on all four changed files; the plugin suite was not run here (the isolated-port backing services are not up in this environment).

plugins/reports/tests.js
```diff
@@ -316,3 +316,708 @@ describe('Testing Reports', function() {
...
+            .expect(200)
+            .end(function(err) {
+                if (err) {
+                    return done(err);
+                }
+                // asserted as "not refused" rather than a specific status: the preview
+                // renderer on this branch dereferences res.message without checking res,
+                // so a report for an app with no data can fail inside rendering. That is
+                // unrelated to authorization, which is what this test is about.
+                request.get('/i/reports/preview?api_key=' + memberApiKey + '&app_id=' + APP_B + '&args=' + encodeURIComponent(JSON.stringify({_id: reportOnB})))
```
ar2rsawseen, line 951: [P1] Avoid exercising a renderer that may never respond in this authorization test

This release branch's preview renderer can leave the request open when the app has no reportable data. Dropping the expected status does not avoid that behavior: Supertest still waits for the response, and the current `test-api-plugins` job times out here after 50 seconds. Please seed enough data for preview to complete, or verify the authorization happy path without invoking this renderer.

plugins/reports/tests.js
```diff
@@ -316,3 +316,708 @@ describe('Testing Reports', function() {
...
+            .expect(200)
+            .end(function(err, res) {
+                if (err) {
+                    return done(err);
+                }
+                var found = ((res.body) || []).filter(function(r) {
+                    return r._id + "" === reportOnA;
+                })[0];
+                should.exist(found);
+                found.enabled.should.not.equal(true);
```
ar2rsawseen: [P1] Make the stored-report assertion robust

The current `test-api-plugins` run reaches this line with `found === undefined`, so the test fails with `Cannot read properties of undefined (reading 'should')` instead of verifying the status behavior. Please make the fixture/list lookup deterministic and explicitly fail when the report is missing before asserting `enabled`.

plugins/reports/tests.js
ar2rsawseen: Fixed in 1ec045b21c4 — though the cause is not the renderer hanging, and that changed which remedy made sense.

The renderer does answer. What it does first is wait. The shared `newReport` fixture asks for `star-rating` and `performance`, neither of which the reports plugin resolves itself, so `metricsToCollections` passes them straight through and `getReport` dispatches `/email/report` for each with a 30 second cancel timeout for the case where nothing answers:

```js
// set plugin report dispatch max duration to 30s
const cancelReportCallTimeout = setTimeout(() => {
    done2("cancel report plugin dispatcher:" + metric, null);
}, 30000);
```

Nothing answers in the plugin test run, which is why the CI log ends with `cancel report plugin dispatcher:performance` immediately before the timeout. The two dispatches run in parallel, so one render costs about 30s — and this test renders twice, once to send and once to preview. About 60s against a 50s cap.

That also explains why only this case times out: every other send/preview case here is refused by the authorization check before any rendering starts.

So neither of the two remedies you suggest is quite the fit. Seeding data would not help, since the wait is for a plugin that never replies rather than for data; and dropping the renderer would drop the only place the happy path is exercised end to end. Instead the reports in this describe now ask only for `analytics`, which resolves in process:

```js
metrics: {analytics: true},
```

Two fast renders, well inside the cap, and the case still asserts exactly what it is about: that the member is **not** refused.

The 30s-per-unanswered-metric wait is left as it is. It is pre-existing renderer behaviour rather than anything this PR touches, and the two `Testing Report CRUD` cases that pay it once still pass inside the cap — worth its own issue rather than a change smuggled into a security backport.

Same fix on #7864 and Countly/countly-platform#919, where the case fails identically.

plugins/reports/tests.js
ar2rsawseen: Fixed in 1ec045b21c4 — one correction to the diagnosis, which changes the fix.

`found` is not undefined. The error is `Cannot read properties of undefined (reading 'should')`, and the stack points at column 31 of

```js
found.enabled.should.not.equal(true);
```

which is `.should`, not `.enabled` — so the lookup did find the report, and what is undefined is `found.enabled`. Had `found` itself been missing, `should.exist(found)` on the line above would have failed first, with a readable message.

The field is absent because nothing ever writes it: `/i/reports/create` does not set `enabled`, and every attempt to switch it on in this suite is correctly refused, which is the very thing the assertion is checking. So the passing case is exactly the case that crashes.

Compared instead of dereferenced:

```js
should.exist(found);
//a report that was never switched on carries no enabled field at all -
//create does not write one - so compare rather than dereference. Reading
//.should off undefined turned this into a TypeError instead of a verdict.
(found.enabled === true).should.equal(false);
```

An absent field now reads as "not enabled", a report that really was enabled fails as an assertion rather than a TypeError, and the missing-report case keeps failing at `should.exist` where it is legible.

Same fix on #7864 and Countly/countly-platform#919, where the case fails identically.

Cookiezaurs (approved):

PR #7866, opened 2026-07-29
fix(hooks): validate internal event trigger configuration on save

ar2rsawseen:
## What changes

`/i/hook/save` validated the hook's `apps` but never its trigger configuration. The event type was not checked at all, so any string was accepted and stored, and the events that name a target object matched on that id alone at delivery time, with nothing tying the target to the hook's own apps.

Validation is added on both create and update:

- the event type must be one of the registered internal events
- **`/hooks/trigger`** must name a source hook whose apps are within this hook's apps. A subset rather than merely overlapping, because the delivered payload includes the source hook's whole document
- **`/cohort/enter`** and **`/cohort/exit`** must name a cohort belonging to one of the hook's apps
- **`/alerts/trigger`**, when it names an alert, must name one belonging to one of the hook's apps. Not required to name one, since the field is optional today

Update validates the **effective** hook, payload merged over stored, so changing only the trigger or only the apps is still checked, and retargeting an existing chain at another app's hook is rejected.

## Two judgement calls worth reviewing

**Cohorts fail open, hooks and alerts fail closed.** `hooks` and `alerts` are core collections, so an unresolvable target is rejected. Cohorts belong to an enterprise plugin and the collection may be absent in a given deployment, so an unresolvable cohort is allowed through and left to the delivery-time check rather than blocking a working feature we cannot verify here.

**Not exempt for global admins.** The delivery-side check is not exempt either, so exempting save would let a global admin store a hook that can never fire. That is the failure mode this removes: such a hook was previously accepted and then silently did nothing. A clear 400 is better than a dead hook.

## Relationship to #7861

#7861 scopes delivery. This scopes the save. They are independent and either can merge first; this one only adds an exported accessor to `internal_event.js` alongside the existing `GLOBAL_EVENT_TYPES` export.

## Tests

Through the real save endpoint: unknown event type rejected, cross-app chain rejected, chain with no source rejected, same-app chain allowed, an event carrying no target id allowed, and retargeting an existing chain at another app's hook rejected on update.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Version 25.03.xx
 Fixes:
+- [hooks] Internal event hooks are validated on save: an unknown event type is rejected, and an event that names a cohort, hook or alert must name one belonging to the hook's own apps
 - [events] Fix sum chart tooltip displaying the raw floating-point value instead of a rounded number
 
 ## Version 25.03.50
--- a/plugins/hooks/api/api.js
+++ b/plugins/hooks/api/api.js
@@ -347,6 +347,16 @@ plugins.register("/i/hook/save", function(ob) {
                         common.returnMessage(params, 403, "User does not have right");
                         return true;
                     }
+                    //validate against the effective hook, since an update may
+                    //change only the trigger or only the apps
+                    const updatedTriggerValidation = await validateTriggerConfiguration(
+                        hookConfig.trigger || existingHook.trigger,
+                        hookConfig.apps || existingHook.apps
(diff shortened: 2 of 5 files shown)


ar2rsawseen:
Folded into #7861 so this feature area gets one review instead of two. Same commits, merged into that branch, nothing dropped: `validateTriggerConfiguration` in `plugins/hooks/api/api.js`, the `getInternalEvents()` accessor, and the `trigger_config_authz.js` suite are all present there.

Branch kept in case anything needs to be traced back.

PR #7867, opened 2026-07-29
fix(hooks): validate internal event trigger configuration on save (24.05)

ar2rsawseen:
Backport of #7866 to `release.24.05`.

## What changes

`/i/hook/save` validated the hook's `apps` but never its trigger configuration. The event type was not checked at all, so any string was accepted and stored, and the events that name a target object matched on that id alone at delivery time, with nothing tying the target to the hook's own apps.

Validation is added on both create and update:

- the event type must be one of the registered internal events
- **`/hooks/trigger`** must name a source hook whose apps are within this hook's apps. A subset rather than merely overlapping, because the delivered payload includes the source hook's whole document
- **`/cohort/enter`** and **`/cohort/exit`** must name a cohort belonging to one of the hook's apps
- **`/alerts/trigger`**, when it names an alert, must name one belonging to one of the hook's apps. Not required to name one, since the field is optional today

Update validates the effective hook, payload merged over stored, so changing only the trigger or only the apps is still checked, and retargeting an existing chain at another app's hook is rejected.

`hooks` and `alerts` are core collections, so an unresolvable target is rejected. Cohorts belong to an enterprise plugin whose collection may be absent, so an unresolvable cohort is allowed through and left to the delivery-time check rather than blocking a feature that cannot be verified here.

Not exempt for global admins, matching the delivery-side check. Exempting save would let a global admin store a hook that can never fire, which is the failure mode this removes.

## Adapted for this branch

Two differences from master, both real rather than cosmetic:

- the save handler and its `findOne` callback are callback-style here rather than `async`, so both were made `async` to host the validation. Master's handler is already `async`, so this brings the two in line rather than introducing a new pattern.
- hooks tests live in a single `plugins/hooks/tests.js`, so the new describe block is appended there instead of added as a separate file.

## Tests

Through the real save endpoint: unknown event type rejected, cross-app chain rejected, chain with no source rejected, same-app chain allowed, an event carrying no target id allowed, and retargeting an existing chain at another app's hook rejected on update.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 ## Version 24.05.51
+Fixes:
+- [hooks] Internal event hooks are validated on save: an unknown event type is rejected, and an event that names a cohort, hook or alert must name one belonging to the hook's own apps
+
 Enterprise Fixes:
 - [data-manager] Fixed bug where event and view transformations occasionally failed to apply to incoming data
 
--- a/plugins/hooks/api/api.js
+++ b/plugins/hooks/api/api.js
@@ -295,7 +295,7 @@ plugins.register("/i/hook/save", function(ob) {
     let paramsInstance = ob.params;
 
 
-    validateCreate(ob.params, FEATURE_NAME, function(params) {
+    validateCreate(ob.params, FEATURE_NAME, async function(params) {
         let hookConfig = params.qstring.hook_config;
         try {
             hookConfig = JSON.parse(hookConfig);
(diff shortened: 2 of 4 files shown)


ar2rsawseen:
Folded into #7863, matching what was done on master with #7866 into #7861. Same commits merged into that branch; the trigger configuration suite is appended to `plugins/hooks/tests.js` alongside the scoping suites.

Branch kept in case anything needs to be traced back.

PR #7868, opened 2026-07-30
refactor(dbviewer): validate aggregation operators against an allow-list

ar2rsawseen:
## What

Reworks how DB Viewer validates a user-supplied aggregation pipeline.

The guard decided which nested arrays were sub-pipelines by looking for a recognised stage name inside them, then stripped whatever it disallowed. Two problems with that:

- Correctness depended on the plugin knowing every stage name MongoDB has, including undocumented internal ones. A name it did not recognise changed how a whole branch was classified.
- Stripping meant a caller who asked for something unavailable to them got different results back with no explanation.

Now the guard walks every object and array to any depth and checks only keys beginning with `$` against a per-role allow-list, matched exactly. Nothing is inferred about the pipeline's shape, so there is no shape to get wrong.

Values are never inspected. That is what makes a keys-only check sound: MongoDB reaches a `$`-prefixed *field* name through `$getField` / `$setField`, where the name is an argument value, not a key. So `{ "$literal": "$lookup" }` and `{ "$setField": { "field": "$lookup", ... } }` stay valid, while a key named `$lookup` is an operator every time.

The pipeline is no longer modified. A disallowed operator rejects the request and names the operator and its path, instead of being deleted and the query run anyway. `find()` projections behave the same way now, rather than having offending fields dropped.

## Allow-list vs deny-list

Deliberate. A deny-list fails open: an operator a future MongoDB adds is permitted until someone notices. An allow-list fails safe: the worst case is a valid query rejected, which arrives as a support ticket rather than as a problem.

Lists were verified against a running MongoDB instead of the docs. `verify_operators.js` is that probe, committed so the lists can be re-checked on an upgrade. It found `$sharedDataDistribution` in the old list, which MongoDB spells `$shardedDataDistribution`, so the real stage was never recognised.

## Unchanged

Server-side JavaScript (`$function`, `$accumulator`, `$where`), write stages (`$out`, `$merge`), and joins into redacted collections are refused for every role at any depth, global admins included.

## Also here: the password-reset token

`password_reset.prid` was returned in full. The reset route looks it up directly as `password_reset.findOne({prid})`, so the value is the reset link. It is now withheld on all three read paths and refused as a join target for every role, the same way member credentials already were.

The three redaction sites (single-document read, collection read, aggregation) each carried their own copy of the field list, which is how a collection ends up covered in two of the three. Moved into `parts/redaction.js` as one table, so adding a collection covers all three, with a test asserting the aggregation and document paths cover the same fields so they cannot drift apart.

`auth_tokens` deliberately stays at the call sites: its secret is the `_id`, which cannot be dropped without breaking the row, so the value is replaced instead.

## Tests

56 cases across `plugins.dbviewer.aggregation-guard.js`, `plugins.dbviewer.query-guard.js` and the new `plugins.dbviewer.redaction.js`.

Both existing suites needed rewriting: they asserted the stripping behaviour, and the query-guard one called the removed `sanitizeProjection`, so it failed to even load. Extended on both sides: nesting the old classification handled inconsistently, plus the near-miss cases that must keep working (`$mergeObjects` next to `$merge`, `$literal` and `$setField` with `$`-prefixed values, window functions, real `$facet` branches).

Full unit suite on this branch: 199 passing. The two failures are `Countly Request`, which makes real HTTP calls and fails without network; they fail the same way on a clean checkout.


plugins/dbviewer/api/parts/aggregation_guard.js
```diff
@@ -211,112 +252,94 @@ function findHardViolation(node) {
...
             }
-            var child = value[k];
-            if (Array.isArray(child) && isSubPipeline(child)) {
-                stripStages(child, allowedStages, changes);
-                if (child.length === 0) {
-                    delete value[k];
-                }
+            // require an explicit `true` so inherited Object.prototype keys
+            // (constructor, __proto__, ...) are never treated as allow-listed
+            if (key.charAt(0) === "$" && allowedOperators[key] !== true) {
```
ar2rsawseen: [P1] Do not treat every nested `$`-prefixed key as an operator. Valid MongoDB syntax contains dollar-prefixed option/data keys below operators—for example `$geoWithin: {$centerSphere: ...}`, `$text: {$language: ...}`, and objects intentionally carried through `$literal`. This recursive check rejects all of those before MongoDB sees them, so existing DB Viewer aggregations will start returning 400. The traversal needs operator-aware handling for option and literal subtrees (while continuing to validate actual executable operators).

plugins/dbviewer/api/parts/aggregation_guard.js
```diff
@@ -1,104 +1,179 @@
...
-// anyone, at any depth — they live inside otherwise-allowed stages.
-const DENIED_OPERATORS = {
-    "$function": true,
-    "$accumulator": true,
-    "$where": true
-};
+// Expression operators, accumulators and window operators. These live inside
+// stages rather than being stages, and a blind walk meets them constantly, so they
+// have to be listed or every real query is rejected.
+const EXPRESSION_OPERATORS = [
```
ar2rsawseen, line 91: [P2] Complete the allow-list for the MongoDB versions this release supports before enforcing it. MongoDB 8 aggregation expressions such as `$bitAnd`, `$bitOr`, `$bitXor`, `$bitNot`, and `$toHashedIndexKey` are absent, so otherwise valid pipelines are rejected as disallowed. The committed verifier can only test names already present in this array; it cannot discover omissions, so this list needs an audit against the supported operator catalog plus regression cases for the missing families.

plugins/dbviewer/api/parts/verify_operators.js
```diff
@@ -0,0 +1,166 @@
...
+ * @param {string} op - operator name
+ * @returns {boolean} true when recognised
+ */
+function queryExists(op) {
+    var candidates = [{ f: {} }, {}];
+    candidates[0].f[op] = 1;
+    candidates[1][op] = 1;
+    for (var i = 0; i < candidates.length; i++) {
+        try {
+            probeDb.runCommand({ find: "probe", filter: candidates[i], limit: 1, maxTimeMS: 400 });
```
ar2rsawseen: [P2] Inspect the command result instead of assuming `runCommand()` throws. In mongosh, a failed `find` command can return `{ok: 0, code, errmsg}`; this branch then unconditionally returns `true`, causing an unknown query operator to be classified as recognized. That makes `existsAnywhere("$notARealOperatorAtAll")` true and invalidates the verifier controls. Check `res.ok` and its error/code fields here, as `attempt()` already does for aggregation probes.

plugins/dbviewer/api/parts/verify_operators.js
```diff
@@ -0,0 +1,166 @@
...
+});
+print("excluded operators checked: " + EXCLUDED.length);
+print("excluded but NOT recognised (typo, or gone from this version): " + (goneExclusions.length ? goneExclusions.join(" ") : "(none)"));
+
+var wronglyAllowed = EXCLUDED.filter(function(op) {
+    return (guard.ALLOWED_OPERATORS_GLOBAL_ADMIN || {})[op] === true;
+});
+print("excluded operators present in the allow-list (must be none): " + (wronglyAllowed.length ? wronglyAllowed.join(" ") : "(none)"));
+
+probeDb.dropDatabase();
```
ar2rsawseen, line 202: [P2] Avoid dropping a fixed-name database. The documented command accepts an arbitrary connection string, and if that cluster already has a `dbviewer_operator_probe` database this deletes all of its collections, not only the probe document created above. Use a unique database name for each run or remove only the collection/document owned by this script.

Cookiezaurs (approved):

plugins/dbviewer/api/parts/aggregation_guard.js
ar2rsawseen: Confirmed and fixed in fff106f86ba. This one was a real break of valid queries, and it came from a claim in this module's own header that is false.

The header said "MongoDB will not let a `$`-prefixed key be anything other than an operator", and the scan was built on it. The geospatial and text query operators spell their **options** with a `$` too. `{$geoWithin: {$centerSphere: [...]}}` is not an edge case — `plugins/geo/api/geo.js`, `exp.conds`, builds exactly that shape — and the guard returned a 400 before MongoDB saw it. Same for `$box`, `$center`, `$polygon`, `$geometry`, `$maxDistance`, `$minDistance`, and the `$text` options `$search`, `$language`, `$caseSensitive`, `$diacriticSensitive`.

Fixed the way you describe, operator-aware, and deliberately **not** by widening the flat list:

```js
const OPERATOR_OPTION_KEYS = {
    "$geoWithin": ["$box", "$center", "$centerSphere", "$geometry", "$polygon"],
    "$geoIntersects": ["$geometry"],
    "$near": ["$geometry", "$maxDistance", "$minDistance"],
    "$nearSphere": ["$geometry", "$maxDistance", "$minDistance"],
    "$text": ["$search", "$language", "$caseSensitive", "$diacriticSensitive"]
};
```

An option key is accepted **only directly inside the operator that owns it**. Adding these to the allow-list would have accepted `{$match: {loc: {$geometry: ...}}}`, which is not valid MongoDB, so that stays rejected — and the context does not leak downwards either: `$geometry` declares no options of its own, so a `$centerSphere` nested inside it is still refused. This is a declared table read one level deep, not structure inferred from contents, so the module keeps both properties it was built for.

**`$literal` is no longer descended into.** Its value is data returned unevaluated, so reading keys inside it contradicted this file's own "keys only, never values" rule — `{$literal: {$lookup: 1}}` is the object `{$lookup: 1}`, not a join. The protected-join scan is left blind on purpose: a false positive there costs a rejected query, which is the safe direction.

Seven test cases: the real geo predicate from `geo.js`, the other shape and distance options, the `$text` options, `$literal` carrying arbitrary keys, an option key used outside its operator (rejected), an unknown key inside an operator that has options (rejected), and the context not leaking into a nested document. Reverting the option handling turns four of them red, so they fail for the right reason. 66 passing.

plugins/dbviewer/api/parts/aggregation_guard.js
ar2rsawseen: Fixed in fff106f86ba, with the limit of what I could verify stated plainly.

Added: `$bitAnd`, `$bitNot`, `$bitOr`, `$bitXor`, `$toHashedIndexKey` — the ones you named — plus `$sampleRate`, a `$match` operator the same pass turned up missing.

You are right that the committed verifier cannot find omissions; it only tests names already in the array, which the header already says. So this is not the audit you are asking for, and I have not claimed it is: a genuine one means running `verify_operators.js` against a live MongoDB 8 of the version this release supports, and that needs a cluster this branch does not have. What I have done is fix the two faults that would have made such a run misleading — see the other two threads — so the audit is worth running when someone points it at one.

Note the shape of the cost here, since it decides how urgent this is. An omission rejects a valid query, which is a support ticket; the reverse, a missing entry in a deny-list, is an exfiltration. That asymmetry is why the module fails closed, and it is also why I would rather add names on evidence than pad the list from memory.

Regression cases added for the bitwise family and for `$toHashedIndexKey`/`$sampleRate`, plus one asserting a name no MongoDB has (`$bitShiftLeft`) is still refused, so the additions cannot quietly become a wildcard.

plugins/dbviewer/api/parts/verify_operators.js
ar2rsawseen: Confirmed and fixed in fff106f86ba. You are right, and the consequence is worse than one wrong answer: it made the whole query half of the script meaningless.

`db.runCommand()` reports a failed command as a document — `{ok: 0, code, errmsg}` — and only raises for some failures, so "it did not throw" said nothing. The first candidate returned unconditionally `true`, every query operator came back recognised, and the deliberately fake control `$notARealOperatorAtAll` was reported real. A control failing is supposed to mean the run's output is worthless, which is exactly what was happening.

Now read the same way `attempt()` already read aggregation probes:

```js
function attemptFind(filter) {
    try {
        var res = probeDb.runCommand({ find: "probe", filter: filter, limit: 1, maxTimeMS: 400 });
        return { ok: res.ok === 1, errmsg: res.errmsg };
    }
    catch (e) {
        return { ok: false, errmsg: e.message };
    }
}
```

and `queryExists()` treats a command that succeeded as recognised, a command that failed for some *other* reason as recognised too — a real operator given a wrong argument — and only "unknown operator" / "unknown top level operator" as not recognised.

One deliberate difference from the aggregation probes: those key off error codes because the wording is unstable, but an unknown query operator is `BadValue`, which plenty of ordinary filter errors also are. The code alone cannot separate "no such operator" from "wrong argument to a real one", so the message test stays here, now applied to the command result as well as to a thrown error.

plugins/dbviewer/api/parts/verify_operators.js
ar2rsawseen: Confirmed and fixed in fff106f86ba. The usage line at the top of the file takes any connection string, so a fixed database name plus an unconditional `dropDatabase()` was a data-loss trap for anyone who happened to have that name in use.

Three changes rather than one, so it fails safe rather than relying on the name being unusual:

```js
var PROBE_DB_NAME = "dbviewer_operator_probe_" + new Date().getTime() + "_" + Math.floor(Math.random() * 1000000);
var probeDb = db.getSiblingDB(PROBE_DB_NAME);
if (probeDb.getCollectionNames().length) {
    print("Refusing to run: " + PROBE_DB_NAME + " already exists and is not empty.");
    quit(1);
}
```

- a unique name per run, so a collision with anything pre-existing is not a thing that happens;
- the script refuses to start if that name somehow exists and is not empty, rather than writing into it;
- and only then is the drop at the end unconditional, with a comment saying why it is allowed to be.

I kept `dropDatabase()` rather than narrowing to the probe collection: the probes create nothing else today, but a future probe that does would otherwise leave the database behind on every run, and a name this run generated and verified empty is safe to remove whole.

PR #7870, opened 2026-07-30
security(configs): return only declared configuration to non global admins, and mask credentials

ar2rsawseen:
## What

Configuration was returned wholesale to anyone who could read it at all, and some of it holds credentials rather than settings: an outbound proxy password, a reCAPTCHA secret, the key that signs report subscribe/unsubscribe tokens.

Two mechanisms, applied in this order.

### 1. An allow-list, which is the control

`plugins.setReadableConfigs(namespace, {key: true})` declares what a caller who is not a global admin may read. Everything else is withheld, and a namespace with nothing declared is dropped entirely.

The direction matters. Marking credentials as secret and returning the rest is a deny-list: every new setting is exposed until somebody remembers to annotate it, so one forgotten annotation is a leaked credential with no signal. Declaring what is needed fails the other way, and the cost of forgetting is a missing input in the UI, which someone notices and files.

Declared, from surveying every reader across countly-server and countly-enterprise-plugins:

| Namespace | Keys | Needed by |
|---|---|---|
| `api` | the 13 in `showInAppManagment`, plus `domain` | App Management inputs, and crash symbolication's return URL |
| `security` | the password policy | client-side password validation |
| `feedback` | `main_color`, `font_color`, `feedback_logo` | Ratings and Surveys widget previews |
| `concurrent_users` | `alert_interval` | the alerts drawer, declared by the consumer since the namespace belongs to an enterprise plugin |

### 2. Marking credentials, which is the backstop

`plugins.setSecretConfigs(namespace, {key: true})` masks a value as `********`. It still earns its place with the allow-list in front of it:

- it covers the values serialized into the dashboard page, a separate path from `/o/configs`
- if a credential is ever declared readable by mistake, its value is withheld anyway
- it documents intent where the config is defined

Marked: `security.proxy_username`, `security.proxy_password`, `push.proxyuser`, `push.proxypass`, `reports.secretKey`, `recaptcha.secret_key`.

## Values the browser needs are deliberately not marked

`recaptcha.site_key` is public by design, and `tracking.self_tracking_app_key` is what the dashboard uses to send its own telemetry. **A value the browser needs cannot be secret**: hiding it from an API response while it stays in page source buys nothing. If a real credential is ever needed browser-side, the fix is to proxy the call server-side.

Before dropping anything from the page I enumerated every `countlyGlobal.security.*` and `countlyGlobal.tracking.*` reference in all three repositories. The dashboard uses exactly five keys from `security`, all password policy.

## Three things that could have gone wrong

- **`getConfig()` is untouched.** Server code keeps reading real values; masking there would have push and reports authenticating with the literal string `********`.
- **Both filters copy.** `getConfig()` returns a fresh object per namespace and neither filter writes into it, so a read cannot alter live configuration. Asserted by test.
- **An unset credential stays empty** rather than showing a mask, so "not configured" remains distinguishable from "configured but hidden".

## Writes need no guard

`/i/configs` is `validateGlobalAdmin`, so the only callers who can write configuration are the only callers who see real values, and a mask cannot be written back over a value. Worth remembering if that gate is ever loosened.

## The failure mode this is built around

The dashboard degrades silently when configuration is missing: the call site uses `.always()`, the model initialises to `{}`, and the input loop is guarded by `if (pluginsData[key])`. So an undeclared value shows up as a setting quietly absent from a screen, with no error anywhere.

Two things address that:

- a withheld namespace is logged at debug, so there is a trail
- the App Management key list is **read out of the frontend by the test** rather than duplicated in it, so adding a key to `showInAppManagment` without declaring it readable fails CI instead of removing a setting from the screen

## Tests

- `test/unit-tests/plugins.pluginManager.secret-configs.js`, 23 cases: both filters, non-mutation, unknown namespaces, prototype keys, `getConfig()` still returning real values, a newly added credential nobody marked being withheld, and a credential wrongly declared readable still being masked.
- `plugins/plugins/tests.js`, 5 cases through the real endpoints with a real non-global app admin.

Full unit suite: 208 passing. The 2 failures are `Countly Request`, which makes real HTTP calls and fails without network on a clean checkout too.

## Enterprise follow-up

Enterprise plugins need declarations for namespaces they own and read: `crashes.custom_domain`, `journey_engine.approver_enabled`, and whatever `data-manager` uses. Until then those fall back to their existing defaults, which for `journey_engine` is a value already injected into the page, and for symbolication is the browser origin.

## Separate issue, not fixed here

`/i/configs` records the full configuration as the `before` payload of its systemlogs entry, persisting these values indefinitely. Global-admin only to read, so not a boundary crossing, but worth masking there too.


frontend/express/app.js
```diff
@@ -1033,9 +1033,15 @@ Promise.all([plugins.dbConnection(countlyConfig), plugins.dbConnection("countly_
...
-                    security: plugins.getConfig("security"),
-                    tracking: plugins.getConfig("tracking"),
+                    //these namespaces are serialized into the page, so every logged in
+                    //user can read them in the source regardless of their permissions.
+                    //Secret values are dropped rather than masked: the page has no use
+                    //for a placeholder. Anything the browser genuinely needs must not
+                    //be marked secret, since masking what is already in page source
+                    //would achieve nothing.
+                    config: plugins.omitSecretConfigs("frontend", req.config),
+                    security: plugins.omitSecretConfigs("security", plugins.getConfig("security")),
```
ar2rsawseen, line 1049: [P1] Register the secret metadata in the dashboard process too. `secretConfigs` is process-local, while `setSecretConfigs("security", ...)` currently runs only from `api/api.js`; the dashboard entry point defines/loads the `security` namespace independently and never executes that registration. Consequently this call omits nothing and `countlyGlobal.security` still serializes stored `proxy_username` and `proxy_password` values for every logged-in user. Move the shared config/secret initialization into a module required by both entry points, or repeat the secret registration in `frontend/express/app.js`, and add a rendered-dashboard regression test.

Cookiezaurs (approved):

frontend/express/app.js
ar2rsawseen: Confirmed and fixed in 3a5a7bb1401. You were right, and the consequence is exactly as you describe: on the dashboard side this PR was doing nothing at all.

`secretConfigs` is state on the pluginManager instance, and the API and the dashboard are separate processes with an instance each. `setSecretConfigs("security", ...)` ran only from `api/api.js`, so in the dashboard process nothing was registered, `omitSecretConfigs("security", plugins.getConfig("security"))` had nothing to omit, and `countlyGlobal.security` kept carrying the stored proxy credentials into the page source of every logged in user.

Worth adding one detail, because it decides where the fix belongs: the dashboard's own `setConfigs("security", ...)` does not even declare `proxy_username`/`proxy_password` — its default list stops at `dashboard_rate_limit_requests`, while the API's continues through the proxy and API rate-limit keys. The credentials reach the page purely as stored values merged in by `getConfig`. So the two processes have *already* drifted on this namespace, which is the argument for your first option rather than your second.

**The shared module.** `api/utils/configMetadata.js` now holds the three core declarations — `api` readable, `security` readable, `security` secret — and exports `register(plugins)`. `api/api.js` and `frontend/express/app.js` each call it. Repeating the registration in the dashboard would have fixed today's bug and left the next one in place.

**Left alone deliberately:** the plugin declarations in `plugins/*/api/api.js`. They are equally process local, but the dashboard serializes only `frontend`, `security` and `tracking`, and no plugin declares secrets in those. `recaptcha` — the one plugin whose secret the dashboard needs — already registers from `frontend/app.js`. Worth knowing rather than worth moving.

**Tests**, in the existing unit suite:

- the proxy keys are marked secret;
- `omitSecretConfigs` drops them from what the dashboard puts in the page, while keeping the password policy the browser needs;
- the password policy is readable and nothing else in the namespace is;
- and the one that guards what you reported: both entry points are asserted to call `register`. I checked it fails for the right reason — commenting out the dashboard call turns that case red and the others stay green.

27 passing.

On the rendered-dashboard regression test you asked for: I did not add one. Reaching `countlyGlobal` in page source means driving the frontend process through Cypress, which is a long way round for an assertion the entry-point guard makes directly and deterministically. Say the word if you want the browser-level one as well.

**Checked across the other repositories in this change set:**

- **countly-platform** (Countly/countly-platform#929) — not affected. Its declarations already live in `api/init_configs.ts`, and `api/api.js`, `frontend/express/app.js`, `api/ingestor.ts` and `api/aggregator.ts` all load it. That is the shape this fix moves countly-server towards. I added the equivalent entry-point guard there so it stays true.
- **countly-enterprise-plugins** (Countly/countly-enterprise-plugins#3335, Countly/countly-enterprise-plugins#3336) — not affected. The only secret it declares is `crashes.symbolication_key`, in a namespace the dashboard never serializes and read only by API-side code and the symbolication job. Its readable declarations are consumed by `/o/configs` in the API process, where they are registered.

PR #7871, opened 2026-07-30
security(configs): mark credential configs secret and withhold them on read (24.05)

ar2rsawseen:
Backport of #7870 to `release.24.05`. Clean cherry-pick, no adaptation needed.

Adds `plugins.setSecretConfigs(namespace, {key: true})` so configuration that holds a credential is declared where it is defined. Declared values are masked in `/o/configs` for anyone who is not a global admin, and left out of the namespaces serialized into the dashboard page.

Marked: `security.proxy_username`, `security.proxy_password`, `push.proxyuser`, `push.proxypass`, `reports.secretKey`, `recaptcha.secret_key`.

Not marked, because the browser needs them: `recaptcha.site_key` and `tracking.self_tracking_app_key`.

`getConfig()` is untouched, so server code keeps reading real values. Full reasoning in #7870.

## Verified on this branch

The call sites differ in line numbers but not in shape, so I checked each landed where intended rather than trusting a clean apply: the mask in `/o/configs`, the three narrowed namespaces in `frontend/express/app.js`, and all four declarations.

19 tests pass here: 15 unit plus 4 API.

### DIFF
--- a/frontend/express/app.js
+++ b/frontend/express/app.js
@@ -144,6 +144,12 @@ plugins.setConfigs("security", {
     dashboard_rate_limit_requests: 500
 });
 
+//the same declarations the API process makes. secretConfigs is process local, so
+//without this omitSecretConfigs() below has nothing registered to omit and the
+//stored proxy credentials go into the page source of every logged in user.
+require('../../api/utils/configMetadata.js').register(plugins);
+
+
 process.on('uncaughtException', (err) => {
     console.log('Caught exception: %j', err, err.stack);
     if (log && log.e) {
--- a/frontend/express/app.js
+++ b/frontend/express/app.js
@@ -992,9 +992,15 @@ Promise.all([plugins.dbConnection(countlyConfig), plugins.dbConnection("countly_
                     csrf_token: req.csrfToken(),
                     auth_token: req.session.auth_token,
                     member: member,
-                    config: req.config,
-                    security: plugins.getConfig("security"),
-                    tracking: plugins.getConfig("tracking"),
+                    //these namespaces are serialized into the page, so every logged in
+                    //user can read them in the source regardless of their permissions.
+                    //Secret values are dropped rather than masked: the page has no use
+                    //for a placeholder. Anything the browser genuinely needs must not
+                    //be marked secret, since masking what is already in page source
+                    //would achieve nothing.
+                    config: plugins.omitSecretConfigs("frontend", req.config),
+                    security: plugins.omitSecretConfigs("security", plugins.getConfig("security")),
+                    tracking: plugins.omitSecretConfigs("tracking", plugins.getConfig("tracking")),
                     plugins: plugins.getPlugins(),
                     pluginsFull: plugins.getPlugins(true),
                     path: countlyConfig.path || "",
--- a/api/api.js
+++ b/api/api.js
@@ -90,6 +90,43 @@ plugins.connectToAllDatabases().then(function() {
         trim_trailing_ending_spaces: false
     });
 
+    //What a non global admin may read from the api namespace. The first group is what
+    //App Management renders: the dashboard infers each input's widget from the type of
+    //the value here, so a key missing from this list means that setting disappears from
+    //the screen. Keep it in step with showInAppManagment in the plugins frontend.
+    //
+    //domain is separate: crash symbolication builds its return URL from it, and falls
+    //back to the browser's origin when it is absent, which is wrong behind a custom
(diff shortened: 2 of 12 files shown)


frontend/express/app.js
```diff
@@ -992,9 +992,15 @@ Promise.all([plugins.dbConnection(countlyConfig), plugins.dbConnection("countly_
...
-                    security: plugins.getConfig("security"),
-                    tracking: plugins.getConfig("tracking"),
+                    //these namespaces are serialized into the page, so every logged in
+                    //user can read them in the source regardless of their permissions.
+                    //Secret values are dropped rather than masked: the page has no use
+                    //for a placeholder. Anything the browser genuinely needs must not
+                    //be marked secret, since masking what is already in page source
+                    //would achieve nothing.
+                    config: plugins.omitSecretConfigs("frontend", req.config),
+                    security: plugins.omitSecretConfigs("security", plugins.getConfig("security")),
```
ar2rsawseen, line 1008: [P1] Register the secret metadata in the dashboard process too. `secretConfigs` is process-local, while `setSecretConfigs("security", ...)` currently runs only from `api/api.js`; the dashboard entry point defines/loads the `security` namespace independently and never executes that registration. Consequently this call omits nothing and `countlyGlobal.security` still serializes stored `proxy_username` and `proxy_password` values for every logged-in user. Move the shared config/secret initialization into a module required by both entry points, or repeat the secret registration in `frontend/express/app.js`, and add a rendered-dashboard regression test.

Cookiezaurs (approved):

frontend/express/app.js
ar2rsawseen: Confirmed and fixed in e09e46c386c. You were right, and the consequence is exactly as you describe: on the dashboard side this PR was doing nothing at all.

`secretConfigs` is state on the pluginManager instance, and the API and the dashboard are separate processes with an instance each. `setSecretConfigs("security", ...)` ran only from `api/api.js`, so in the dashboard process nothing was registered, `omitSecretConfigs("security", plugins.getConfig("security"))` had nothing to omit, and `countlyGlobal.security` kept carrying the stored proxy credentials into the page source of every logged in user.

Worth adding one detail, because it decides where the fix belongs: the dashboard's own `setConfigs("security", ...)` does not even declare `proxy_username`/`proxy_password` — its default list stops at `dashboard_rate_limit_requests`, while the API's continues through the proxy and API rate-limit keys. The credentials reach the page purely as stored values merged in by `getConfig`. So the two processes have *already* drifted on this namespace, which is the argument for your first option rather than your second.

**The shared module.** `api/utils/configMetadata.js` now holds the three core declarations — `api` readable, `security` readable, `security` secret — and exports `register(plugins)`. `api/api.js` and `frontend/express/app.js` each call it. Repeating the registration in the dashboard would have fixed today's bug and left the next one in place.

**Left alone deliberately:** the plugin declarations in `plugins/*/api/api.js`. They are equally process local, but the dashboard serializes only `frontend`, `security` and `tracking`, and no plugin declares secrets in those. `recaptcha` — the one plugin whose secret the dashboard needs — already registers from `frontend/app.js`. Worth knowing rather than worth moving.

**Tests**, in the existing unit suite:

- the proxy keys are marked secret;
- `omitSecretConfigs` drops them from what the dashboard puts in the page, while keeping the password policy the browser needs;
- the password policy is readable and nothing else in the namespace is;
- and the one that guards what you reported: both entry points are asserted to call `register`. I checked it fails for the right reason — commenting out the dashboard call turns that case red and the others stay green.

27 passing.

On the rendered-dashboard regression test you asked for: I did not add one. Reaching `countlyGlobal` in page source means driving the frontend process through Cypress, which is a long way round for an assertion the entry-point guard makes directly and deterministically. Say the word if you want the browser-level one as well.

**Checked across the other repositories in this change set:**

- **countly-platform** (Countly/countly-platform#929) — not affected. Its declarations already live in `api/init_configs.ts`, and `api/api.js`, `frontend/express/app.js`, `api/ingestor.ts` and `api/aggregator.ts` all load it. That is the shape this fix moves countly-server towards. I added the equivalent entry-point guard there so it stays true.
- **countly-enterprise-plugins** (Countly/countly-enterprise-plugins#3335, Countly/countly-enterprise-plugins#3336) — not affected. The only secret it declares is `crashes.symbolication_key`, in a namespace the dashboard never serializes and read only by API-side code and the symbolication job. Its readable declarations are consumed by `/o/configs` in the API process, where they are registered.

PR #7872, opened 2026-07-31
security(alerts): authorize the apps an alert targets, not just who owns it

ar2rsawseen:
## What

The alert endpoints authorize the caller against `params.qstring.app_id`, and the create path checks the *submitted* `selectedApps`. Neither says anything about the apps the **stored** alert points at, and an update may omit `selectedApps` to keep whatever is stored.

The by-id mutation paths filtered only on `_id` and `createdBy`. That is ownership, not authorization: a member who created an alert for one app and later lost access to it, while keeping alerts rights elsewhere, could still edit and re-enable that alert by sending the request with an app they do still hold.

`/o/alert/list` was already fixed for this and hides such alerts, so the write side was the remaining half.

## Three changes

**Update by id** loads the alert and requires update rights on every app it targets, on top of the existing check on any newly submitted `selectedApps`.

**The status endpoint** requires the same before switching an alert **on**. Switching one **off** stays allowed: it only reduces what the alert does, and refusing would leave somebody who lost access unable to stop mail they no longer want, which is worse than the problem being fixed.

**The scheduled job** checks the owner still has read access before sending. Nothing re-checked anything once an alert was enabled, so one created before access was revoked kept emailing that app's metrics indefinitely, with no request from anybody. Fixing only the endpoints would have left that in place, and it is the more likely case in practice.

`delete` deliberately stays owner-only. It cannot reach another app's data, and the owner needs to be able to clean up an alert they can no longer manage.

## The legacy trap this had to avoid

Members created before the `permission` object reach apps through `user_of`, and `hasUpdateRight` / `hasReadRight` then fall through to `admin_of` alone. A strict reading would judge them unauthorized, which would have **locked them out of their own alerts and stopped delivery**. `/o/alert/list` already makes that allowance, and now all three paths agree, otherwise an alert could be listed but not editable, or editable but not delivered.

The shared decision lives in `parts/app-authorization.js` because the request path and the scheduled job both need it. Two copies of that allowance would drift.

## One deliberate fail-open

An alert with no resolvable `createdBy` is reported at warn level and still sent. Those predate the field, cannot be attributed, and refusing would silently stop alerts that have worked for years. That is a worse outcome than the exposure closed here, and it is visible in the log rather than silent.

## Tests, both directions

- `test/unit-tests/plugins.alerts.app-authorization.js`, 19 cases: modern and legacy members, `admin_of` vs `user_of`, global admin, multi-app targets requiring all apps, empty target lists, missing members, and that read access does not confer update.
- `plugins/alerts/tests.js`, API tests through the real endpoints with a real member whose access is revoked mid-test.

The exploit cases: the update with `selectedApps` omitted is refused, enabling is refused, the alert is verified unchanged afterwards, and it stays hidden from the member's list.

The happy paths, which matter just as much: the member can still create, update, enable and disable their own alerts on the app they hold, can still switch off **and delete** the alert for the revoked app, and a global admin is unaffected.

Full unit suite: 204 passing. The 2 failures are `Countly Request`, which makes real HTTP calls and fails without network on a clean checkout too.

## Note on scope

The status pre-check is scoped to the caller's own alerts, so an id belonging to somebody else keeps behaving as it did, a silent no-op, rather than returning a 403 that would confirm an alert with those apps exists.


plugins/alerts/api/api.js
```diff
@@ -248,22 +250,41 @@ function getScheduleTextExpression(period, offset) {
...
-                        function(err, result) {
-                            if (!err) {
-                                if (result && result.value) {
-                                    plugins.dispatch("/updateAlert", { method: "alertTrigger", alert: result.value });
+                    //The guard above only saw the submitted selectedApps, and an update
+                    //may leave that out to keep whatever is stored. So load the alert
+                    //and authorize the apps it currently targets before changing it,
+                    //otherwise an alert for an app the caller has since lost access to
+                    //stays editable through any app they do still hold.
+                    return common.db.collection("alerts").findOne({_id: common.db.ObjectID(id)}, function(findErr, existingAlert) {
```
ar2rsawseen: [P1] Scope this pre-load with `query`, not `_id` alone. For a non-global caller, the subsequent mutation intentionally includes `createdBy`, but this lookup can fetch somebody else’s alert and return a distinct 403 based on its `selectedApps`. That creates an alert-existence/target metadata oracle and changes the prior silent no-op for non-owners. Loading with the same ownership selector preserves that behavior while still authorizing the stored apps for alerts the caller can mutate.

plugins/alerts/api/api.js
```diff
@@ -248,22 +250,41 @@ function getScheduleTextExpression(period, offset) {
...
                     }
-                    return common.db.collection("alerts").findAndModify(
-                        query,
-                        {},
-                        {$set: alertConfig},
-                        function(err, result) {
-                            if (!err) {
-                                if (result && result.value) {
-                                    plugins.dispatch("/updateAlert", { method: "alertTrigger", alert: result.value });
+                    //The guard above only saw the submitted selectedApps, and an update
```
ar2rsawseen, line 266: [P1] Preserve the stored creator on updates. A global admin may edit another member’s alert, and the existing assignment immediately above transfers `createdBy` to the admin; because ownership, list visibility, and the new send-time authorization all depend on `createdBy`, the original owner loses the alert and delivery is evaluated against the wrong member. Strip any request-supplied `createdBy` and set it only on insert.

Cookiezaurs (approved):

plugins/alerts/api/api.js
ar2rsawseen: Confirmed and fixed in e2cc8478ad3.

You are right on both halves — the leak and the behaviour change. The check loads the alert to authorize its stored `selectedApps`, but by `_id` alone it loaded anybody's, so a non-owner received a 403 that both says the alert exists and describes what it targets. Before this PR the same request was a silent no-op, because `findAndModify` carries `createdBy` for a non-global caller and simply matched nothing.

It now loads with `query` — the same selector the write uses, ownership included. For an alert the caller can mutate nothing changes: the stored apps are still authorized before the update. For one they cannot, the lookup finds nothing, the check is skipped, and the write stays the no-op it always was. No oracle, no behaviour change.

plugins/alerts/api/api.js
ar2rsawseen: Confirmed and fixed in e2cc8478ad3.

`createdBy` was assigned to the caller before the update was built, so a global admin editing another member's alert took it over. That matters more here than it would have before this PR: ownership is what `/o/alert/list` filters by *and* what the send-time authorization added in this change evaluates, so the original owner lost the alert from their list and delivery started being judged against the wrong member's app access.

It is also not the request's to set, as you say. Now:

```js
delete alertConfig.createdBy;
```

on the update path, so `$set` never touches the stored value, and it is still assigned on insert where it means what it says. `query.createdBy` is built from `params.member._id` directly and is unaffected.

PR #7877, opened 2026-08-03
security(exports): neutralize spreadsheet formulas in the streamed CSV export

ar2rsawseen:
Reported via the bug bounty programme. Reproduced and confirmed on latest master.

Analytics data is written by whoever holds the app key, which ships inside published SDKs, so cell contents are attacker-influenced. A cell that reaches a spreadsheet client beginning with `=`, `+`, `-`, `@`, tab or carriage return is parsed as a **formula** rather than text. In clients that permit external data functions (`WEBSERVICE`, `IMPORTDATA`, `IMPORTXML`) the exported file can then be made to call out to an address of the attacker's choosing, carrying adjacent cells of the same export with it.

`preventCSVInjection` already existed. The streamed path never called it. It only called `processCSVvalue`, which quote-wraps the field, and **quoting is not neutralization**: those quotes are the CSV text qualifier and the client strips them while parsing.

## Three gaps, all reachable via `/o/export/db` and `/o/export/requestQuery`

1. **Values.** `getValues` only routes a row through `flattenObject`, where the neutralizer lived, when **no projection** is given (`collectProp: listAtEnd`). Supply a projection and the values bypass it completely. That is the case the report demonstrated, and it is the one an operator hits from any Vue datatable export, since those send a projection.
2. **Headers.** The header row is written from two separate places and neither neutralized, so an attacker-supplied segment or property key could become a formula.
3. **Two prefixes were missing.** `preventCSVInjection` covered `= + - @` but not a leading tab or carriage return, which the client consumes while parsing and so exposes whatever follows. This gap affected the **non-streaming** path too, which did call the neutralizer.

## The fix

Neutralization now happens inside `processCSVvalue`, which is internal to this file and is the single point every streamed CSV cell passes through, headers included. Putting it there rather than at the three call sites means a future call site cannot forget it.

Because `flattenObject` still neutralizes as well, a value on the no-projection path now passes through twice, so the neutralizer is **idempotent**: a value already carrying the marker is returned untouched. Verified there is exactly one marker on both paths.

Two deliberate choices:

- **Kept the backtick marker** rather than switching to the apostrophe the report suggested. Both work, since neither introduces a formula, but the backtick is what this code already emits and changing it would alter the content of every existing export.
- **Numbers are untouched.** Only strings are marked, so exported figures including negative ones are unaffected. A string that looks like `-1` does get marked, which is pre-existing behaviour on the non-streaming path and is now consistent across both.

XLSX is not affected and is unchanged: `xlsx-write-stream` emits every cell as `t="inlineStr"` with `<is><t>`, never an `<f>` formula node. I verified that in the installed dependency rather than taking it on trust. The reporter had already scoped it out correctly.

## Verification

Driving `exports.stream` directly with the reported payload, using a naive unquoting parser to see what a client would see:

| | before | after |
|---|---|---|
| cells parsed as formulas | **5** (header, `=`, `+`, tab, CR) | **0** |

9 unit tests in `test/unit-tests/api.exports.csv-formula-neutralization.js`, covering both directions: every prefix is defanged on both streamed paths and in headers, and ordinary values (`plain`, `user@example.com`, `a=b`) come through with no marker added, since otherwise every export quietly grows stray characters.

**4 of the 9 fail without the fix and all 9 pass with it**, and the four failures map one-to-one onto the three gaps above.

Full unit suite: **185 passing before, 194 after** (+9, the new tests), with the same 2 pre-existing failures either way (`Countly Request`, network-dependent). Baseline established by reverting the change and re-running, not assumed. eslint clean.

## Propagation

Going to `release.24.05` and countly-platform `main`; both carry the same gap. Platform's copy is TypeScript but structurally identical.


api/parts/data/exports.js
```diff
@@ -102,16 +102,31 @@ function flattenObject(ob, fields) {
     return toReturn;
 }
 
+//A leading one of these makes a spreadsheet client read the cell as a formula. The
+//first four are the formula introducers; a leading tab or carriage return is consumed
+//by the client while it parses the file, exposing whatever follows it, so they belong
+//here too.
+var CSV_FORMULA_PREFIXES = ["@", "=", "+", "-", "\t", "\r"];
```
ar2rsawseen, line 109: [P2] Keep CSV neutralization out of the shared flattening path

`flattenObject()` feeds CSV, XLS/XLSX, and JSON export paths, but this now applies the CSV-only leading-apostrophe marker to every format (and expands the affected values to tab/CR-prefixed strings). As a result, legitimate values in an XLS/XLSX export are stored with an extra apostrophe even though those writers do not need CSV formula escaping. Please return raw values from the shared flattening step and apply `preventCSVInjection()` only at CSV serialization boundaries.

ar2rsawseen: Confirmed and fixed in cbd0c0fb942. You are right, and it reached further than the flattening path: this change also neutralized explicitly in the xlsx *streaming* branch, which is not a CSV writer at all.

`flattenObject()` returns raw values now. Neutralization happens where CSV text is produced, in two places:

- **`processCSVvalue`**, which this PR already added and which covers every cell the streamed writer emits, headers included. That is what makes removing it from the flattening path safe rather than a regression.
- **`convertData`'s csv branch**, which does *not* go through `processCSVvalue` — it hands rows straight to json2csv. Without adding it there, taking it out of `flattenObject` would have quietly removed the protection from the buffered csv export. Field names go through it too: json2csv writes them as the header row, and a segmentation key is attacker supplied.

The xlsx branch writes `paramList` and `values` unchanged again, and json comes back untouched — there is a case for that now.

**One more thing, found while running it:** this PR's own suite was failing. The assertions were written against a backtick marker while the implementation prepends an apostrophe, so the single case that actually checked the marker failed, and four `indexOf("\`").should.equal(-1)` assertions passed for any output whatsoever, neutralized or not. They read the marker from one place now.

10 passing, from 8 passing and 1 failing before.

Cookiezaurs (approved):

PR #7878, opened 2026-08-03
security(exports): neutralize spreadsheet formulas in the streamed CSV export (24.05)

ar2rsawseen:
Backport of #7877 to `release.24.05`, which carries the identical gap. See that PR for the full analysis.

The streamed CSV path calls `processCSVvalue` (which quote-wraps) but never `preventCSVInjection`, and CSV quoting is not neutralization: the quotes are the text qualifier the spreadsheet client strips while parsing.

Same three parts as master:

1. **Values** bypass the neutralizer whenever a projection is given, since only the no-projection path routes rows through `flattenObject`.
2. **Header cells** were never neutralized at all.
3. `preventCSVInjection` covered `= + - @` but not a leading tab or carriage return, which affected the non-streaming path on this branch too.

Neutralization moved into `processCSVvalue`, the single point every streamed cell passes through, and made idempotent so the no-projection path marks a value exactly once.

The patch applied to this branch cleanly, and the 9 unit tests pass here unchanged. eslint clean.

### DIFF
--- a/api/parts/data/exports.js
+++ b/api/parts/data/exports.js
@@ -447,19 +447,23 @@ exports.stream = function(params, stream, options) {
         });
         xc.pipe(params.res);
         if (listAtEnd === false) {
-            xc.write(paramList);
+            xc.write(paramList.map(preventCSVInjection));
         }
         stream.stream(options.streamOptions).on('data', function(doc) {
             var values = [];
             var valuesMap = {};
             getValues(values, valuesMap, paramList, doc, {mapper: mapper, collectProp: listAtEnd});
-            xc.write(values);
+            //getValues only neutralizes by way of flattenObject, which its collectProp=false
+            //branch skips, so a projected export would otherwise write formulas into the sheet
+            //verbatim. Headers go through the same guard, since a column name is attacker
+            //controlled too when it comes from event segmentation.
+            xc.write(values.map(preventCSVInjection));
         });
 
         stream.once('close', function() {
             setTimeout(function() {
                 if (listAtEnd) {
-                    xc.write(paramList);
+                    xc.write(paramList.map(preventCSVInjection));
                 }
                 xc.end();
             }, 100);
--- a/api/parts/data/exports.js
+++ b/api/parts/data/exports.js
@@ -102,16 +102,31 @@ function flattenObject(ob, fields) {
     return toReturn;
 }
 
+//A leading one of these makes a spreadsheet client read the cell as a formula. The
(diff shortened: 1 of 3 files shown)


api/parts/data/exports.js
```diff
@@ -102,16 +102,31 @@ function flattenObject(ob, fields) {
     return toReturn;
 }
 
+//A leading one of these makes a spreadsheet client read the cell as a formula. The
+//first four are the formula introducers; a leading tab or carriage return is consumed
+//by the client while it parses the file, exposing whatever follows it, so they belong
+//here too.
+var CSV_FORMULA_PREFIXES = ["@", "=", "+", "-", "\t", "\r"];
```
ar2rsawseen, line 109: [P2] Keep CSV neutralization out of the shared flattening path

`flattenObject()` feeds CSV, XLS/XLSX, and JSON export paths, but this now applies the CSV-only leading-apostrophe marker to every format (and expands the affected values to tab/CR-prefixed strings). As a result, legitimate values in an XLS/XLSX export are stored with an extra apostrophe even though those writers do not need CSV formula escaping. Please return raw values from the shared flattening step and apply `preventCSVInjection()` only at CSV serialization boundaries.

ar2rsawseen: Confirmed and fixed in 918f045a9c1. You are right, and it reached further than the flattening path: this change also neutralized explicitly in the xlsx *streaming* branch, which is not a CSV writer at all.

`flattenObject()` returns raw values now. Neutralization happens where CSV text is produced, in two places:

- **`processCSVvalue`**, which this PR already added and which covers every cell the streamed writer emits, headers included. That is what makes removing it from the flattening path safe rather than a regression.
- **`convertData`'s csv branch**, which does *not* go through `processCSVvalue` — it hands rows straight to json2csv. Without adding it there, taking it out of `flattenObject` would have quietly removed the protection from the buffered csv export. Field names go through it too: json2csv writes them as the header row, and a segmentation key is attacker supplied.

The xlsx branch writes `paramList` and `values` unchanged again, and json comes back untouched — there is a case for that now.

**One more thing, found while running it:** this PR's own suite was failing. The assertions were written against a backtick marker while the implementation prepends an apostrophe, so the single case that actually checked the marker failed, and four `indexOf("\`").should.equal(-1)` assertions passed for any output whatsoever, neutralized or not. They read the marker from one place now.

10 passing, from 8 passing and 1 failing before.

Cookiezaurs (approved):

PR #7885, opened 2026-08-04
chore(security): standardise the default response headers

ar2rsawseen:
Standardises the shipped security response headers so the three repos agree, and drops the ones that should no longer be sent.

## Removed

**`X-XSS-Protection`.** Deprecated and removed from every current browser, so it does nothing. While it existed it could be turned against the page: steered into disabling one of the page's *own* scripts, and with `mode=block` the aborted load is observable cross-origin, which makes it an oracle for reading page contents. It belongs on the remove list, and scanners flag its presence.

**The `preload` token on HSTS.** Preload requires submitting the domain to the browser preload list and is painful to reverse; with `includeSubDomains` on a shared parent domain it can affect a customer's other services. `max-age` and `includeSubDomains` are unchanged — this only stops us shipping the irreversible part as a default.

## Added

`Referrer-Policy`, `Permissions-Policy`, `Cross-Origin-Opener-Policy`, `Cross-Origin-Resource-Policy`, and `X-Content-Type-Options` where it was missing.

## Not added, and why

**`Content-Security-Policy`.** The classic dashboard inlines `countlyGlobal` per request via `res.expose`, so its hash is not stable and a static policy would need `'unsafe-inline'` — which removes the protection CSP exists for. It needs that inline script dealt with, or per-request nonces, and a flat config string cannot express a nonce. This is why a CSP attempt here previously broke things rather than helping.

**`Cross-Origin-Resource-Policy` on the api set.** That set carries `Access-Control-Allow-Origin: *` because web SDKs post from arbitrary customer origins. The two work against each other.

## The embeddable-route pattern

The isolation headers are applied globally, like the rest, and removed on the routes customers embed from their own origin — which is exactly what those routes already do for `X-Frame-Options`.

`Cross-Origin-Opener-Policy` matters there specifically: an SDK may open a widget as a *window* rather than an iframe, and `same-origin` would sever `window.opener` along with any callback the SDK expects.

## Existing installs need the upgrade script

Config defaults only seed keys that are **absent** — `checkConfigs` uses `getObjectDiff`, which copies a default across only when the stored value is `undefined`. So an install that already has these keys keeps its original values forever, and a change here would reach new installs only. Without the migration this PR would be cosmetic for the existing fleet.

The script **edits rather than overwrites**, so anything an operator added by hand survives: it drops the deprecated header, strips only the `preload` token from HSTS while leaving `max-age`/`includeSubDomains`, and appends what is missing. It is idempotent.

12 unit tests cover the rewrite: what it removes, what it adds, what it preserves, ordering, CRLF input, blank lines, case-insensitive header matching, and non-string input. It edits customer configuration, so both directions are asserted.

---

### Removed after review: the widget frame-ancestors work

An earlier revision also scoped which origins may frame the SDK-embeddable widget routes, driven off each app's `allow_access_control_origin` list. That has been taken out — it was a separate feature that had grown alongside the header standardisation, and it does not earn its place.

**The widget page has nothing to protect.** The app key is read from the query string by the page itself and ships inside published SDKs, so it is public either way. No session, no credential, and same-origin policy already stops a framing parent reading the frame's contents.

**The data path does not need the frame.** A response is submitted with a plain `GET /i/feedback/input?...`, so anyone who wants to write into an app's ratings does it directly — no victim, no iframe. A framing restriction cannot protect a public endpoint that is reachable without framing.

Against that it cost a `feedback_widgets` lookup plus an `apps` lookup on **every widget render** — a public, unauthenticated, SDK-driven route — and a report-only mode that could not safely be switched to enforce, since `allow_access_control_origin` is labelled a CORS list and says nothing about framing.

**Kept:** the `Cross-Origin-Opener-Policy` removal on those routes. That one follows from this change rather than being a feature — COOP is now a shipped dashboard default and would sever `window.opener` for an SDK that opens the widget as a window. `X-Frame-Options` was already being removed there before this branch.

What remains is the headers themselves, the nginx defaults, and the migration that gets them to existing installs — config defaults only seed keys that are absent, so a change to a shipped default otherwise reaches new installs only.


bin/upgrade/DEV/scripts/standardize_security_headers.js
```diff
@@ -0,0 +1,113 @@
...
+    const kept = [];
+
+    for (const line of lines) {
+        const name = line.split(':')[0].trim().toLowerCase();
+        if (name === 'x-xss-protection') {
+            continue;
+        }
+        if (name === 'strict-transport-security') {
+            // remove only the preload token, keep max-age and includeSubDomains
+            kept.push(line.replace(/;\s*preload\b/i, ''));
```
ar2rsawseen: [P1] Preserve operator-managed HSTS preload

This rewrites every stored HSTS line, including configurations where an operator deliberately opted into browser preload and submitted the domain to the preload list. Removing `preload` during an upgrade does not undo that browser-side commitment, but it does make the served policy inconsistent and can prevent continued preload eligibility. Remove it from the shipped default for new installs, but preserve it when migrating an existing custom header block.

bin/upgrade/DEV/scripts/standardize_security_headers.js
```diff
@@ -0,0 +1,113 @@
...
+        if (Object.keys(update).length) {
+            await db.collection('plugins').updateOne({_id: 'plugins'}, {$set: update});
+            console.log('Security headers updated');
+        }
+        else {
+            console.log('Nothing to update');
+        }
+    }
+    catch (err) {
+        console.error('Error while standardizing security headers', err);
```
ar2rsawseen: [P1] Fail the upgrade when the migration fails

The catch only logs the database error, then closes the connection and lets Node exit successfully. The upgrade runner can therefore mark this step complete even though none of the header changes were applied, and it may never retry it. Set a non-zero exit status or rethrow after closing the database so deployment stops and the migration can be rerun safely.

bin/upgrade/DEV/scripts/standardize_security_headers.js
```diff
@@ -0,0 +1,113 @@
...
+ */
+
+const pluginManager = require('../../../../plugins/pluginManager.js');
+
+const KEYS = ['dashboard_additional_headers', 'api_additional_headers'];
+
+const REQUIRED = [
+    {name: 'X-Content-Type-Options', line: 'X-Content-Type-Options: nosniff'},
+    {name: 'Referrer-Policy', line: 'Referrer-Policy: strict-origin-when-cross-origin'},
+    {name: 'Permissions-Policy', line: 'Permissions-Policy: camera=(), microphone=(), geolocation=(), payment=()'}
```
ar2rsawseen: [P1] Migrate the dashboard COOP header too

Existing installations with `security.dashboard_additional_headers` set never receive the new default, yet this migration's `REQUIRED` list omits `Cross-Origin-Opener-Policy`. The same change already removes COOP from the embeddable rating response, so leaving it out here means upgraded dashboards remain without the intended opener isolation indefinitely. Add `Cross-Origin-Opener-Policy: same-origin-allow-popups` for the dashboard key (while preserving the widget exemption).

bin/upgrade/DEV/scripts/standardize_security_headers.js
ar2rsawseen: Confirmed and fixed in 614b6b3f0d6. The reasoning is the part that matters: removing the token does not undo the browser-side commitment, so the only effect on an operator who opted in deliberately is an inconsistent policy and a risk to their place on the list.

Removal is now restricted to the exact line this product seeded:

```js
const SEEDED_HSTS = 'strict-transport-security:max-age=31536000; includesubdomains; preload';
```

compared with whitespace collapsed, since the stored value has been through a text field and may have picked up spacing. Any other `Strict-Transport-Security` line is left exactly as found.

The shipped default for new installs no longer carries preload, which was the other half of your suggestion.

Two cases added: an operator line with a different max-age keeps preload, and the seeded line loses it whatever the spacing.

bin/upgrade/DEV/scripts/standardize_security_headers.js
ar2rsawseen: Confirmed and fixed in 614b6b3f0d6. This was the worst of the three, because the failure is silent *and* permanent: the runner records the step as done, advances the database version, and nothing else rewrites these headers, so a transient database error meant the migration never ran and never would.

```js
catch (err) {
    console.error('Error while standardizing security headers', err);
    db.close();
    process.exit(1);
    return;
}
```

Connection closed first so the process does not exit with a socket open, then a non-zero status so the deployment stops and the step is retried.

bin/upgrade/DEV/scripts/standardize_security_headers.js
ar2rsawseen: Confirmed and fixed in 614b6b3f0d6. The asymmetry you describe is exactly it: the new default carries COOP, config defaults only ever seed keys that are **absent**, so an install that already had `dashboard_additional_headers` would have gone without opener isolation indefinitely while a fresh install got it.

```js
//dashboard only: the api block is not a browsing context, and the widget responses
//remove it for themselves
const REQUIRED_DASHBOARD = [
    {name: 'Cross-Origin-Opener-Policy', line: 'Cross-Origin-Opener-Policy: same-origin-allow-popups'}
];
```

`rewrite()` takes the key now and appends this list only for the dashboard block. The widget exemption is preserved because it does not live here — the rating routes strip COOP from their own responses, which is what makes adding it to the dashboard block safe.

Three cases added: added for the dashboard key, not added for the api key, and not duplicated when an operator already set one.

Cookiezaurs (approved):

PR #7887, opened 2026-08-04
fix(star-rating): check consent link destinations as urls (24.05)

ar2rsawseen:
Backport of #7886 to `release.24.05`, which carries the same code. See that PR for the full analysis.

Consent links are rendered into anchors in the drawer preview and the public popup, and nothing checked that a destination was a usable url — so anything typed into that field became the `href`.

Both render paths escape the value, and escaping cannot cover this: it touches `<`, `>` and `&`, and `javascript:alert(1)` contains none of them, so the value arrives unchanged and the browser runs it when the link is clicked.

Checked in three places, because they are three runtime contexts that have to agree or a value refused in one renders in another: the widget endpoint (in the shared `links` preprocessor, which both create and edit run), the dashboard drawer, and the popup page, which shares no code with the dashboard.

**The render-side checks are not redundant.** Widgets stored before this change still carry whatever was saved; without them those links keep working until someone re-saves the widget.

Accepted: `http(s)`, a root-relative path, or a fragment, matching what `countlyCommon`'s `onTagAttr` already permits for an `href` elsewhere. Leading whitespace and control characters are stripped first, since a browser ignores them when resolving a url, and protocol-relative `//host` is refused.

Also escapes link labels before building the regular expression they are matched with: unescaped, a label is a *pattern*, and one like `(a+)+$` turns matching into catastrophic backtracking.

## Notes on this branch

The api and drawer changes applied unchanged. The popup template differs here, so that part was hand-authored to match.

58 unit tests, all passing here. Full unit suite **101 passing before, 159 after**, same 5 pre-existing failures. eslint clean.

### DIFF
--- a/plugins/star-rating/frontend/public/javascripts/countly.views.js
+++ b/plugins/star-rating/frontend/public/javascripts/countly.views.js
@@ -1,4 +1,38 @@
 /*global $, countlyReporting, countlyGlobal, CountlyHelpers, starRatingPlugin, app, jQuery, countlyPlugins, countlyCommon,  CV, countlyVue, moment, countlyCohorts*/
+
+/**
+ * Escape regular expression metacharacters in a literal.
+ *
+ * textValue is interpolated into a RegExp below. Unescaped, the value IS a pattern,
+ * so one like `(a+)+$` turns matching into catastrophic backtracking on a long
+ * enough finalText.
+ *
+ * @param {string} value - literal to be matched
+ * @returns {string} the literal, safe to embed in a pattern
+ */
+function escapeForRegExp(value) {
+    return String(value).replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
+/**
+ * Whether a consent link destination is usable as an href.
+ *
+ * HTML escaping is no protection here: `javascript:alert(1)` contains no HTML
+ * metacharacter, so it survives escaping unchanged and runs when the link is clicked.
+ * The accepted set matches the widget endpoint and the public popup: http(s), a
+ * root-relative path, or a fragment. Leading whitespace and control characters are
+ * stripped first, because a browser ignores them when resolving a URL.
+ *
+ * @param {string} value - the stored link destination
+ * @returns {boolean} true when it may be used as an href
+ */
+function isSafeConsentLink(value) {
+    if (typeof value !== 'string') {
+        return false;
+    }
+    return /^(?:https?:\/\/|\/(?!\/)|#)/i.test(value.replace(/[\u0000-\u0020]+/g, ''));
+}
+
 (function() {
     var FEATURE_NAME = 'star_rating';
     var CLY_X_INT = 'cly_x_int';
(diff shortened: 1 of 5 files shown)


ar2rsawseen:
Pushed a follow-up commit that narrows the accepted set to http(s) only.

The first pass allowed a root-relative path and a fragment as well, on the reasoning that `countlyCommon`'s `onTagAttr` permits those for an href elsewhere in the dashboard. The closer precedent is the surveys widget, which renders the same kind of consent link and already accepts http(s) only, falls back to `about:blank`, and sets `rel` on the anchor (`plugins/surveys/frontend/public/javascripts/countly.common.components.js`). The two widgets sit next to each other and do the same job, so having them disagree on what a link may point at is the outcome worth avoiding.

Dropping relative paths costs nothing here: the popup is served from the Countly server, so `/terms` would resolve against the server rather than against the site the widget is embedded in.

Also in this commit:

- both renderers fall back to `about:blank` rather than an empty href, which would have pointed the link at the current page and reloaded it on click
- both anchors carry `rel="noopener noreferrer"`
- the drawer escapes the href it builds, since that string is handed to `v-html` and a value that passes the scheme check can still carry a quote
- the CHANGELOG line no longer mentions relative paths and fragments

Tests updated to match: `/terms` and `#terms` moved to the refused set, a whitespace-padded url and a `mailto:` case added, and two cases assert the render sites use the `about:blank` fallback and set `rel`. 68 passing, `eslint` clean.

plugins/star-rating/frontend/public/javascripts/countly.views.js
```diff
@@ -147,8 +182,21 @@
...
+                        //because the endpoint now rejects it: widgets saved before that check
+                        //still carry whatever was stored, and this drawer also renders values
+                        //that have not been saved at all.
+                        //about:blank rather than an empty href, which would point the link back
+                        //at the dashboard and reload it on click.
+                        const href = isSafeConsentLink(link.linkValue) ? link.linkValue.trim() : 'about:blank';
+                        //This string is rendered with v-html, so escape the href: a value that
+                        //passes the scheme check can still carry a quote and close the attribute.
+                        const escHref = countlyCommon.encodeHtml(href);
+                        finalText = finalText.replace(regex, `<a href="${escHref}" target="_blank" rel="noopener noreferrer">${link.textValue}</a>`);
```
ar2rsawseen: [P1] Escape the link label before returning this HTML

`finalTxt` is rendered with `v-html`, but only the destination is encoded here; `link.textValue` is interpolated back into the HTML string verbatim. A stored entity-encoded label such as `&lt;img src=x onerror=alert(1)&gt;` is decoded by the browser when another user opens the drawer, and an unsaved raw label is even more direct. Please HTML-encode the label as well (and match against the correspondingly encoded sentence), or build the anchor as a Vue/DOM node so label text never passes through an HTML sink.

ar2rsawseen: Confirmed and fixed in 80488b59d29 — and the sentence around the label had the same problem, so it is fixed at that level rather than per label.

`finalText` is the operator's consent sentence and it was going into `v-html` raw; only the href was being encoded. So rather than encoding just `link.textValue`, the sentence is encoded first and the anchors are added afterwards:

```js
finalText = countlyCommon.encodeHtml(finalText);
links.forEach(link => {
    const label = countlyCommon.encodeHtml(link.textValue);
    const regex = new RegExp(`\b${escapeForRegExp(label)}\b`, 'g');
    ...
    finalText = finalText.replace(regex, `<a href="${escHref}" ...>${label}</a>`);
});
```

That makes the anchors the only markup in the string, which is the reason `v-html` is used at all. As you note, the matching then has to happen against the **encoded** label, since the sentence being searched is encoded too — for an ordinary label like "Terms and Conditions" encoding is a no-op, so nothing changes for the normal case.

The other branch of `finalTxt`, the one that returns `links.finalText` when `links` is not an array, reaches the same sink with no anchors to add, so it is encoded on the way out too.

I did not rebuild the anchor as a DOM node: that would mean giving up `v-html` and restructuring how the drawer renders the sentence, and encoding at the sentence level gets the same guarantee — nothing but the anchors is markup.

Cookiezaurs (approved):

PR #7890, opened 2026-08-05
[fix][populator] keep a template while other apps still have environments from it

ar2rsawseen:
Removing a populator template deleted **every** app's environment records for it, selecting by `templateId` alone, while deleting environment *users* only for the app making the request:

```js
// users: scoped to the requesting app
{_id: {$regex: new RegExp("^" + params.qstring.app_id + "_" + template_id)}}

// records: every app
{templateId: ob.params.qstring.template_id}
```

Two deletes in the same code path, disagreeing about scope. Another app was left holding generated users with no record to manage them by.

That app could not recover from it either. An environment is only ever reachable *through* its template — the templates table marks which templates have environments, and the environment view is routed by template id:

```js
if (self.environments.filter(x => x.templateId === item._id).length) { item.hasEnvironment = true; }
self.filterByEnvironmentOptions = envs.filter(x => x.templateId === self.templateId)
```

So once the shared template is gone there is no row to match and no route to follow. Those users cannot be listed, reused or removed.

### Why keep the template rather than cascade

Deleting the other apps' rows would tidy the database at the cost of discarding another app's records at the request of someone with no access to that app. So the remove path now looks for environments belonging to other apps first and refuses, saying how many. Each app clears its own through `/o/populator/environment/remove`, which already exists, is already app-scoped, and is already wired into the environment view — so there is no new UI and no new concept.

The environment delete that follows is now scoped to the requesting app too, matching every other query against that collection (`environment/check`, `environment/list`, `environment/remove`, the app delete/reset/clear hooks and the age-based cleanup job were all already scoped — this line was the only exception). The check above it means no other app has one at that point; if one appears in between, leaving a row for a cleanup job to collect beats deleting someone else's.

### Behaviour change

A template delete that would previously have destroyed other apps' environment records now returns 400 with the count. Deleting a template nobody else has environments for is unchanged.

### Templates stay shared

Any member with the populator feature on any app can still list, use, edit and remove templates. That is deliberate — a template describes how to generate synthetic data and holds no collected data — so the intent is now recorded in a comment beside the code, since an unscoped query is otherwise indistinguishable from a missing check.

### Verification

`plugins/populator/tests.js` is supertest-based against a live instance and has no fixture for a second app, so this is not covered there. Instead the function was lifted and run against stubbed collections to exercise both paths: with another app's environment present it returns 400, names the count, and issues no template or environment delete at all; with none present the template is removed, the response is 200, and the environment delete carries both `templateId` and `appId`. `node --check` and `npx eslint` clean.


plugins/populator/api/api.js
```diff
@@ -115,33 +122,58 @@ const FEATURE_NAME = 'populator';
...
-                    });
+            //Because the template is shared, other apps may have generated environments from it.
+            //An environment is only ever reachable through its template: the templates table
+            //marks which ones have environments, and the environment view is routed by template
+            //id. So removing the template would leave those rows with nothing able to list,
+            //reuse or delete them. Deleting them here instead would discard another app's
+            //records at the request of someone who has no access to that app. Refuse instead,
+            //and let each app remove its own environment first, through
+            ///o/populator/environment/remove.
+            common.db.collection('populator_environments').find({
```
ar2rsawseen: [P2] Do not let environments from deleted apps block the template forever

The app-deletion hook removes only `populator_environment_users`; it leaves that app's `populator_environments` documents behind. This new query then treats those stale rows as live cross-app blockers, but no user can switch to the deleted app to call the suggested environment-removal endpoint, so the shared template becomes undeletable. Please clean these rows during app deletion or join/filter against applications that still exist before returning this 400.

plugins/populator/api/api.js
```diff
@@ -115,33 +122,58 @@ const FEATURE_NAME = 'populator';
...
+            }).toArray(function(errOtherEnvs, otherEnvs) {
+                if (errOtherEnvs) {
+                    log.e("Error checking populator environments before deleting template", errOtherEnvs);
+                    common.returnMessage(obParams, 500, errOtherEnvs.message);
+                    return false;
                 }
-                else {
-                    common.returnMessage(ob.params, 500, removeTemplateErr.message);
+                if (otherEnvs && otherEnvs.length) {
+                    common.returnMessage(obParams, 400, "Cannot remove this template: " + otherEnvs.length + " environment(s) in other applications were generated from it. Those applications have to remove their environments first.");
```
ar2rsawseen: [P2] Propagate this actionable error to the dashboard

This branch adds a specific recovery instruction, but `countlyPopulator.removeTemplate` discards the response body and always shows the generic `populator.failed-to-remove-template` notification. A dashboard user cannot learn which condition blocked deletion or that environments in other applications must be removed first. Please surface this response message (or add a dedicated error code the UI maps to the same instruction).

plugins/populator/api/api.js
ar2rsawseen: [P2] Close the check-then-delete race across applications

This existence check and the later template removal are separate operations. Another app can create an environment from the shared template after this cursor returns empty but before `populator_templates.remove()` runs; the template is then deleted while the newly created cross-app environment is left orphaned—the exact state this change is meant to prevent. Please serialize creation/deletion with a transaction or a template-level deleting state/lock that environment creation also honors.

ar2rsawseen: Confirmed and fixed in 64d4e4fd05e — and the hook was the odd one out, which is what made it easy to miss.

`/i/apps/delete` removed `populator_environment_users` and stopped there, while `/i/apps/clear`, `/i/apps/clear_all` and `/i/apps/reset` sitting directly below it all remove `populator_environments` as well. So the rows were left behind only on the one path where the app itself is gone, which is precisely the path where nothing can ever reach them again.

The hook now removes them too. That is the cleanup half of your suggestion rather than the filter half: filtering the check against existing applications would leave the stale rows accumulating and would put a per-app lookup in front of every template deletion, while the rows have no reason to exist once their app does not.

plugins/populator/api/api.js
ar2rsawseen: Confirmed and fixed in 64d4e4fd05e. The message was written to tell the user what to do and then thrown away by the caller, which made it decoration.

```js
error: function(xhr) {
    var reason = "";
    try {
        var body = JSON.parse(xhr && xhr.responseText);
        reason = (body && (body.result || body.message)) || "";
    }
    catch (ignored) {
        reason = "";
    }
    CountlyHelpers.notify({message: reason || CV.i18n("populator.failed-to-remove-template", templateId), type: "error"});
}
```

I took the response-message route rather than a dedicated error code the UI maps: the message already names how many environments and in what terms, and a code would need the count passed separately to say anything as useful. The generic i18n string stays as the fallback for a failure with no body — a network error, or a 500 with nothing to say.

plugins/populator/api/api.js
ar2rsawseen: Confirmed and fixed in 64d4e4fd05e. You are right that the two operations are separate and that the window produces exactly the orphaned row the check exists to prevent.

Implemented as the template-level mark rather than a transaction, since that does not require the deployment to be a replica set:

1. `findOneAndUpdate` sets `deletingAt` on the template — one document, so it is atomic, and it fails if another removal already holds it;
2. environment creation refuses a template carrying a live `deletingAt`, so anything that gets past that point is already visible to the check;
3. the check runs, and the mark is cleared on every path that does not go on to remove the template.

One thing I did differently from a plain boolean: the mark is a **timestamp**, and one older than 60 seconds can be taken over. A boolean flag introduces a worse failure than the one being fixed — a process dying between marking and removing would leave a template that nobody can delete and nobody can generate from, with no way to clear it short of a database edit. With a timestamp that state heals itself a minute later.

Verified the shape compiles and lints; the behaviour needs the populator suite, which wants the plugin test backing services.

Cookiezaurs (approved):

Cookiezaurs (approved):

Cookiezaurs (approved):

PR #7893, opened 2026-08-05
[fix][push] render message text as text in the notification preview (24.05)

ar2rsawseen:
The push details view builds its phone-style preview by splicing personalization elements into the message and handing the result to `innerHTML`, so it can walk the child nodes back out and turn each into a component:

```js
// countly.models.js
var messageInHTMLString = this.decodeHtml(message);      // then placeholders spliced in
...
htmlTitle.innerHTML = content;                            // parsed back out here
```

The message text travels inside that string, so a tag in the message was parsed as markup rather than shown as text, and its event handlers ran.

### Why the decode has to stay

It looks like the obvious thing to remove, but it is load-bearing. The keys in `messagePers` and `titlePers` are character offsets into the message *as it was typed*, and the API escapes on output, so `&` arrives as five characters instead of one. Decoding is what puts the string back into the coordinate space those offsets refer to. Drop it and every element after an escaped character lands in the wrong place.

That also rules out the simpler-looking `textContent = content`, which would remove the markup the walk exists to find and take the personalization chips with it.

### What changed

Placeholders now go in as opaque tokens, the whole string is escaped, and the tokens are swapped for their elements afterwards. Only markup generated by this code survives the escape.

The offset arithmetic is untouched — a token stands in for an element and is measured the same way, so every position it computes is the position it computed before. Escaping is a plain string replacement rather than `countlyCommon.encodeHtml`, which round-trips through `innerText` and would fold newlines into `<br>`, changing both the text and those offsets.

This covers the title as well as the message, since both go through `buildMessageText`. The element markup itself was already safe: `getUserPropertyElement` builds it with `setAttribute` and `innerText` and serialises with `outerHTML`, so the label, key and fallback were already escaped by the serialiser.

### Verification

Compared against the pre-change source, in a real DOM:

- the pre-change source does produce the element, which confirms the harness reaches the code path
- after the change no element and no event-handler attribute are produced, and the value shows as the text that was typed
- preview components are byte-identical across a single placeholder mid-text, one at the start, two placeholders, adjacent placeholders, text containing an ampersand, and a message with no personalization at all

`node --check` and `npx eslint` clean on all three branches. Not covered by an automated test in-repo: the push frontend has no DOM test harness, and adding one is out of scope here.

### DIFF
--- a/plugins/push/frontend/public/javascripts/countly.models.js
+++ b/plugins/push/frontend/public/javascripts/countly.models.js
@@ -924,7 +924,15 @@
                     message = "";
                 }
                 if (!userPropertiesDto) {
-                    return message;
+                    //the same round trip the personalization path makes below, for the same
+                    //reason: the result is handed to innerHTML in
+                    //getPreviewMessageComponentsList. Returning the value untouched relied on
+                    //the api having escaped it, which holds for a stored notification and not
+                    //for text typed into the editor, whose state this preview also renders.
+                    //
+                    //Decode then escape rather than escape alone: escaping an already escaped
+                    //message would show &lt;b&gt; to the user instead of what they wrote.
+                    return self.escapeMessageText(self.decodeHtml(message));
                 }
                 // var html = '',
                 //     keys = this.sortUserProperties(userPropertiesDto),
@@ -965,11 +973,19 @@
                 //length purposes, so every position it computes is the position it computed
                 //before. Only markup generated here survives the escape.
                 var messageInHTMLString = this.decodeHtml(message);
+                //A fixed token could be typed into the message. Then the replacement below
+                //cannot tell the author's text from ours: it becomes an extra personalization
+                //span, or disappears when the index is out of range, and the edit state carries
+                //that corruption back on save. A per call nonce cannot be guessed or typed.
+                //
+                //Length is not constrained - the arithmetic below measures the token it
+                //actually inserted - so this changes nothing about the offsets.
+                var tokenNonce = "CLYPERS" + Math.random().toString(36).slice(2, 10);
                 var placeholderElements = [];
                 var buildMessageLength = 0;
                 var previousIndex = undefined;
                 this.sortUserProperties(userPropertiesDto).forEach(function(currentUserPropertyIndex, index) {
-                    var userPropertyStringElement = "@@CLY_PERS_" + placeholderElements.length + "@@";
+                    var userPropertyStringElement = "@@" + tokenNonce + "_" + placeholderElements.length + "@@";
                     placeholderElements.push(self.getUserPropertyElement(currentUserPropertyIndex, userPropertiesDto[currentUserPropertyIndex]));
                     if (index === 0) {
                         messageInHTMLString = self.insertUserPropertyAtIndex(messageInHTMLString, currentUserPropertyIndex, userPropertyStringElement);
@@ -987,7 +1003,7 @@
                     previousIndex = currentUserPropertyIndex;
                 });
                 //The token has no character the escape touches, so it comes through intact.
-                return self.escapeMessageText(messageInHTMLString).replace(/@@CLY_PERS_(\d+)@@/g, function(match, tokenIndex) {
+                return self.escapeMessageText(messageInHTMLString).replace(new RegExp("@@" + tokenNonce + "_(\\d+)@@", "g"), function(match, tokenIndex) {
                     return placeholderElements[Number(tokenIndex)] || "";
                 });
             },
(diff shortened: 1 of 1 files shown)


plugins/push/frontend/public/javascripts/countly.models.js
```diff
@@ -900,6 +900,24 @@
...
+             * Escape text so it cannot contribute markup once the result reaches innerHTML.
+             *
+             * Deliberately a string replacement rather than countlyCommon.encodeHtml, which
+             * round-trips through innerText and would normalise newlines into <br>, changing
+             * the text and the offsets the personalization indexes rely on.
+             *
+             * @param {string} str - untrusted text
+             * @returns {string} the text, safe to place in an html string
+             */
+            escapeMessageText: function(str) {
```
ar2rsawseen, line 913: [P1] Escape messages that have no personalization metadata

`buildMessageText()` returns early when `userPropertiesDto` is absent, before this new escaping helper is ever called. Ordinary notifications without personalization therefore still return the raw decoded message to `getPreviewMessageComponentsList`, which assigns it to `innerHTML`, so a stored `` message still executes in the preview. Escape the message on the no-properties path as well; only the generated personalization spans should bypass text escaping.

plugins/push/frontend/public/javascripts/countly.models.js
```diff
@@ -936,11 +954,23 @@
...
+                //is untouched: a token takes the element's place and is the same string for
+                //length purposes, so every position it computes is the position it computed
+                //before. Only markup generated here survives the escape.
                 var messageInHTMLString = this.decodeHtml(message);
+                var placeholderElements = [];
                 var buildMessageLength = 0;
                 var previousIndex = undefined;
                 this.sortUserProperties(userPropertiesDto).forEach(function(currentUserPropertyIndex, index) {
-                    var userPropertyStringElement = self.getUserPropertyElement(currentUserPropertyIndex, userPropertiesDto[currentUserPropertyIndex]);
+                    var userPropertyStringElement = "@@CLY_PERS_" + placeholderElements.length + "@@";
```
ar2rsawseen: [P2] Preserve literal placeholder-looking message text

The replacement cannot distinguish internal tokens from literal notification content such as `@@CLY_PERS_0@@`. When at least one personalization exists, an author's literal text is replaced with an extra generated span (or removed for an out-of-range index), corrupting the preview and potentially the message when it is edited again. Use a collision-proof sentinel/structured segment assembly, or escape pre-existing token-shaped text before inserting and replacing these placeholders.

plugins/push/frontend/public/javascripts/countly.models.js
ar2rsawseen: Fixed in bf7d37068e2 — with one correction to the mechanism, which changed the fix.

You are right that the early return skips the escaping and that the result reaches `innerHTML`. But the value on that path is not the *decoded* message: it is whatever came in, and for a stored notification the api has already escaped it. Escaping it again, which is what the note asks for, would have shown the user `&lt;b&gt;` where they wrote `<b>` — a visible regression on every ordinary notification.

Where it does bite is the other input to this function. The preview renders `pushNotificationUnderEdit.message[locale]`, which is the **editor's live state**: for a message being typed that is raw input that has been through no escaping at all.

So the path now makes the same round trip the personalization path already makes:

```js
return self.escapeMessageText(self.decodeHtml(message));
```

Decode then escape, which normalises either shape to exactly one level of escaping. I checked both rather than reasoning about it: an api-escaped `&lt;img src=x onerror=alert(1)&gt;` comes back byte-identical, and a raw typed `` comes back neutralised.

plugins/push/frontend/public/javascripts/countly.models.js
ar2rsawseen: Confirmed and fixed in bf7d37068e2. Both failure modes are real — the extra span and the disappearing text — and the second consequence you name is the worse one: the corrupted preview is the edit state, so it is what gets saved.

Went with a per-call nonce rather than escaping pre-existing token-shaped text:

```js
var tokenNonce = "CLYPERS" + Math.random().toString(36).slice(2, 10);
...
var userPropertyStringElement = "@@" + tokenNonce + "_" + placeholderElements.length + "@@";
```

Escaping the existing text would have been the more obvious fix and is the one thing that cannot work here: it changes the message's length, and the personalization indexes are offsets into that exact string, so every span after the first literal would land in the wrong place.

The nonce costs nothing in that respect. Token length is not constrained — the arithmetic measures the token it actually inserted, `userPropertyStringElement.length` — so the offset handling is untouched, and a literal the author types cannot match a value generated per call.

cihadtekin (approved):

PR #7895, opened 2026-08-05
[fix][core] keep caller input out of the export query specification (24.05)

ar2rsawseen:
`/o/export/requestQuery` re-runs an internal endpoint and uses the collection and pipeline that endpoint returns. That is sound while the specification is the endpoint's own work: the endpoint authorises the request and builds the query itself, so the query inherits that authorisation. `views`, `heatmaps` and `surveys` all work this way, returning `{db, collection, pipeline, projection}` from a dedicated branch.

It stopped being the endpoint's own work because the caller's `data` was forwarded as the inner request's body, and `processRequest` copies every body key onto `params.qstring`:

```js
if (params.req.body && typeof params.req.body === "object") {
    for (let i in params.req.body) { params.qstring[i] = params.req.body[i]; }
}
```

So anything in `data` reaches the endpoint indistinguishable from a query parameter. An endpoint that keeps parameters it does not recognise and then returns what it stored will hand a caller-supplied `collection` and `pipeline` straight back, and the export runs it. The authorisation that took place was for the endpoint's own action and had nothing to do with the resulting read.

### What changed

**`data` is no longer forwarded.** No caller sends it: `views`, `heatmaps`, `surveys`, the shared datatable export and `countly.helpers` all pass their parameters in the path's query string. So the inner body is empty and there is nothing for any endpoint to echo.

**Write stages are refused**, as a second and independent check. No endpoint building an export query has any reason to write, and none of the three emits one, so a specification containing `$out` or `$merge` is a bug or worse. The search covers sub-pipelines inside `$lookup`, `$unionWith` and `$facet`, so a write cannot sit one level down.

`exports.fromRequest` forges the same request shape but exports the endpoint's *response* rather than running a query built from it, so it is deliberately untouched here.

### Also included: only declared producers may be named

The endpoint named in `path` decides what the export reads, so it is now refused unless it declares itself by answering `/export/query/producers`, giving the path it answers on, the parameter values that select its export branch, and the database its query targets. `views` declares itself here; `heatmaps` and `surveys` do so in countly-enterprise-plugins (PRs #3344 and #3345), and `heatmaps` also in countly-platform.

The path is parsed rather than substring-matched, since a repeated parameter leaves `method=views` present in a string whose effective value is something else. Paths outside `/o` are refused outright, because a write endpoint is never a producer. The declared parameters are then pinned into the inner request so the caller cannot select a producer and steer it into a different mode. `db` comes from the declaration too: the surveys client sends `db=countly_drill` from the browser today, which is why that had to move to the producer.

The three producers have no shared convention — `views` is `/o` plus `method=views&action=getExportQuery`, `heatmaps` is its own path `/o/heatmaps/export` with no parameters, `surveys` is `/o/surveys/survey/data` plus `method=export` — so the contract is path plus required parameter values rather than fixed fields.

### Deliberately not included

- **Restricting pipeline stages by role.** The existing `dbviewer` aggregation guard cannot be reused as-is: `views` legitimately emits `$lookup` and `surveys` legitimately emits `$unionWith`, both of which `ALLOWED_STAGES_USER` excludes, and `sanitizeAggregation` strips rather than rejects. Applying it would quietly return wrong export data to non-global members.

### Verification

The write-stage walk was exercised against pipelines shaped like all three real producers (which pass untouched) and against `$out`/`$merge` at the top level, inside a `$unionWith` sub-pipeline, inside `$facet`, and two levels down inside `$lookup`, plus non-array and null inputs. `node --check` and `npx eslint` clean on the countly-server branches. On countly-platform, `tsc` reports the same 2 pre-existing errors in this file and the same 5876 repo-wide, before and after.

### DIFF
--- a/api/parts/data/exports.js
+++ b/api/parts/data/exports.js
@@ -741,19 +741,30 @@ exports.fromRequest = function(options) {
  * hidden one level down. Names are matched as object keys, which is the only place a stage
  * operator appears.
  *
+ * The node budget is a stop on runaway work, not a verdict. Exhausting it means the
+ * pipeline could not be verified, which is a reason to refuse rather than to proceed:
+ * returning "nothing found" there would let a large enough structure in front of an
+ * $out carry the write through.
+ *
  * @param {Array} pipeline - aggregation pipeline to inspect
- * @returns {string|null} the offending stage name, or null when there is none
+ * @returns {string|null} why the pipeline must be refused, or null when it is safe
  */
 function findWriteStage(pipeline) {
     var WRITE_STAGES = ["$out", "$merge"];
+    var MAX_NODES = 10000;
     var seen = 0;
+    var exhausted = false;
 
     /**
      * @param {*} node - value to walk
      * @returns {string|null} offending stage name or null
      */
     function walk(node) {
-        if (seen++ > 10000 || !node || typeof node !== "object") {
+        if (seen++ > MAX_NODES) {
+            exhausted = true;
+            return null;
+        }
+        if (!node || typeof node !== "object") {
             return null;
         }
         if (Array.isArray(node)) {
@@ -776,7 +787,12 @@ function findWriteStage(pipeline) {
         }
         return null;
     }
-    return walk(pipeline);
+    var found = walk(pipeline);
+    if (found) {
+        return "write stage " + found;
+    }
+    //fail closed: unverified is not the same as verified safe
+    return exhausted ? "pipeline too large to verify (over " + MAX_NODES + " nodes)" : null;
 }
 
 exports.fromRequestQuery = function(options) {
@@ -815,7 +831,7 @@ exports.fromRequestQuery = function(options) {
                 //or tampered spec from turning a read into a write.
                 var writeStage = findWriteStage(body.pipeline);
                 if (writeStage) {
-                    log.e("Refusing export query containing a write stage: " + writeStage);
+                    log.e("Refusing export query: " + writeStage);
                     var refused = new Transform({
                         objectMode: true,
                         transform: (data, _, done) => {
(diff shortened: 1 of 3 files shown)


api/parts/data/exports.js
```diff
@@ -734,6 +734,51 @@ exports.fromRequest = function(options) {
...
+function findWriteStage(pipeline) {
+    var WRITE_STAGES = ["$out", "$merge"];
+    var seen = 0;
+
+    /**
+     * @param {*} node - value to walk
+     * @returns {string|null} offending stage name or null
+     */
+    function walk(node) {
+        if (seen++ > 10000 || !node || typeof node !== "object") {
```
ar2rsawseen: [P2] Fail closed when the traversal budget is exhausted

Once `seen` passes 10,000 this guard returns `null`, which is also the value meaning “no write stage found.” A pipeline with a sufficiently large user-influenced `$or`/`$in` structure before a later `$out` or `$merge` therefore causes the security check to stop walking and permits the write stage to execute. Exceeding the budget should produce an explicit rejection (or throw a handled validation error), not a successful no-match result.

ar2rsawseen: Confirmed and fixed in a00d2bfe274. The budget was a stop on runaway work being read as a verdict — and the two states shared a return value, which is what made it a fail-open: `null` meant both "walked the whole thing, found nothing" and "gave up part way".

Exhaustion is tracked separately now:

```js
if (seen++ > MAX_NODES) {
    exhausted = true;
    return null;
}
...
const found = walk(pipeline);
if (found) {
    return 'write stage ' + found;
}
//fail closed: unverified is not the same as verified safe
return exhausted ? 'pipeline too large to verify (over ' + MAX_NODES + ' nodes)' : null;
```

The caller refuses on any non-null value, so a pipeline the walk could not finish is refused rather than run. `null` now means only what it should: the walk completed and found nothing.

The log line reads "Refusing export query: <reason>" instead of naming a write stage that may not exist.

Cookiezaurs (approved):

PR #7918, opened 2026-08-12
[fix][reports] build a report from declared fields, and derive the rest at send time (24.05)

ar2rsawseen:
## What was wrong

`/i/reports/create` stored the submitted `args` object as the report document, and `/i/reports/update` passed the same object as `$set`. Only `report_type === "core"` reached an app authorization check, so `"dashboards"` walked past it, and nothing limited *which fields* a request could set.

A report document is not only configuration. `reports.send` reads `report.messages[i].html` and makes the trusted template a fallback:

```js
let html = report.messages && report.messages[i] && report.messages[i].html;
if (!html && message.data && message.template) { // report from dashboard
```

With `sendPdf`, that string is what `pdf.renderPDF` opens, and `api/utils/pdf.js` navigates to it as a `data:` document with `waitUntil: 'networkidle0'` and `setBypassCSP(true)`. So a field nobody intended to be writable was renderer input, and the render happens on the server with its network reach.

The dashboards report builder never clears it either: `plugins/dashboards/api/api.js` sets `report.subject` and `report.data` and leaves `messages` alone.

## Three changes, each sufficient on its own

**1. Create and update build the document from a declared list.** `REPORT_FIELDS` is what the drawer binds (`title`, `report_type`, `apps`, `dashboards`, `date_range`, `emails`, `frequency`, `day`, `hour`, `minute`, `timezone`, `sendPdf`, `selectedEvents`) plus the `metrics` map `onSubmit` assembles. `_id` and `user` fall out of the list rather than needing their own `delete`.

The direction is deliberate. Removing known-dangerous keys leaves every field added later writable until someone remembers to deny it, and the cost of forgetting is unbounded because these values are consumed, not just stored. Declaring what the drawer sends fails the other way: the cost of forgetting is a setting that stops saving, which someone notices and files.

**2. `getReport` strips what the generator derives.** `messages`, `data`, `subject`, `mailTemplate`, `properties`, `period`, `start`, `end`, `date`, `total_new`, `universe` are all assigned at send time; none needs to survive a database round trip. This closes documents stored before the allow-list existed.

It goes in `getReport` rather than `loadReport` because `/o/reports/preview` and `/o/reports/pdf` run their own `findOne` and call `getReport` directly. That is the one point all three paths share.

**3. The renderer may only fetch from Countly's own origins.** `renderPDF` intercepts requests and refuses any origin outside an allow-list: the configured Countly origin plus whatever the caller passes. `data:`, `blob:` and `about:` pass through, carrying no network request. Both report call sites pass the origins their templates load images from. Origins compare scheme, host and port together, since host alone would be too loose when Countly itself sits on loopback.

`web.host` and `web.port` are read from the frontend config rather than the api one, which is where they are actually defined, guarded so `pdf.js` still loads where that file is absent.

### --disable-web-security stays, and that is a reversal

An earlier commit on this branch removed it. Measured against real Chromium, that was wrong: the html is opened as a `data:` url, whose origin is opaque, and Chromium refuses http subresources from an opaque origin. With a single own-origin image, one request is served with the flag and **none** without it, so removing it would have shipped dashboard PDFs with every image missing.

What made the flag dangerous was that any request could be made at all. The allow-list settles that, and with it the flag can only relax checks for Countly's own origin.

## Verified against the unpatched file

`test/unit-tests/plugins.reports.public-fields.js` (`plugins/reports/tests/public-fields.unit.js` on platform), 3 cases: every derived field is stripped before the generator sees the report, the report's own configuration survives untouched, and no `args` array asks for `--disable-web-security`.

Against these branches all 3 pass. Against the unpatched file 2 of 3 fail, the exception being the case that asserts configuration is preserved, which should pass either way.

## Verified against real Chromium

Two loopback listeners, one on the port the config calls Countly's own and one on another port standing in for an internal service, with a document requesting an image from each.

| | own-origin image | foreign origin |
|---|---|---|
| before | served | **served** |
| after | served | **refused** |

### DIFF
--- a/api/utils/pdf.js
+++ b/api/utils/pdf.js
@@ -152,9 +152,14 @@ exports.renderPDF = async function(html, callback, options = null, puppeteerArgs
         log.d('pdf generated');
     }
     catch (error) {
-        log.d('Error:', error);
+        //a failed render used to be logged at debug and then replaced by a
+        //TypeError from the close below, so the actual reason (a missing chromium,
+        //a bad launch argument) never reached the log at all
+        log.e('pdf generation failed', error);
     }
     finally {
-        await browser.close();
+        if (browser) {
+            await browser.close();
+        }
     }
 };
--- a/api/utils/pdf.js
+++ b/api/utils/pdf.js
@@ -17,6 +17,52 @@ catch (err) {
     }
 }
 var log = require('./log.js')('core:pdf');
+var webConfig = {};
+try {
+    //web.host and web.port live in the frontend config, which is the origin the
+    //report templates point the renderer at. Optional: pdf.js is core and may be
+    //loaded where that file is absent, and callers pass their origins anyway.
+    webConfig = require('./../../frontend/express/config.js').web || {};
+}
+catch (error) {
+    log.d('no frontend config available for render origins', error.message);
+}
+
+/**
+* The origins a rendered document is allowed to fetch from.
+*
+* Rendering runs on the server, so anything the document requests is requested
+* from the server's network position: loopback, the private network, a cloud
+* metadata service. The report templates only ever point at Countly itself, at
+* `host` or at the localhost form the caller builds for exactly that reason, so
+* an allow-list of those costs nothing and closes the rest.
+*
+* Origins are compared as scheme, host and port together. Host alone would be
+* too loose, since Countly is frequently on loopback itself and every other
+* service on that interface would come along with it.
+* @param {array} extra - origins the caller knows it needs, e.g. its public host
+* @returns {Set} the set of allowed origins
+**/
+function renderOrigins(extra) {
+    var origins = new Set();
+    var protocol = process.env.COUNTLY_CONFIG_PROTOCOL || "http";
+
+    if (webConfig.host) {
+        origins.add(protocol + "://" + webConfig.host + (webConfig.port ? ":" + webConfig.port : ""));
+    }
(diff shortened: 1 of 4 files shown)


ar2rsawseen:
### Conflict resolved

The base gained the `authorized` fix for reports (`delete props.authorized` on the update path) while this branch replaced those ad-hoc deletes with the `publicReportFields` allow-list. The two collided on the same lines.

Kept **this branch's allow-list**, because it subsumes the deny-list rather than competing with it — verified rather than assumed, against the merged file:

```
publicReportFields({_id, user, authorized, title, frequency, apps, bogus})
  -> {"title":"T","apps":["a"],"frequency":"daily"}
  drops _id: true   drops user: true   drops authorized: true   drops bogus: true
```

`REPORT_FIELDS` names none of the three, so all three are dropped by construction, and the next such field is dropped without anyone having to remember it — which is the point of the change.

The reasoning the base added is not lost: I folded its explanation of `authorized` (the flag comes back through the object passed to the dispatch, so a submitted one must reach neither that nor `$set`) into the comment above the allow-list call, alongside the existing notes on `_id` and `user`.

eslint clean. I could not run `plugins.reports.public-fields.js` locally — it loads the reports plugin, which pulls `offline-geocoder`, installed only by CI's per-plugin `npm install` — so that suite is CI's to confirm.

Cookiezaurs (approved):

PR #7919, opened 2026-08-12
[fix][core] do not keep the request's credentials in stored documents

ar2rsawseen:
## Why

`api_key` and `auth_token` are accepted as request parameters (`api/utils/rights.js`). Any handler that keeps input it does not recognise therefore persists the caller's own credential into the document, and a read that returns the document hands that credential to everyone allowed to read it. That is exactly how a date preset came to disclose its owner's api key.

Four subsystems store a parsed client payload without declaring their own field set:

| Handler | Write paths | Read |
|---|---|---|
| `remote-config` parameters and conditions | 4 | dashboard list, unprojected |
| `alerts` | 1 | `/o/alert/list`, unprojected |
| `hooks` | 1 | `/o/hook/list`, unprojected |
| `ab-testing` experiments | 2 | dashboard, unprojected |

## What this does, and what it does not

`common.stripRequestCredentials(doc)` removes `api_key` and `auth_token` from an object about to be written. It is called on the parsed payload at each of those write paths.

**This is the floor, not the fix.** Only storing declared fields is the fix, and it needs a per-plugin survey of every writer plus the fields the UI round-trips through storage, which is a larger change with real regression risk for each plugin. This change is deliberately the part that needs no knowledge of the payload's shape: a handler that cannot enumerate itself can still refuse to keep a credential.

Top level only, deliberately. That is where a copy of the request puts them. A value the caller nested inside their own payload is their own to disclose, and descending to arbitrary depth would mean guessing at shapes. The test pins that boundary.

It sits next to `common.reqInfo`, which already keeps request `api_key` out of the logs for the same reason.

## Notes on the branches

- `release.24.05`'s hooks has two parse sites where master has one, the second being `/i/hook/test`, which runs a mock rather than storing. Stripped there as well, for consistency and at no cost.
- In countly-enterprise-plugins the two lines are written out rather than calling `common`, because ab-testing resolves `common` from the server tree at runtime and this way the change does not depend on which repo merges first.

## Tests

`test/unit-tests/api.utils.common.request-credentials.js`, 4 cases: both parameters removed, everything else untouched including a nested `api_key`, the same object returned so it can be used inline, and non-objects tolerated.


plugins/alerts/api/api.js
```diff
@@ -211,6 +211,7 @@ function getScheduleTextExpression(period, offset) {
             }
             try {
                 alertConfig = JSON.parse(alertConfig);
+                common.stripRequestCredentials(alertConfig);
```
ar2rsawseen, line 224: [P1] Remove credentials that are already stored during updates

Deleting these properties from the incoming object only keeps a new credential out of the later `$set`; MongoDB does not remove `api_key` or `auth_token` already present on a document created before this fix. Editing an affected alert, hook, remote-config item, or experiment therefore leaves the old credential readable. Please add `$unset` for both fields on update, redact them on every read, or include a migration/cleanup for existing documents.

ar2rsawseen: Confirmed and fixed in cb15e301f56. You are right, and it is the update path specifically that made it matter: editing an affected document is exactly when it would have been cleaned, and `$set` leaves a field it does not name untouched, so it was not.

A companion helper adds the removal to the update document:

```js
common.unsetRequestCredentials({$set: doc})
```

applied at every update path the strip already covers — alerts, hooks, the four remote-config ones, the experiment. The two helpers are meant to be used together and the comment says so: MongoDB refuses an update naming the same field in `$set` and `$unset`, so the strip is what makes the unset legal.

**On your third option, the migration:** I did not add one, and I think this makes it unnecessary rather than skipping it. Every affected document is cleaned the next time it is written, which is the same moment its owner has to touch it anyway, and it avoids a migration that would have to be written, ordered and reasoned about in three repositories for a field that clears itself. What it does not cover is a document nobody ever edits again — worth knowing, and worth a migration only if you want the guarantee rather than the trend.

Redacting on read I deliberately did not take: it hides the value while leaving it in the database, so a backup or a direct query still carries it.

Four cases added: both fields land in `$unset`, an existing `$unset` survives, the pair never puts the same field in both halves, and calling it with nothing does not throw. 8 passing.

Cookiezaurs (approved):

PR #7923, opened 2026-08-12
[fix][core] keep segmentation values out of Object.prototype

ar2rsawseen:
## What was wrong

An event segmentation **value** on the unauthenticated `/i` endpoint becomes a MongoDB field name at `d.<day>.<value>.<metric>`. A value of `__proto__` passed every transform — no `$`, no `.`, not one of the day numbers in `forbiddenSegValues` — and stored a document with a literal `__proto__` field.

On read, `deepMerge` in `fetch.js` walks the document with `for...in`. The driver returns `__proto__` as an own enumerable property, so the loop yields it, and:

```js
else if (ob1[i] && typeof ob1[i] === "object") {
    ob1[i] = deepMerge(ob1[i], ob2[i]);   // ob1["__proto__"] is Object.prototype
}
```

merges the stored object into `Object.prototype` of the worker. That persists for the worker's whole life, so every later read, for every app it serves, returns corrupted numbers. The daily `api:topEvents` job triggers the read across all apps unprompted.

`#7634` guarded the segmentation **key**. This is the **value** path four lines down, which that guard never reached, plus two more sites of the same class and the sink itself.

## What changed

- `common.isForbiddenFieldName` — one predicate for `__proto__` / `constructor` / `prototype`, now also used by the key guard `#7634` hard-coded, so the whole class reads through one function.
- `events.js` — the segmentation value is prefixed with `[CLY]` when it names a prototype member, exactly as forbidden day numbers already are.
- `common.js` `recordSegmentMetric` — the same for the metric value, which builds an identical `d.<...>` field name.
- `fetch.js` `deepMerge` — skips inherited keys (`hasOwnProperty`) and the three names. This is the durable guard: it also neutralises documents already poisoned in existing databases, which the input guards alone cannot, since those re-pollute on every read.

## Verified against the unpatched code

The real `deepMerge` was lifted and fed a document with an own-enumerable `__proto__` field, the way the mongo driver deserializes one:

```
before (pristine):  ({}).c = 100   ({}).s = 500   -> Object.prototype polluted process-wide
after  (patched):   ({}).c = undefined            -> clean, and a legitimate merge still sums (Chrome.c = 6)
```

Unit test `test/unit-tests/api.data.segmentation-value-prototype.js`: the exported predicate, plus source-level assertions that the two value paths run through it and that `deepMerge` skips inherited and prototype-member keys. 5 cases, all failing against the unpatched files.

## Scope

The corruption is in the running process, not the stored data — the database is never wrong, and a poisoned document re-pollutes each fresh worker on read until the code is fixed. Property names are confined to the `common.dbMap` metrics and the values are numbers, so this is data-integrity corruption, not RCE.

## Second sink, found after the first commit

`getMergedEventData` in the same file walks stored event documents through five nested `for...in` levels and merges with `mergedEventOutput[l1][l2][l3][l4][l5] += …`. Same defect, same reachability: a segmentation value is a key at those levels, and each level's guard is `if (!mergedEventOutput[…])`, which a prototype is truthy for, so it is never replaced before the assignment writes through it.

Fixing `deepMerge` alone therefore left the read path exploitable from any already-poisoned document, which defeats the point of the sink fix.

Worth recording why it was missed: it is not named like a merge helper and it does not recurse, so both a search for merge functions and a search for recursive walkers skip it. Only a search for nested bracket assignment finds it.

- `isMergeableKey` is now the single guard for every walk of a stored document in this file, and `deepMerge` uses it too.
- All five levels guarded, plus the `meta` reduce, where a prototype key would call `.concat` on `Object.prototype` and throw rather than pollute.

Proof, on all three targets:

```
before:  ({}).c = 7   ({}).s = 77   -> polluted process-wide from one merged document
after:   ({}).c = undefined         -> clean, Chrome.c still 3+3 = 6
```

The harness runs the lifted loop in the host realm rather than a `vm` context. A vm has its own `Object.prototype`, so pollution inside it is invisible to a probe outside and the test passes either way; my first attempt at this made exactly that mistake.

---

## Client / dashboard side (added in this PR)

The same class exists in the dashboard: a segmentation value, event key or metric name becomes an object key in the browser too, and an API response delivered as JSON can carry `__proto__` as an own enumerable property. Two hand-rolled merges in `countly.common.js` write *through* such a key and reach `Object.prototype` for the life of the page.

- `countlyCommon.isForbiddenFieldName` — the browser mirror of the server predicate.
- `mergeMetricsByName` — pollutes through a data **value** index (`uniqueNames[newName][k]`, `newName` a segmentation/range value); the value is now skipped before use.
- `extendDbObj` — walks a fetched day document with `for (var level1 in …)`; the existing `hasOwnProperty` does not exclude an own `__proto__`, so both levels now skip the prototype names.
- The seven segmentation/response `for…in` walks in `countly.event.js` (readers / write-AT loops), guarded for consistency.

Proof, real functions lifted out of the browser IIFE and fed an own-`__proto__` payload:

```
before:  ({}).mmbn_marker = "PWN"   ({}).edo_marker = 7   -> Object.prototype polluted page-wide
after:   both undefined                                   -> clean, and Chrome still sums to 5
```

## The eslint rule, extended to the dashboard

`no-prototype-pollution-sink` was scoped to `api/**` + `plugins/*/api/**`. It is now also switched on for `frontend/express/public/javascripts/countly/**` and `plugins/*/frontend/**`, and every site it reports there is guarded in source — keeping the no-exceptions contract. It surfaced two genuinely global sinks the manual review had not named:

- **plugins/views `countly.models.js`** — `for (var k in json)` over an API response, writing through `_graphDataObj[k][z]`.
- **plugins/sources `countly.views.js`** — an index through a `source` name *derived from* the response; fixed by skipping the prototype names and writing onto an already-selected bucket, so the derived value no longer sits mid-chain in the assignment target.

The remaining reports (auth app / permission maps, the session sparkline walk, the two `_myRequests` walks, the sdk config and push emoji loops) iterate objects the code builds itself, so the guard is a cheap no-op that keeps the rule clean. eslint reports **0** across api + plugins-api + frontend + plugins-frontend.

Proof, the two global plugin sinks:

```
before:  ({}).pollViews = {x:1}   ({}).PWN = {...}   -> polluted page-wide
after:   both undefined                              -> clean, real view/source still processed
```

Unit test `test/unit-tests/frontend.segmentation-value-prototype.js`: the predicate, the two lifted `common.js` sinks and two lifted `event.js` functions run on an own-`__proto__` payload, plus guard-presence and eslint-scope assertions and a behavioural proof for the two plugin sinks. 21 cases.


api/parts/data/fetch.js
```diff
@@ -214,6 +214,21 @@ fetch.fetchMergedEventData = function(params) {
...
+* deserializes as an own enumerable property, but indexing the target with it
+* resolves to the target's prototype rather than an own slot, so assigning through
+* it writes into Object.prototype for the life of the worker. Inherited keys are
+* skipped for the same reason.
+* @param {object} source - the object being iterated
+* @param {string} key - the key for...in produced
+* @returns {boolean} true when the key is safe to merge
+**/
+function isMergeableKey(source, key) {
+    return Object.prototype.hasOwnProperty.call(source, key) && !common.isForbiddenFieldName(key);
```
ar2rsawseen, line 229: [P2] Preserve legal event keys at non-segmentation levels

`isMergeableKey()` is used at every recursive level, so this drops any field named `prototype` or `constructor`, even where that field is a legitimate custom-event key rather than an unsafe segmentation value. Those names pass event-key validation and can be stored under daily event statistics; after this change their totals disappear depending on which shard is merged. Please encode these names or restrict the forbidden-name check to positions that are actually used as dynamic accumulator keys, instead of discarding them globally.

api/utils/common.js
```diff
@@ -3297,6 +3326,11 @@ common.mergeQuery = function(ob1, ob2) {
             }
             else if (key === "$push") {
                 for (let val in ob2[key]) {
+                    // a key off a stored document or a parsed payload can be a prototype
+                    // member name; writing through one would reach Object.prototype
+                    if (val === "__proto__" || val === "constructor" || val === "prototype") {
```
ar2rsawseen: [P2] Apply the same forbidden-name guard to `$addToSet`

The immediately preceding `$addToSet` merge indexes through the same attacker/data-derived `val` without this guard. For `val === "__proto__"`, `ob1[key][val]` resolves to `Object.prototype`, the initializer is skipped because it is already an object, and the subsequent `. $each.indexOf(...)` access throws because `$each` is absent. A crafted update can therefore still crash the merge/request path. Guard every nested operator loop consistently (or use null-prototype accumulators), not only `$push`.

ar2rsawseen: Confirmed, and it went further than `$addToSet` — fixed in 4e21957711e.

**The crash reproduces exactly as you describe.** Running the unguarded loop against a parsed payload:

```
TypeError: Cannot read properties of undefined (reading 'indexOf')
```

`ob1.$addToSet.__proto__` is `Object.prototype`, `typeof` it is `"object"`, so the *"create as object if it is single value"* initializer never fires, and `.$each` is `undefined` on the next line.

**One detail worth recording, because it decides whether a test for this is real.** The payload has to be *parsed*. In an object literal, `"__proto__": x` is the prototype-setting form and creates no own key at all, so a literal-built fixture iterates zero keys and passes whether or not the guard exists. `JSON.parse` and the BSON decoder both produce a genuine own enumerable `__proto__` — which is what an update assembled from a request, or read back from a document, actually looks like. The new tests build their fixtures with `JSON.parse` and assert the own-key precondition first.

**On "guard every nested operator loop consistently":** `$addToSet` was not the only gap — `$set`/`$setOnInsert`/`$unset`, `$inc`, `$mul`, `$min`, `$max` and the `$push` modifier loop all index `ob1[key][val]` with the same data-derived name. Rather than repeat the three-name literal nine times, the keys now go through one helper:

```js
function mergeableKeys(source) {
    if (!source || typeof source !== "object") {
        return [];
    }
    return Object.keys(source).filter(function(key) {
        return !common.isForbiddenFieldName(key);
    });
}
```

and every loop became `for (let x of mergeableKeys(...))`. `Object.keys` also drops **inherited** keys, which `for...in` did not — so an update object built on an already-poisoned prototype no longer contributes phantom operators either. The two inline guards the PR had are gone, replaced by the uniform one.

Six behavioural tests cover it, and four of them fail against the previous state of this branch.

api/parts/data/fetch.js
ar2rsawseen: You are right that the key is dropped, and right that the fix belongs on the write side. Addressed in 4e21957711e — though checking it turned up that the write path was not just losing data, it was still the pollution source.

**`fixEventKey` passes `__proto__` through unchanged.** It strips `system.`, `..` and `$`; none of the three names contain any of those. The key then becomes a stored field name at `d.<day>.<key>.<metric>` on the totals document (`fillTimeObjectMonth(params, tmpTotalObj, shortEventName + '.' + common.dbMap.count, …)`), so an SDK sending an event named `__proto__` writes a document that the read path then has to defend against — **on every ingest**. The read-side guard was the only thing standing between that and `Object.prototype`.

So the key now gets the same treatment a forbidden segmentation value already gets in this PR:

```js
if (common.isForbiddenFieldName(shortEventName)) {
    return "[CLY]" + shortEventName;
}
```

Inside `fixEventKey` rather than at a call site, because every consumer derives from that one return value — the per-event collection hash `sha1(shortEventName + app_id)`, the stored paths, and the short keys that drill, data-manager and journey_engine compute. Prefixing at the chokepoint keeps them all agreeing.

**What that does to your data-loss case:** an event legitimately named `prototype` is stored as `[CLY]prototype` from here on and merges normally — nothing is dropped. Only documents written *before* the upgrade still lose that key at merge, and those are precisely the documents that re-pollute on every read, so the read-side guard has to stay for them regardless.

**On "restrict the check to positions actually used as dynamic accumulator keys":** in these loops that is every position. Each level writes `mergedEventOutput[levelOne][levelTwo]…` / `mergedDataObj[year][month][prop]` — the key from the stored document *is* the accumulator index at all five levels, which is why the guard is applied uniformly rather than at a chosen depth.

One thing I did **not** change, for the record: `prototype` is the only one of the three that is inert against a plain-object accumulator (`({}).prototype` is `undefined`, so the initializer fires and an own slot is created — unlike `constructor`, which resolves to the `Object` function). Keeping it in the list is defence-in-depth and, more importantly, keeps the write-side prefix and the read-side skip describing the same set of names; splitting them would make the guard much harder to reason about later.

test/unit-tests/frontend.segmentation-value-prototype.js
```diff
@@ -0,0 +1,310 @@
...
+    for (var j = start + 1; j < lines.length; j++) {
+        if (lines[j] === endLine) {
+            return lines.slice(start, j + 1).join("\n");
+        }
+    }
+    throw new Error("end not found for: " + startPrefix);
+}
+
+// minimal stand-ins for the browser globals the lifted functions touch
+var moment = function() {
```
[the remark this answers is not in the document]
ar2rsawseen: These two are **not** unused — deleting them breaks the test.

The suite lifts the guarded functions out of the shipping `countly.common.js` / `countly.event.js` and runs them through `eval()` **in this scope**, so the lifted bodies resolve `moment`, `jQuery` and `_` against these stand-ins. CodeQL cannot see through `eval`, so it reports the definitions as dead. Removing them turns the alert into a `ReferenceError` from inside an eval'd function body.

Kept them, and made the dependency an explicit read so the signal goes away for the right reason (`4e21957711e`):

```js
// moment, jQuery and _ above are read by the lifted source at eval() time, from this
// scope. Static analysis cannot see through eval and reports them as unused, so assert
// they are present: the reads are real, and a missing stand-in should say so here rather
// than as a ReferenceError from inside an eval'd function body.
[moment, jQuery, _].forEach(function(standIn, at) {
    if (!standIn) {
        throw new Error("browser stand-in " + at + " is missing");
    }
});
```

That is a genuine runtime check rather than a suppression, and it documents why the stand-ins exist. Suite still green: 20 passing.

test/unit-tests/frontend.segmentation-value-prototype.js
```diff
@@ -0,0 +1,310 @@
...
+};
+var _ = {
+    isObject: function(o) {
+        return o !== null && typeof o === "object";
+    },
+    values: function(o) {
+        return Object.values(o);
+    }
+};
+var jQuery = { i18n: { map: { "common.unknown": "Unknown" } } };
```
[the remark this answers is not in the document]
ar2rsawseen: These two are **not** unused — deleting them breaks the test.

The suite lifts the guarded functions out of the shipping `countly.common.js` / `countly.event.js` and runs them through `eval()` **in this scope**, so the lifted bodies resolve `moment`, `jQuery` and `_` against these stand-ins. CodeQL cannot see through `eval`, so it reports the definitions as dead. Removing them turns the alert into a `ReferenceError` from inside an eval'd function body.

Kept them, and made the dependency an explicit read so the signal goes away for the right reason (`4e21957711e`):

```js
// moment, jQuery and _ above are read by the lifted source at eval() time, from this
// scope. Static analysis cannot see through eval and reports them as unused, so assert
// they are present: the reads are real, and a missing stand-in should say so here rather
// than as a ReferenceError from inside an eval'd function body.
[moment, jQuery, _].forEach(function(standIn, at) {
    if (!standIn) {
        throw new Error("browser stand-in " + at + " is missing");
    }
});
```

That is a genuine runtime check rather than a suppression, and it documents why the stand-ins exist. Suite still green: 20 passing.

Cookiezaurs (approved):

PR #7926, opened 2026-08-13
[fix][core] keep server side rendering on the dashboard origin

ar2rsawseen:
## What

`api/utils/render.js` builds the url it opens by concatenating the configured dashboard host with the requested view:

```js
var host = (process.env.COUNTLY_CONFIG_PROTOCOL || "http") + "://" + (process.env.COUNTLY_CONFIG_HOSTNAME || "localhost") + countlyConfig.path;
...
await page.goto(host + view);
```

`/o/render` takes that view from the request (`params.qstring.view`). With the default `countlyConfig.path` of `""` the prefix is exactly `http://localhost`, so a view that does not begin with `/` changes the host rather than the path:

| view | url actually opened |
| --- | --- |
| `/#/dashboard` | `http://localhost/#/dashboard`, as intended |
| `@169.254.169.254/latest/...` | host `169.254.169.254` |
| `:8500/v1/kv/?recurse` | host `localhost:8500` |
| `.internal.example/x` | host `localhost.internal.example` |

The headless browser then loads that from the server's own network position.

## Change

1. Parse the concatenated url and require its origin to match the dashboard's before navigating. What gets opened is the concatenation itself, unchanged, so a configured `countlyConfig.path` keeps behaving exactly as it did. Parsing settles the whole family of shapes at once, including the ones the url parser normalises away (tab, newline, carriage return), and it agrees with what Chromium does with the same string since both use the WHATWG parser.
2. Restrict the renderer to the dashboard origin with request interception, the same control `api/utils/pdf.js` uses, so the page's subresources are bounded too and not just the navigation.

A private range denylist would be the wrong control here: the intended render target is loopback, which `api/utils/ssrf-protection.js` blocks by design.

## Scope

Everything that drives a headless browser, with a verdict for each:

| site | verdict |
| --- | --- |
| `api/utils/render.js`, `page.goto(host + view)` | fixed here |
| `api/utils/render.js`, login token navigation | token is server generated and the host comes from config, and it is covered by the interception anyway |
| `api/utils/pdf.js` | already limited to Countly's own origins |
| `api/utils/requestProcessor.js`, the `/o/render` caller | passes `params.qstring.view` through, now checked at the sink |
| `plugins/dashboards/api/api.js` caller | hardcodes `"/dashboard?ssr=true#/custom/" + id`, unaffected and covered by the tests below |
| countly-enterprise-plugins | no `render.js` and no `renderView` caller |

Neither caller sets `options.host`, so the base always stays the configured dashboard.

## Verification

- `test/unit-tests/api.utils.render.js`, 38 cases: every host rewriting shape is refused for each host form, and every view the product itself renders is allowed and returned byte identical. Swapping the check back for plain concatenation fails 13 of them, so the tests do pin the behaviour.
- Real Chromium measurement of the interception, since the flag removal in 405b98212c9 showed this file is easy to reason about wrongly: own origin image and stylesheet still load, the screenshot is still produced, the off origin request never reaches the other server, and no unhandled rejections are raised.
- countly-platform's existing live `renderView` and `renderPDF` integration suite (`test/2.api/99.api.utils.render.js`) passes against the patched file, 7 of 7, including both real screenshot cases.
- Grepped the dashboard shell for external subresources: there are none, the only CDN references in the tree are the populator demo templates, so nothing in a normal render is refused by the origin restriction.


api/utils/render.js
```diff
@@ -130,6 +178,38 @@ exports.renderView = function(options, cb) {
...
+                        return request.continue();
+                    }
+                    var requestOrigin;
+                    try {
+                        requestOrigin = new URL(requestUrl).origin;
+                    }
+                    catch (error) {
+                        requestOrigin = null;
+                    }
+                    if (renderOrigin && renderOrigin !== "null" && requestOrigin === renderOrigin) {
```
ar2rsawseen: [P2] Allow explicitly trusted off-origin dashboard assets

Restricting every subresource to the dashboard origin breaks legitimate rendered views. Deployments using `countlyConfig.cdn` load core scripts/styles from that configured origin, and map widgets load OpenStreetMap tiles from their provider; this handler aborts both, leaving screenshots/PDFs unstyled, blank, or unable to initialize. Keep the navigation same-origin check, but add a narrow allowlist for configured CDN and required renderer asset origins (or proxy those assets through the dashboard origin).

ar2rsawseen: Confirmed — both examples check out against the tree, and both are fixed in 5269647ef87.

| what | where | what the render did |
|---|---|---|
| `countlyConfig.cdn` | `frontend/express/views/dashboard.html:84-100` prefixes **every** core stylesheet and script | a CDN deployment loses its bootstrap assets entirely and the render times out waiting for the dashboard to initialize |
| OSM tiles | `vis.js:3001` / `:3264`, `https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png` | map widgets render blank |

The `{s}` detail matters for the fix: `vis.js` passes no `subdomains` option, so leaflet expands it to `a`, `b`, `c` — three origins, plus the bare host.

**The check is now against an allowlist rather than a single origin:**

```js
function renderAssetOrigins(host) {
    var dashboard = originOf(host + "/");
    if (!dashboard) {
        return [];
    }
    …
    var origins = [dashboard, cdnOrigin()]
        .concat(TILE_ORIGINS)
        .concat(configured.map(originOf));
    return origins.filter(function(origin, at) {
        return origin && origins.indexOf(origin) === at;
    });
}
```

with `countlyConfig.render.allowedOrigins` for anything else a deployment needs — a self-hosted tile server, a second asset host. Entries may be a bare origin or any URL on one. Documented in `api/config.sample.js`.

**On the security side, taking your "keep the navigation origin restriction" literally:** `sameOriginView` is untouched. That is the control that stops the renderer reaching what the server can reach, and it is independent of this one — allowing an asset origin does not widen it.

Three details worth flagging:

- **The CDN is read from the dashboard's config, not the API's.** In the standard layout both are in the same tree and it is picked up with no operator action. Where the API runs without the dashboard the file is absent, the `require` fails closed, and the origin goes in the config array instead.
- **Tile origins are four exact strings, not a host-suffix match.** The allowlist stays a set of strings to compare with no pattern to get wrong — `tile.openstreetmap.org.evil.example` is a different origin and stays out. There is a test for exactly that.
- **An unparseable dashboard host now yields an empty allowlist**, which refuses everything, rather than the previous shape where a null `renderOrigin` was checked separately. Empty must never degrade into "allow anything".

I did not take the proxy alternative: proxying tiles through the dashboard origin would put the server in the path of every map fetch, which is a bigger change and a worse one.

Ten tests added, plus — separately — the platform PR had **no** render test file at all, so it now carries the same suite the server one has (47 passing on all three).

One correction while I was in there: a comment claimed this mirrors `api/utils/pdf.js`. It does not — `pdf.js` renders supplied HTML and intercepts nothing. Comment fixed.

Cookiezaurs (approved):

PR #7929, opened 2026-08-14
[core] Fix long report/funnel names overlapping header action buttons

mrmeghana:
## Problem
On report detail pages (e.g. a funnel), a very long title/name displayed in the shared `cly-header` **overlaps the action buttons on the right** (Edit / more-options). Because `cly-header` is reused across reports, the issue appears in all of them.

## Root cause
The header's mid-level row is a flex row: a shrinkable left slot (`bu-is-flex-shrink-1` + inline `min-width:0`) next to the right-side buttons. The left slot had **no `overflow` handling**, so overflowing content bled over the right slot instead of being clipped, and nothing pinned the buttons.

## Fix
In `_header.scss`, for `.cly-vue-header__level`:
- clip the left slot's overflow so long content stays in its column;
- pin the right slot (`flex-shrink: 0`) with a 16px gap so the action buttons keep their natural width and position;
- ellipsize the default `<h2>` title.

This stops the overlap in every report using `cly-header`. Plugins with a custom `header-left` (e.g. funnels) additionally get proper ellipsis via their own small change.

## Testing
Verified the compiled rules land in `manifest.css` and confirmed the flex/truncation behavior in an isolated reproduction of the header DOM (long name → ellipsis, buttons stay put; short name unchanged).


mrmeghana:
Closing: per review on Countly/countly-platform#1072, the fix is being scoped to a funnel-specific class instead of the shared `cly-header` component. Since funnels is an enterprise plugin (not in this repo), there's no shared change to land here. The funnel fix lives in Countly/countly-enterprise-plugins#3367.

PR #7930, opened 2026-08-14
[fix][star-rating] keep the uploaded logo name inside the images directory

ar2rsawseen:
Two problems with the same request parameter on `/i/feedback/logo`: where the file lands, and whose file it replaces.

## 1. The identifier chose the write location

`uploadFile()` concatenated the identifier into the upload path:

```js
var pp = path.resolve(__dirname, './../images/' + id + "." + detectedExt);
```

`id` is `params.qstring.identifier`, so separators or leading dots in it selected a directory rather than only a file name. With `fileStorage` set to `"fs"` the destination reaches `fs.writeFile` unchanged, so the write could land in any directory that already exists. Under the shipped default of `"gridfs"` the same call keeps only the basename, so the path never reaches disk.

**Fix:** validate the identifier in `image-utils.js`, next to the existing `parseFeedbackLogoName`, and use the validated value for the path, the stored id and the returned name. The dashboard sends `Date.now()`, so accepting a plain name covers every real upload.

## 2. The identifier also chose whose logo was replaced

The route is authorized with `validateCreate` against the caller's own `app_id`, but every app's logos share one directory and a widget's `logo` field holds just the file name. So the identifier alone decided which app's logo was overwritten, and an account with create rights on its own app could replace the logo shown in another app's end-user feedback popup. The name is not private: it comes back with the widget from the sdk facing by-id lookups.

**Fix:** refuse a name that a widget belonging to a different app already points at.

The sibling `/i/feedback/upload` route already does this, decoding its target app out of the file name with a comment saying an admin of one app must not be able to plant a logo for another. This brings the older route into line.

Two details worth stating:

- Matching on the full name including the extension is deliberate. A different extension is a different file and overwrites nothing.
- Re-uploading your own app's logo is unaffected, because the check only considers widgets whose app differs from the caller's.

## Scope

Every place a request supplied value reaches a file write, with a verdict:

| site | verdict |
| --- | --- |
| `plugins/star-rating/api/api.js` `uploadFile` (`/i/feedback/logo`) | both problems fixed here |
| `plugins/star-rating/api/api.js` `uploadFeedbackFile` (`/i/feedback/upload`) | already validated, and already app scoped by name |
| `api/parts/mgmt/apps.js` `iconUpload` | already sanitized via `common.sanitizeFilename` |
| `api/utils/render.js` screenshot path | name is server generated |
| countly-enterprise-plugins `crash_symbolication` | `symbol_id` passes `common.db.ObjectID()` and a `findOne` match before the write |

`uploadFile` has exactly one caller, so both checks cover the whole reachable surface.

## Verification

- `test/unit-tests/star-rating.image-utils.js`, 29 cases: traversal, separators in both slash directions, absolute and drive style paths, NUL and tab truncation, plus the numeric identifiers the dashboard sends and plain names with dashes and underscores. Reverting the validator to the previous pass through fails four of them.
- The cross-app check is a single query against `feedback_widgets` and is not covered by an automated case: the plugin suite has one application, so a second one would have to be created first. Verified by reading the query and its two branches instead, and noted here rather than implied.
- Lint and `node --check` clean on the changed files.


plugins/star-rating/api/api.js
```diff
@@ -284,21 +296,49 @@ function uploadFile(myfile, id, callback) {
...
-                countlyFs.saveData("star-rating", pp, data, { id: "" + id + "." + detectedExt, writeMode: "overwrite" }, function(err3) {
+            //The images directory is shared by every app and a widget's logo field holds
+            //just this file name, so the name alone decides whose logo is replaced. This
+            //request was authorized against the caller's own app, so a name that another
+            //app's widget already points at is not ours to overwrite. The sibling
+            ///i/feedback/upload route decodes its target app out of the name for the same
+            //reason. Matching on the full name including the extension is deliberate: a
+            //different extension is a different file and overwrites nothing.
+            var storedName = safeId + "." + detectedExt;
+            common.db.collection('feedback_widgets').findOne({logo: storedName, app_id: {$ne: appId + ""}}, {projection: {_id: 1}}, function(ownerErr, otherAppWidget) {
```
ar2rsawseen, line 362: [P2] Make cross-app logo ownership atomic

This is a read-then-write check against a filesystem name shared by all apps. Two applications uploading the same attacker-chosen identifier concurrently can both observe no conflicting widget and then both call `saveData`; the later write overwrites the earlier app's logo, so the cross-app protection is bypassed by a race. Namespace stored filenames by `appId`, or reserve the name with an atomic unique database record/index before writing.

ar2rsawseen: Confirmed, and taken the first of your two options — the name is namespaced rather than reserved. Fixed in 240d37e3d1a.

Worth recording that the window is **wider than a race**, which is why I did not reach for a unique index. The upload happens *before* the widget that will point at the file exists, so at check time there is frequently nothing to check against at all:

```
app A: POST /i/feedback/logo?identifier=x   -> no widget has "x.png" -> writes x.png
app B: POST /i/feedback/logo?identifier=x   -> still no widget has "x.png" -> writes x.png
app A: creates its widget with logo = "x.png"   (now pointing at B's file)
```

No interleaving needed — minutes apart is enough. A reservation record would have to be created at upload and reconciled when the widget is saved or abandoned, which is a lot of moving parts to keep a shared namespace that has no reason to be shared.

So the stored name is now:

```
<app id>_<identifier>.<extension>
```

built in `imageUtils.logoStorageName`. The same shape the sibling `/i/feedback/upload` route already uses, where the target app is likewise carried by the name rather than taken from the query string. Two apps cannot produce the same name, so there is nothing left to contend for.

Two things checked along the way, since both end up in a path:

- **The app id** is validated as an ObjectID before it is concatenated. It arrives from the query string; `validateCreate` proves the caller may create for it, not that it is shaped like an id.
- **The extension** is validated too. It comes from sniffing the file today and never from the request — this keeps that true if a caller is ever added.

**I kept the widget lookup** rather than deleting it along with the problem it was solving. A logo stored before this change is a bare `<identifier>.<extension>`, and `LOGO_IDENTIFIER_RE` allows `_`, so an identifier crafted to look like `<other app id>_<name>` can still land on a legacy file. That is the only remaining way to collide, and the check covers it.

**No migration needed:** existing widgets keep their stored names and their files, and the endpoint returns the name the dashboard writes into `widget.logo`, so old and new names coexist.

Tests cover the point directly — two apps get different names for the same identifier — plus the app-id and extension rejections. The platform PR had no test file for `image-utils` at all, so it now carries the `safeLogoIdentifier` cases as well.

Cookiezaurs (approved):

PR #7932, opened 2026-08-14
[fix][core] keep the stored member language a language code

ar2rsawseen:
## What

The member language is stored by `/user/settings/lang` in `frontend/express/app.js` with no validation:

```js
if (req.body.lang) {
    updatedUser.lang = req.body.lang;
    countlyDb.collection('members').update({"_id": ...}, {'$set': updatedUser}, ...);
```

That stored value then reaches file paths in around 25 readers. `api/utils/localization.js` sanitizes it in both `getProperties` and `getProperty`, but the two plugin localization lookups in `plugins/plugins/api/api.js` concatenate it as it comes:

```js
local_path = fullpath + "/frontend/public/localization/" + resultObj.code + "_" + params.member.lang + ".properties";
```

## Change

Two independent layers:

1. **Validate where it is stored**, so every reader inherits it. A language code is accepted, anything else is refused and the request answers false as it already does for a missing value.
2. **Sanitize the two readers that build a path by hand**, the same way `localization.js` already does.

A language code pattern rather than an allow list drawn from `locale.conf.js`, so a deployment shipping its own localization file keeps working, while leaving no way for path syntax to reach storage.

## Why both layers

The value has around 25 readers across six files. Everything except these two goes through the sanitizing localization helpers or `moment.locale`, which is not a path. Fixing only the readers leaves the next reader exposed; fixing only the source leaves rows already stored. Doing both closes it now and keeps it closed.

## Scope

Every reader of the stored language, with a verdict:

| reader | verdict |
| --- | --- |
| `api/utils/localization.js` `getProperties`, `getProperty` | already sanitized |
| `plugins/plugins/api/api.js` plugin localization lookup | sanitized here |
| `plugins/plugins/api/api.js` plugins localization lookup | sanitized here |
| `plugins/reports/api/utils.js` `getLocaleLangString` | goes through `locale.getProperty`, sanitized |
| `plugins/reports/api/reports.js` `moment.locale(lang)` | not a path |
| `api/parts/mgmt/mail.js`, `plugins/dashboards`, `plugins/star-rating` | all via the localization helpers |

## Verification

The accepted pattern against real and hostile values:

- accepted: `en`, `tr`, `de`, `zh`, `pt`, `pt-br`, `zh_CN`, `es`
- refused: `../../../../etc/passwd`, `..`, `/tmp/x`, `en/../../x`, `en%2f..`, empty, `.`, an 80 character name, `en;rm -rf /`

Lint clean on both changed files in each repo. No behaviour change for the dashboard, which only sends codes from its own list.


frontend/express/app.js
```diff
@@ -75,6 +75,10 @@ var versionInfo = require('./version.info'),
     { validateCreate } = require('../../api/utils/rights.js'),
     tracker = require('../../api/parts/mgmt/tracker.js');
 
+//Language codes as the dashboard uses them, e.g. "en", "pt-br", "zh_CN". Anything
+//else is refused rather than sanitized, so a bad value never reaches storage.
+var LANG_CODE_RE = /^[A-Za-z]{2,3}(?:[-_][A-Za-z0-9]{2,8})?$/;
```
ar2rsawseen: [P2] Accept valid locales with multiple subtags

This pattern permits only one optional subtag, so a deployment that extends the locale list with a valid code such as `zh-Hans-CN` can no longer save that selection even when the matching localization files exist. The dashboard still posts the configured locale, but this route silently ignores it and reloads with the old preference. Allow repeated safe `-`/`_` subtags, or validate membership in the server's configured localization list.

ar2rsawseen: Confirmed, and taken your second option — membership in the configured list — because it is the stronger of the two. Fixed in 29015af07c5.

A pattern has to *guess* which codes are legitimate, and guessing narrowly is exactly what produced this bug. The list is right there:

```js
languages = require('../../frontend/express/locale.conf'),
```

That is the same list `countlyGlobal.languages` is built from, and both callers post a code straight out of the menu built from it — `countly.template.js:1837` and `sidebar.js:539` each send `l.code` on a menu click, never a browser-derived value. So the allowlist accepts everything the UI can offer, including anything a plugin adds through `plugins.extendModule("locale.conf", …)`, and nothing else at all.

**The shape check stays**, now allowing repeated subtags, but as a second gate rather than the decision:

```js
function isSelectableLang(lang) {
    if (typeof lang !== "string" || !LANG_CODE_RE.test(lang)) {
        return false;
    }
    return Array.isArray(languages) && languages.some(function(locale) {
        return locale && locale.code === lang;
    });
}
```

`locale.conf.js` is deployment configuration, not something this route validated, and the stored value decides a file path — so a configured entry containing a traversal is still refused. There is a test for that direction specifically, alongside one that walks the whole shipped list and asserts every offered code is accepted.

**On the silent part of your report:** a rejection is now logged. Since the allowlist accepts everything the menu can send, the only way to reach that branch is a hand-made request, which is worth a line in the log.

---

One unrelated bug found in a line this PR already touches, fixed here rather than left:

```js
path = fullpath + "/frontend/public/localization/plugins" + "_" + … + ".properties";
```

in `plugins/plugins/api/api.js` assigns to `path` — which is `require('path')` at the top of that file. One call to `/o/email_test` by a non-English global admin left `path.resolve` broken for the rest of the process. It now uses a local name.

Cookiezaurs (approved):

PR #7935, opened 2026-08-14
[fix][views] apply the token owner's read right on the heatmap endpoint

ar2rsawseen:
## What

The `countly-token` branch of `/o/actions` resolves the target app from the caller supplied `app_key`:

```js
common.readBatcher.getOne("apps", {$or: [{'key': params.qstring.app_key + ""}, ...]}, (err1, app) => {
    params.qstring.app_id = app._id + "";
    authorize.verify_return({ ..., callback: function(owner, expires_after) {
        if (owner) {  // only used as a truthiness check
```

and then served the data as soon as the token verified. The `owner` it gets back was never used for anything: the member was not loaded and no read right was checked, so `getHeatmap` ran with `matchQuery.a` set to whatever app the `app_key` named.

Everywhere else a token acts as its owner. `verify_return` returns the owner, then `validateUser` / `validateRead` load that member and apply their real rights, while the token's `app`, `endpoint` and `ttl` fields only narrow it further. This branch skipped that step, which left the optional app restriction as the only thing bounding the read. A token saved without an app restriction is not narrowed at all, which is correct in itself, so nothing remained to bound it.

## Change

Load the token's owner and require a `views` read right on the app resolved from `app_key`, mirroring what `validateRead(params, FEATURE_NAME, getHeatmap)` already does for the `api_key` branch a few lines below.

## Compatibility

The heatmap feature has been out of the product for over two years and this endpoint stays only so a long-standing integration does not break, so the existing flow is preserved rather than tightened:

- The dashboard mints this token scoped to the active app: `createToken("View heatmap", "/o/actions", true, countlyCommon.ACTIVE_APP_ID, 1800, ...)` in `plugins/views/frontend/public/javascripts/countly.views.js`. Same call in every repo checked.
- Its owner is the member who opened the heatmap, who holds the read right for that app, so the new check passes for them.
- Nothing about token scoping changes, and `verify_token` is untouched: an empty app scope still means "not additionally restricted", which is correct on every path that applies the owner's rights.

## Verification

No automated test accompanies this change, and that is deliberate rather than an omission. `plugins/views/tests/heatmaps.js` is commented out of `plugins/views/tests/index.js`, so anything added there would not run, and it cannot simply be re-enabled here: the suite reads `plugins.drill` from `/o/apps/plugins` and connects to `countly_drill`, neither of which exists in this repo's CI. Since the feature is no longer supported, re-plumbing a suite for it is not worth the cost.

Checked by hand instead:

- The authorization predicate is `hasReadRight('views', app._id, member)`, the same helper `validateRead` uses for this feature, so the token branch and the `api_key` branch now apply the same rule.
- Both dispatch paths inside the branch, the token extension path and the direct one, are behind the check, since it wraps the whole `if (owner)` body.
- `lint`, `node --check`, and the rest of CI cover the change mechanically.


plugins/views/api/api.js
```diff
@@ -1584,31 +1584,44 @@ const escapedViewSegments = { "name": true, "segment": true, "height": true, "wi
...
-                                    });
+                                //A token acts as its owner: everywhere else it is resolved to the
+                                //member who created it and that member's own rights decide what the
+                                //request may read, while the token's app and endpoint fields only narrow
+                                //it further. This branch used to skip that step, so the app resolved from
+                                //the caller's app_key was served without consulting the owner at all.
+                                //Apply the owner's read right here, the way validateRead does for the
+                                //api_key branch below.
+                                common.db.collection('members').findOne({_id: common.db.ObjectID(owner + "")}, function(memberErr, member) {
+                                    if (memberErr || !member || !hasReadRight(FEATURE_NAME, app._id + "", member)) {
```
ar2rsawseen: [P2] Preserve read access for legacy token owners

`hasReadRight()` does not include the legacy `user_of` fallback that `validateRead()` still supports. A valid token owned by a legacy member who can read this app through `user_of` but has no modern `permission` object now receives 401, breaking existing heatmap integrations even though the equivalent API-key path remains allowed. Apply the same full read-validation semantics here, including the legacy membership fallback.

ar2rsawseen: Confirmed — `hasReadRight` has no `user_of` branch, and `validateRead` does. Fixed in 5214e34eabc.

Checked which of the two really diverge, since `hasReadRight` covers more than it looks like it does — through `hasAdminAccess(member, app_id, "r")` it already handles `permission.r[app_id].all`, `permission._.a` and legacy `admin_of`:

| granted by | `validateRead` | `hasReadRight` |
|---|---|---|
| `permission.r[app].allowed[feature]` | ✅ | ✅ |
| `permission.r[app].all` | ✅ | ✅ (via `hasAdminAccess`) |
| `permission._.a` contains app | ✅ | ✅ (via `hasAdminAccess`) |
| legacy `admin_of` | ✅ | ✅ (via `hasAdminAccess`) |
| **legacy `user_of`, no `permission` at all** | ✅ | ❌ |
| **`member.locked`** | refused | not checked |

So `user_of` is the only right that was lost — exactly your case — and the `locked` check was a second gap: a locked member's token kept working, while their api_key did not.

The branch now calls `viewsUtils.ownerCanRead`:

```js
ownerCanRead: function(member, appId, feature) {
    if (!member || member.locked) {
        return false;
    }
    if (member.global_admin) {
        return true;
    }
    if (typeof member.permission === "undefined") {
        return Array.isArray(member.user_of) && member.user_of.indexOf(appId) !== -1;
    }
    return !!hasReadRight(feature, appId, member);
}
```

`user_of` is consulted only when `permission` is undefined, mirroring `validateRead` — a migrated member is governed by the permission object alone, so an old `user_of` array left on the document does not quietly re-grant anything. There is a test for that direction too.

It sits in `viewsUtils` rather than inline in `api.js` so it can be tested without standing up a request: 11 cases across both member shapes.

test/unit-tests/plugins.views.token-owner-rights.js
```diff
@@ -0,0 +1,71 @@
+var should = require("should");
```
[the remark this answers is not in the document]
ar2rsawseen: Correct — the suite uses the `value.should.equal(...)` style, so `should` is needed for what it does to `Object.prototype` and never through the local name. Changed to `require("should");`, matching the other suites in this directory. 11 passing, unchanged.

Cookiezaurs (approved):

PR #7936, opened 2026-08-14
[fix][views] apply the token owner's read right on the heatmap endpoint (24.05)

ar2rsawseen:
Backport of https://github.com/Countly/countly-server/pull/7935 to release.24.05.

No test is added here: this branch has no `plugins/views/tests/heatmaps.js`, the file is newer than 24.05.

## What

The `countly-token` branch of `/o/actions` resolves the target app from the caller supplied `app_key`:

```js
common.readBatcher.getOne("apps", {$or: [{'key': params.qstring.app_key + ""}, ...]}, (err1, app) => {
    params.qstring.app_id = app._id + "";
    authorize.verify_return({ ..., callback: function(owner, expires_after) {
        if (owner) {  // only used as a truthiness check
```

and then served the data as soon as the token verified. The `owner` it gets back was never used for anything: the member was not loaded and no read right was checked, so `getHeatmap` ran with `matchQuery.a` set to whatever app the `app_key` named.

Everywhere else a token acts as its owner. `verify_return` returns the owner, then `validateUser` / `validateRead` load that member and apply their real rights, while the token's `app`, `endpoint` and `ttl` fields only narrow it further. This branch skipped that step, which left the optional app restriction as the only thing bounding the read. A token saved without an app restriction is not narrowed at all, which is correct in itself, so nothing remained to bound it.

## Change

Load the token's owner and require a `views` read right on the app resolved from `app_key`, mirroring what `validateRead(params, FEATURE_NAME, getHeatmap)` already does for the `api_key` branch a few lines below.

## Compatibility

The heatmap feature has been out of the product for over two years and this endpoint stays only so a long-standing integration does not break, so the existing flow is preserved rather than tightened:

- The dashboard mints this token scoped to the active app: `createToken("View heatmap", "/o/actions", true, countlyCommon.ACTIVE_APP_ID, 1800, ...)` in `plugins/views/frontend/public/javascripts/countly.views.js`. Same call in every repo checked.
- Its owner is the member who opened the heatmap, who holds the read right for that app, so the new check passes for them.
- Nothing about token scoping changes, and `verify_token` is untouched: an empty app scope still means "not additionally restricted", which is correct on every path that applies the owner's rights.

## Verification

- `plugins/views/tests/heatmaps.js` gains a case that exercises the `countly-token` branch end to end: mint a token the way the dashboard does, send it in the header, and assert the heatmap rows still come back. That branch had no test coverage at all before, the existing cases all use `api_key` and go through `validateRead`.
- `node --check` and eslint clean on the changed files.

### DIFF
--- a/plugins/views/api/api.js
+++ b/plugins/views/api/api.js
@@ -8,7 +8,7 @@ var pluginOb = {},
     plugins = require('../../pluginManager.js'),
     fetch = require('../../../api/parts/data/fetch.js'),
     log = common.log('views:api'),
-    { validateRead, validateUpdate, validateDelete, hasReadRight } = require('../../../api/utils/rights.js');
+    { validateRead, validateUpdate, validateDelete } = require('../../../api/utils/rights.js');
 
 const viewsUtils = require("./parts/viewsUtils.js");
 const FEATURE_NAME = 'views';
@@ -1554,7 +1554,7 @@ const escapedViewSegments = { "name": true, "segment": true, "height": true, "wi
                                 //Apply the owner's read right here, the way validateRead does for the
                                 //api_key branch below.
                                 common.db.collection('members').findOne({_id: common.db.ObjectID(owner + "")}, function(memberErr, member) {
-                                    if (memberErr || !member || !hasReadRight(FEATURE_NAME, app._id + "", member)) {
+                                    if (memberErr || !member || !viewsUtils.ownerCanRead(member, app._id + "", FEATURE_NAME)) {
                                         common.returnMessage(params, 401, 'User does not have view right for this application');
                                         return false;
                                     }
--- a/plugins/views/api/api.js
+++ b/plugins/views/api/api.js
@@ -8,7 +8,7 @@ var pluginOb = {},
     plugins = require('../../pluginManager.js'),
     fetch = require('../../../api/parts/data/fetch.js'),
     log = common.log('views:api'),
-    { validateRead, validateUpdate, validateDelete } = require('../../../api/utils/rights.js');
+    { validateRead, validateUpdate, validateDelete, hasReadRight } = require('../../../api/utils/rights.js');
 
 const viewsUtils = require("./parts/viewsUtils.js");
 const FEATURE_NAME = 'views';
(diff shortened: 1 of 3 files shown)


plugins/views/api/api.js
```diff
@@ -1546,31 +1546,44 @@ const escapedViewSegments = { "name": true, "segment": true, "height": true, "wi
...
-                                    });
+                                //A token acts as its owner: everywhere else it is resolved to the
+                                //member who created it and that member's own rights decide what the
+                                //request may read, while the token's app and endpoint fields only narrow
+                                //it further. This branch used to skip that step, so the app resolved from
+                                //the caller's app_key was served without consulting the owner at all.
+                                //Apply the owner's read right here, the way validateRead does for the
+                                //api_key branch below.
+                                common.db.collection('members').findOne({_id: common.db.ObjectID(owner + "")}, function(memberErr, member) {
+                                    if (memberErr || !member || !hasReadRight(FEATURE_NAME, app._id + "", member)) {
```
ar2rsawseen: [P2] Preserve read access for legacy token owners

`hasReadRight()` does not include the legacy `user_of` fallback that `validateRead()` still supports. A valid token owned by a legacy member who can read this app through `user_of` but has no modern `permission` object now receives 401, breaking existing heatmap integrations even though the equivalent API-key path remains allowed. Apply the same full read-validation semantics here, including the legacy membership fallback.

ar2rsawseen: Confirmed — `hasReadRight` has no `user_of` branch, and `validateRead` does. Fixed in 6708ad05983.

Checked which of the two really diverge, since `hasReadRight` covers more than it looks like it does — through `hasAdminAccess(member, app_id, "r")` it already handles `permission.r[app_id].all`, `permission._.a` and legacy `admin_of`:

| granted by | `validateRead` | `hasReadRight` |
|---|---|---|
| `permission.r[app].allowed[feature]` | ✅ | ✅ |
| `permission.r[app].all` | ✅ | ✅ (via `hasAdminAccess`) |
| `permission._.a` contains app | ✅ | ✅ (via `hasAdminAccess`) |
| legacy `admin_of` | ✅ | ✅ (via `hasAdminAccess`) |
| **legacy `user_of`, no `permission` at all** | ✅ | ❌ |
| **`member.locked`** | refused | not checked |

So `user_of` is the only right that was lost — exactly your case — and the `locked` check was a second gap: a locked member's token kept working, while their api_key did not.

The branch now calls `viewsUtils.ownerCanRead`:

```js
ownerCanRead: function(member, appId, feature) {
    if (!member || member.locked) {
        return false;
    }
    if (member.global_admin) {
        return true;
    }
    if (typeof member.permission === "undefined") {
        return Array.isArray(member.user_of) && member.user_of.indexOf(appId) !== -1;
    }
    return !!hasReadRight(feature, appId, member);
}
```

`user_of` is consulted only when `permission` is undefined, mirroring `validateRead` — a migrated member is governed by the permission object alone, so an old `user_of` array left on the document does not quietly re-grant anything. There is a test for that direction too.

It sits in `viewsUtils` rather than inline in `api.js` so it can be tested without standing up a request: 11 cases across both member shapes.

---

**One thing specific to this backport, which the master PR does not have.** Two of those tests fail on `release.24.05` as it stands, and not because of anything in this change: `hasAdminAccess` here still starts from `isAdmin = true` and only clears it when an entry for the app *exists* with `all` falsy. An app the member has no entry for never clears it, so `hasReadRight` returns **true** for every app they were never granted.

Applying the owner's read right on top of that would have been fail-open — the fix would have looked applied and let everything through.

`security/rights-admin-access-fail-open-2405` (#7873) fixes precisely that, so I stacked this branch on it, the same way the alerts backport is stacked. **#7873 has to merge first.** With it merged in, all 23 tests pass.

Cookiezaurs (approved):

PR #7938, opened 2026-08-14
[fix][core] validate scalar request parameter types before using them in queries (24.05)

ar2rsawseen:
Backport of #7937 to `release.24.05`.

## What

`params.qstring` values are not guaranteed to be strings. `api/api.js` fills `params.qstring` straight from formidable's `fields`, and a POST body sent as `application/json` puts nested objects there. A body like `{"view": {"$ne": null}}` leaves `params.qstring.view` an object rather than the string the endpoint expects. Form-urlencoded bracket syntax (`view[$ne]=x`) does not do this, it keeps a literal string key, so JSON bodies are the case that matters.

Several endpoints then use such a parameter as a plain value inside a Mongo query document. In a value position Mongo reads an object as a query expression, so an equality match on one document becomes a match on many. Two consequences: the query returns rows the endpoint never meant to return, and it loses the bound on how much data it has to scan.

`getHeatmap` is the expensive one. It builds one query from `view`, `actionType` and `segment` and runs it against the drill action collection over a caller-chosen `period`, so widening that match turns a single cheap request into a full scan of that collection. Note this differs from `master`, where the same function is an aggregation that also `$unionWith`s the older drill collection; the parameters and the fix are the same, the cost profile on 24.05 is a large `find` rather than a large aggregation.

## How

Adds `common.isQueryScalar(value)`: true for strings, numbers, booleans, `null` and `undefined`, false for objects and arrays. `null`/`undefined` stay scalars on purpose, so the existing truthiness checks at each call site keep deciding whether an absent parameter belongs in the query at all.

Applied where a scalar parameter reaches a query with no type check:

| Site | Parameter(s) |
|---|---|
| `plugins/views/api/api.js` `getHeatmap` | `view`, `actionType`, `segment` |
| `plugins/star-rating/api/api.js` `/o/feedback/data` | `widget_id`, `version`, `platform`, `uid` |
| `plugins/star-rating/api/api.js` `/o/feedback/widgets` | `is_active` |
| `plugins/crashes/api/api.js` `method=user_crashes` | `uid` |
| `api/parts/mgmt/users.js` `fetchNotes` | `note_type` |
| `api/utils/requestProcessor.js` `/i/token/delete` | `tokenid` |
| `plugins/systemlogs/api/api.js` member lookup | `api_key` |

The full survey of the pattern across `api/` and every plugin `api/` directory, including the sites that were already validated and the false positives, is in #7937.

`common.parseUserQuery` and `common.findUnsafeMongoOperator` were considered first and are not the right tool here: they validate parameters that are queries in their own right, and they only reject the JS-executing operators, so they accept `$ne` and `$regex` by design. Endpoints that legitimately take a whole user query already go through them and are untouched.

## Compatibility

Strings and numbers pass through unchanged, so legitimate callers see no difference. `/o/actions` still takes `view` as a string URL, `actionType` as `"click"` or `"scroll"` and `segment` as a string. Only non-scalars are refused, with a 400 naming the parameter, the same way `device` and `period` were already checked in the same function.

## Tests

`test/unit-tests/api.utils.common.js` gains unit tests for `common.isQueryScalar`. The heatmap integration cases that #7937 adds are not backported: `plugins/views/tests/heatmaps.js` does not exist on this branch.

## Verification

- `node --check` on all 8 changed files: clean.
- `eslint -c .eslintrc.json` on all 8 changed files: 0 errors, 0 warnings.
- `mocha test/unit-tests/api.utils.common.js`: 32 passing with this change, 28 passing on the unmodified branch, so all 4 new tests pass and nothing regressed. Both runs also show the same 3 failures in the pre-existing `validateArgs` ObjectID cases; those come from the local setup (the mongodb 6.x driver was linked into this 24.05 checkout, which pins 4.17.2, and `mongodb.ObjectID` was removed in 6.x) and are identical before and after this change.

### DIFF
--- a/plugins/star-rating/api/api.js
+++ b/plugins/star-rating/api/api.js
@@ -1259,7 +1259,7 @@ function uploadFile(myfile, id, callback) {
         //be scalars. A JSON request body can put an object here, which Mongo
         //would read as a query expression and which would widen the match
         //instead of narrowing it
-        var scalarParams = ['widget_id', 'version', 'platform', 'uid'];
+        var scalarParams = ['widget_id', 'version', 'platform', 'device_id', 'sSearch', 'uid'];
         for (var s = 0; s < scalarParams.length; s++) {
             if (!common.isQueryScalar(params.qstring[scalarParams[s]])) {
                 common.returnMessage(params, 400, 'Invalid parameter: ' + scalarParams[s]);
--- a/plugins/star-rating/api/api.js
+++ b/plugins/star-rating/api/api.js
@@ -1255,6 +1255,18 @@ function uploadFile(myfile, id, callback) {
         var limit = parseInt(params.qstring.iDisplayLength || 0);
         var colNames = ['rating', 'comment', 'email', 'ts'];
 
+        //these are matched as plain values in the query below, so they have to
+        //be scalars. A JSON request body can put an object here, which Mongo
+        //would read as a query expression and which would widen the match
+        //instead of narrowing it
+        var scalarParams = ['widget_id', 'version', 'platform', 'uid'];
+        for (var s = 0; s < scalarParams.length; s++) {
+            if (!common.isQueryScalar(params.qstring[scalarParams[s]])) {
+                common.returnMessage(params, 400, 'Invalid parameter: ' + scalarParams[s]);
+                return true;
+            }
+        }
+
         if (params.qstring.widget_id) {
             query.widget_id = params.qstring.widget_id;
         }
@@ -1414,6 +1426,12 @@ function uploadFile(myfile, id, callback) {
             var collectionName = 'feedback_widgets';
             var query = {type: "rating"};
             if (params.qstring.is_active) {
+                //matched as a plain value, so an object here would be read by
+                //Mongo as a query expression and would widen the match
+                if (!common.isQueryScalar(params.qstring.is_active)) {
+                    common.returnMessage(params, 400, 'Invalid parameter: is_active');
+                    return true;
+                }
                 query.is_active = params.qstring.is_active;
             }
 
--- a/plugins/views/api/api.js
+++ b/plugins/views/api/api.js
@@ -1360,6 +1360,20 @@ const escapedViewSegments = { "name": true, "segment": true, "height": true, "wi
             console.log('Parse device failed: ', params.qstring.device);
         }
 
+        //view, actionType and segment are matched as plain values against the
+        //drill collections below, so they have to be scalars. A JSON request
+        //body can put an object here, and Mongo would read that object as a
+        //query expression instead of a value: the query would then match far
+        //more than the one view asked for, with no bound on how much of the
+        //drill collection it has to scan.
+        var scalarParams = ["view", "actionType", "segment"];
+        for (var sp = 0; sp < scalarParams.length; sp++) {
+            if (!common.isQueryScalar(params.qstring[scalarParams[sp]])) {
+                common.returnMessage(params, 400, 'Bad request parameter: ' + scalarParams[sp]);
+                return false;
+            }
+        }
+
         var actionType = params.qstring.actionType;
 
         if (!(device.minWidth >= 0) || !(device.maxWidth >= 0)) {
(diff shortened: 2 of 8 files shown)


plugins/star-rating/api/api.js
```diff
@@ -1255,6 +1255,18 @@ function uploadFile(myfile, id, callback) {
         var limit = parseInt(params.qstring.iDisplayLength || 0);
         var colNames = ['rating', 'comment', 'email', 'ts'];
 
+        //these are matched as plain values in the query below, so they have to
+        //be scalars. A JSON request body can put an object here, which Mongo
+        //would read as a query expression and which would widen the match
+        //instead of narrowing it
+        var scalarParams = ['widget_id', 'version', 'platform', 'uid'];
```
ar2rsawseen: [P1] Include `device_id` in this scalar guard

The endpoint later assigns `params.qstring.device_id` directly to `query.did`, but this list does not validate it. A JSON body such as `{"device_id":{"$regex":".*"}}` is still interpreted by MongoDB as a query expression, widening the feedback lookup and potentially forcing the same expensive scan this fix prevents for the adjacent filters. Add `device_id` to the guarded parameters before constructing the query.

plugins/views/api/api.js
```diff
@@ -1360,6 +1360,20 @@ const escapedViewSegments = { "name": true, "segment": true, "height": true, "wi
             console.log('Parse device failed: ', params.qstring.device);
         }
 
+        //view, actionType and segment are matched as plain values against the
+        //drill collections below, so they have to be scalars. A JSON request
+        //body can put an object here, and Mongo would read that object as a
+        //query expression instead of a value: the query would then match far
+        //more than the one view asked for, with no bound on how much of the
+        //drill collection it has to scan.
+        var scalarParams = ["view", "actionType", "segment"];
```
ar2rsawseen: [P1] Require a concrete heatmap action type

`isQueryScalar()` deliberately accepts `null` and `undefined`, yet the aggregation always assigns `actionType` to `matchQuery['sg.type']`. MongoDB equality against null also matches rows where the field is absent, and ingestion permits action events without a type, so an omitted/null value can still broaden the query over unrelated rows. Require this parameter to be present—ideally one of the supported `click`/`scroll` values—before building the pipeline.

plugins/star-rating/api/api.js
ar2rsawseen: Confirmed — `device_id` goes straight into the query and was not in the list. Fixed in 1d7edf6863c.

```js
if (params.qstring.device_id) {
    query.device_id = params.qstring.device_id;
}
```

Added, and `sSearch` with it: it lands in `query.$text = {$search: …}`, which takes an object just as readily. Walking the rest of the builder, everything else is either already guarded or coerced — `rating` goes through `parseInt`, `iSortCol_0` through `parseInt` inside a try, and `ts` is built by `getTimestampRangeQuery`.

plugins/views/api/api.js
ar2rsawseen: Confirmed, and this one is the more interesting of the two — fixed in 1d7edf6863c.

You are right that being a scalar was never the property that mattered for this parameter. `isQueryScalar` accepts `null`/`undefined` deliberately, on the basis that whether an absent parameter belongs in the query is the call site's decision — and this call site was not making it. `matchQuery['sg.type'] = actionType` runs unconditionally, an equality predicate against `null` also matches documents where the field is absent, and ingestion accepts `[CLY]_action` events with no `sg.type`. So omitting the parameter *selected* those rows across the caller's whole period rather than selecting nothing, with the `$unionWith` across the drill collections unbounded.

`actionType` now has to be one of the two values `heatmap.js` ever sends:

```js
const ACTION_TYPES = ['click', 'scroll'];
if (ACTION_TYPES.indexOf(params.qstring.actionType) === -1) {
    common.returnMessage(params, 400, 'Bad request parameter: actionType');
    return false;
}
```

**The same defect is one field over, so it is fixed with it.** `view` has exactly the shape you describe:

```js
if (actionType === 'scroll') {
    matchQuery['sg.view'] = params.qstring.view;
}
else {
    matchQuery['up.lv'] = params.qstring.view;
}
```

Both unconditional. An omitted `view` matched users with **no recorded last view** rather than matching nothing — and neither branch has a meaningful "all views" reading to preserve, so it now has to be a non-empty string. Both are removed from `scalarParams`, since the stronger check subsumes it.

Tests cover omitted, `null` and unsupported `actionType`, omitted and empty `view`, and that both supported types still return 200. One placement note: `plugins/views/tests/heatmaps.js` only exists on master — the platform copy is `heatmaps.js.old` and the 24.05 branch has no views test directory — so those two carry the code fix and the `isQueryScalar` unit cases.

PR #7940, opened 2026-08-14
[fix][core] stop returning credentials from the current user endpoint (24.05)

ar2rsawseen:
Backport of https://github.com/Countly/countly-server/pull/7939 to release.24.05.

## What

`/o/users/me` returned the caller's member document with only the password removed:

```js
usersApi.getCurrentUser = function(params) {
    delete params.member.password;
    common.returnOutput(params, params.member);
```

So the response carried the account's `api_key`, and for members using two factor auth the stored secret for that factor as well. The neighbouring reads already treat both as sensitive: `getUserById` and `getAllUsers` project `api_key` away, and the member event payloads delete it with a comment saying it must never be forwarded. This was the one read path that handed it out.

Separately, a token that was restricted to specific applications could reach this endpoint. The restriction in `verify_token` is only compared when the request itself names an application, and this request never does, so the restriction was not applied.

## Change

1. Answer with a copy of the member that has `password`, `api_key` and `two_factor_auth` removed. A copy, so nothing later in the request sees a mutated member.
2. Refuse a token restricted to specific applications on this endpoint. It reports the caller's own account and belongs to no application, so an application restricted token has no business reading it.

Tokens with no application restriction behave exactly as before.

## Why this is safe to change

- **Nothing in the product reads the key from here.** The dashboard takes `api_key` from the server rendered globals, and there is a dedicated `/api-key` route for a member who wants their own key. No frontend code calls `/o/users/me` at all.
- **No test expects the key.** The existing assertions on this endpoint check `email`, `full_name`, `global_admin` and `username`.
- **The existing token suites still pass by construction.** The token used against `/o/users/me` in `test/2.api/14.authorize.token.js` is created with no `app`, so it is unrestricted and unaffected. The token in `App-scoped token enforces its app restriction` is pinned to `^/o/actions`, so it never reaches this endpoint either.
- `two_factor_auth.enabled` remains available where the UI actually reads it, from the user listing.

## Verification

Two cases added:

- `test/2.api/02.read.user.js`: the response still carries the account's own fields but has no `api_key`, no `password` and no `two_factor_auth`.
- `test/2.api/14.authorize.token.js`: a token restricted to one application, with no endpoint restriction, is refused on `/o/users/me`. This is a different token shape from the existing app restriction case, which is endpoint pinned.

Lint and `node --check` clean on all changed files.

### DIFF
--- a/api/utils/requestProcessor.js
+++ b/api/utils/requestProcessor.js
@@ -1721,7 +1721,31 @@ const processRequest = (params) => {
                     validateUserForGlobalAdmin(params, countlyApi.mgmt.users.getAllUsers);
                     break;
                 case 'me':
-                    validateUserForMgmtReadAPI(countlyApi.mgmt.users.getCurrentUser, params);
+                    validateUserForMgmtReadAPI(function() {
+                        //This endpoint answers with the caller's own account and belongs to no
+                        //application, so a token that was deliberately limited to some
+                        //applications has no business reading it. Without this an app limited
+                        //token still reached account level data, because the app restriction in
+                        //verify_token is only compared when the request itself names an app.
+                        var authToken = params.qstring.auth_token || params.req.headers["countly-token"] || "";
+                        if (!authToken) {
+                            return countlyApi.mgmt.users.getCurrentUser(params);
+                        }
+                        authorize.read({
+                            db: common.db,
+                            token: authToken,
+                            callback: function(tokenErr, tokenData) {
+                                //save() stores app as "" when unrestricted and as an array
+                                //otherwise, so a non empty length is what marks a restriction
+                                if (tokenData && tokenData.app && tokenData.app.length) {
+                                    common.returnMessage(params, 401, 'Token is restricted to specific applications');
+                                    return false;
+                                }
+                                return countlyApi.mgmt.users.getCurrentUser(params);
+                            }
+                        });
+                        return true;
+                    }, params);
                     break;
                 case 'id':
                     validateUserForGlobalAdmin(params, countlyApi.mgmt.users.getUserById);
--- a/api/utils/requestProcessor.js
+++ b/api/utils/requestProcessor.js
@@ -1727,24 +1727,19 @@ const processRequest = (params) => {
                         //applications has no business reading it. Without this an app limited
                         //token still reached account level data, because the app restriction in
                         //verify_token is only compared when the request itself names an app.
-                        var authToken = params.qstring.auth_token || params.req.headers["countly-token"] || "";
-                        if (!authToken) {
-                            return countlyApi.mgmt.users.getCurrentUser(params);
+                        //
+                        //params.token_data is the document the validation above already read.
+                        //It is deliberately not looked up again: verify_token consumes a single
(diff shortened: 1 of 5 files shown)


api/utils/requestProcessor.js
```diff
@@ -1721,7 +1721,31 @@ const processRequest = (params) => {
...
+                        if (!authToken) {
+                            return countlyApi.mgmt.users.getCurrentUser(params);
+                        }
+                        authorize.read({
+                            db: common.db,
+                            token: authToken,
+                            callback: function(tokenErr, tokenData) {
+                                //save() stores app as "" when unrestricted and as an array
+                                //otherwise, so a non empty length is what marks a restriction
+                                if (tokenData && tokenData.app && tokenData.app.length) {
```
ar2rsawseen: [P1] Fail closed when restriction metadata cannot be re-read

A single-use (`multi=false`) token can be consumed by the outer `validateUserForMgmtReadAPI` verification before this second `authorize.read()` runs. That lookup then returns no `tokenData`, and this condition treats absence (or a lookup error) as unrestricted and returns the account response. Carry the app restriction from the original verification, or reject unless this lookup succeeds with a token document; missing metadata must not mean unrestricted.

ar2rsawseen: Confirmed, and taken the first of your two options — the restriction is carried out of the original verification rather than re-read. Fixed in ca0fcf64a1c.

The consumption is right there in `verify_token`:

```js
//consume token if expired or not multi
if (!res.multi || (res.ttl > 0 && res.ends < Math.round(Date.now() / 1000))) {
    options.db.collection("auth_tokens").remove({_id: options.token});
}
```

So for `multi: false` the row is gone before the second `authorize.read` ever runs — and the check read that absence as *unrestricted*. The restriction was dropped for exactly the tokens meant to be the most limited. Same for a lookup error.

`validate_token_if_exists` now asks `verify_return` for the document instead of the owner id alone, and keeps it:

```js
return_data: true,
callback: function(valid) {
    if (valid && typeof valid === "object") {
        params.token_data = valid;
        resolve(valid.owner);
    }
    ...
```

`resolve(valid.owner)` is the same value every existing caller was already getting (`verify_token` returns `res.owner` in the `return_owner` branch and `res` in the `return_data` one), so nothing downstream changes. The handler then reads `params.token_data` and the second query is gone along with the window. A request authorized by an `api_key` carries no `token_data` and has no restriction to honour, which is the only case where absence legitimately means "unrestricted".

Two incidental improvements from doing it this way:

- **The `app` field is normalised by then.** `verify_token` converts a legacy bare-string `app` to an array before the callback. The old re-read got the raw document, where `.length` on a string happened to be truthy — right answer, wrong reason.
- **One fewer database round trip** on every `/o/users/me`.

**Test:** a `multi: false` token restricted to an app, refused on the single use it gets. It passes only because the restriction is carried — against the previous shape the token is already deleted when the check runs. The platform PR had no tests at all, so it now carries both the reusable and the single-use case.

`params.token_data` is declared on the `Params` type; tsc diagnostic count is unchanged (7358 before, 7358 after).

Cookiezaurs (approved):

PR #7941, opened 2026-08-14
[fix][hooks] keep app credentials out of the effect payload

ar2rsawseen:
## What

An app document carries the credentials that authenticate writes to that app: `key` (the sdk key), `keys[]` (every rotated key), `id_key`, `salt` and `checksum_salt`. Four internal events carry such a document into the hooks pipeline:

| dispatch | payload |
| --- | --- |
| `plugins/crashes/api/api.js` `/crashes/new` | `{crash, user, app}` |
| `api/parts/mgmt/apps.js` `/i/apps/update` | `{app: appBefore, update}` |
| `api/parts/mgmt/apps.js` `/i/apps/reset` | the app document |
| `api/parts/mgmt/apps.js` `/i/apps/delete` | the app document |

Hooks hands the payload to its effects, and an effect can emit it verbatim: the http effect's body is a template, and `{{payload_json}}` stringifies the whole payload to a url the hook's author chose. So a member who can create a hook on an app could have those fields posted to a host they control.

## Change

Remove those fields where hooks takes the payload over: the single funnel every internal event passes through in `internal_event.js`, before the `_originalInput` snapshot copies it as well.

Keyed off the event type rather than by field name, because `key` is an ordinary field elsewhere. An event has one, and a blanket scrub would break any hook that references it.

## Why here and not at the dispatch sites

Because `systemlogs` records those payloads whole, deliberately, so that a deleted or reset app can be recovered afterwards. Stripping at the source would remove exactly the fields that recovery needs.

The scrub therefore works on copies and never mutates the object the other subscribers of the same dispatch see. That is safe to rely on: across the three repositories these four events have **150 subscriber registrations**, and none of them reads `key`, `keys`, `id_key`, `salt` or `checksum_salt`.

| repo | registrations | payload fields actually read |
| --- | --- | --- |
| countly-server | 35 | `appId` x31, `data` x4, `params` x3 |
| countly-enterprise-plugins | 37 | `appId` x37, `data` x1 |
| countly-platform | 78 | `appId` x72, `data` x5, `params` x3 |

The `data` readers use `data._id`, `data.update.name`, `data.update`, `data.app` for a field diff, and `triggerByEvent(data)` in alerts. All of that survives the scrub. Checked by extracting each handler body and resolving destructuring and aliases, not by a single line grep.

## Second change, correctness rather than security

Three checks compared `rule.apps[0]` instead of testing membership, so a hook configured with several apps only fired for the first one. The neighbouring cohort and crash checks already use `indexOf`. Replaced with an `indexOf` membership test. This one failed closed, so it never delivered across apps.

## Verification

- `plugins/hooks/tests/internal_event_payload.js`, 16 cases, wired into `plugins/hooks/tests/index.js` so it actually runs: every app carrying event is stripped, the dispatched payload itself is left intact for each one, the fields an effect uses (`_id`, `name`, `timezone`, `appId`) survive, and four unrelated payloads that legitimately contain `key` are unchanged. Replacing the scrub with a pass through fails 5 of them.
- Lint and `node --check` clean on the changed files.


plugins/hooks/api/parts/triggers/internal_event.js
```diff
@@ -59,6 +59,60 @@ async function isRuleOwnerGlobalAdmin(rule, cache) {
...
+ */
+function withoutSecrets(params, eventType) {
+    if (!params || typeof params !== "object") {
+        return params;
+    }
+    const out = Object.assign({}, params);
+    if (out.data && typeof out.data === "object") {
+        //crashes/new and /i/apps/update nest the app document under data.app
+        if (out.data.app && typeof out.data.app === "object") {
+            out.data = Object.assign({}, out.data, {app: withoutAppSecrets(out.data.app)});
```
ar2rsawseen: [P1] Strip credentials from the app update object too

For `/i/apps/update`, the dispatch payload contains both `data.app` and `data.update`. The update object is built from the accepted app keys and can include `keys` on every update plus `key` during rotation (and legacy `id_key`), but this branch sanitizes only `data.app`. `{{payload_json}}` can therefore still forward credentials through `data.update`, leaving the exfiltration path open. Sanitize both app-shaped objects before the effect pipeline.

ar2rsawseen: Confirmed, and worse than "potentially" — fixed in a1a356da70d.

`data.update` is not the safer of the two objects. `api/parts/mgmt/apps.js` rebuilds the accepted key list on **every** update, unconditionally:

```js
var keysArr = (Array.isArray(appBefore.keys) && appBefore.keys.length)
    ? appBefore.keys.slice()
    : [{key: appBefore.key, added_at: …, last_data: …}];
…
updatedApp.keys = keysArr;
```

So `data.update.keys` carries every key the app has ever had on every single app update — renaming an app was enough. `data.update.key` is there whenever the key is rotated, and `data.update.id_key` is filled in the first time an app that predates it is touched. The exfiltration path was open on the most common app operation there is.

The cause was the `if / else if` shape: it scrubbed whichever app-shaped object it found first and stopped. Both named fields are now scrubbed, and the "the payload *is* an app document" branch is only taken when neither is present — a payload carrying `app` or `update` is not itself an app:

```js
const APP_SHAPED_FIELDS = ["app", "update"];
```

Still keyed off field names and event type rather than by looking for a field called `key` anywhere — `key` is an ordinary field elsewhere (an event has one), and a blanket scrub would break any hook that references it. Those cases are in the suite.

**Why it wasn't caught:** the fixture used `update: {name: 'renamed'}`, which has nothing to strip. It now builds the update object the way `apps.js` builds it — `keys`, `key`, `id_key`, `edited_at`, no `checksum_salt` — and there is a case asserting `data.update` is cleaned *while `data.app` sits beside it*, which is the exact shape that regressed.

I also checked the neighbouring dispatches while I was here: `/i/users/create|update|delete` already delete `password` and `api_key` at the dispatch site, and `/i/apps/clear` is not in `InternalEvents`, so neither is a second instance of this.

One placement note: the 24.05 branch has no `plugins/hooks/tests` directory, so the suite lives in `test/unit-tests` there; the platform PR had no test for this at all and now carries it as a `.unit.js` beside the plugin. 17 passing on all three.

Cookiezaurs (approved):

PR #7950, opened 2026-08-17
security(core): escape "<" in the res.expose script island (24.05)

ar2rsawseen:
Backport of #7949 to `release.24.05`.

`frontend/express/libs/express-expose.js` serializes the exposed `countlyGlobal` object into an inline `<script>` block. It neutralized only the exact `</script>` sequence, but the HTML tokenizer also ends a script at `</script >`, `</script/>` and other whitespace/slash spellings, so an **application name** containing one broke out. Because a global admin's dashboard lists every app, an **app admin of one app** could store such a name and have it execute in any global admin's session — an app-admin-to-global-admin escalation. Medium under [SECURITY.md](https://github.com/Countly/countly-server/blob/master/SECURITY.md).

Fix (identical to #7949): escape every `<` as `<` in both serialization paths — string values and object keys (`escape_js_string`). `<` parses back to `<`, so values read from the exposed object are unchanged. Also render the active-app name with `.text()` instead of `.html()` in `countly.template.js` (the only `.html()` sink fed by an app name).

Verified on this branch: no raw `<` survives for any breakout spelling in values or keys, and an eval round-trip reproduces the input object byte-for-byte and matches the previous serializer output (no dashboard-visible change).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ Fixes:
 Enterprise Fixes:
 - [data-manager] Fixed editing an event whose key contains `&` creating undeletable duplicate rows in the events table
 
+Security Fixes:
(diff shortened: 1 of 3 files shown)


ar2rsawseen:

PR #7955, opened 2026-08-18
security(star-rating): authorize the /o?method=star ratings read

ar2rsawseen:
### Summary

`/o?method=star` returns the platform and application-version combinations that have received star ratings. It performed **no authorization**, so any caller who supplies an `app_id` received that application's data with no `api_key`, no `auth_token` and no session.

Measured against the running API with no credentials at all:

```
GET /o?method=star&app_id=<app>&period=30days
  -> 200  {"Linux":["6:2:1","6:2:0"],"iOS":["1:23"],"tvOS":[...],"MacOS":[...]}

GET /o/feedback/data?app_id=<app>&period=30days        (sibling, same request, no creds)
  -> 400  {"result":"Missing parameter \"api_key\" or \"auth_token\""}
```

### Why it was reachable

Authentication on `/o` is **per method**, not global. Every core method authorizes itself (`case 'durations'` → `validateUserForDataReadAPI(...)`, `case 'events'` → `validateUserForDataReadAPI(...)`), and the `default:` branch hands the validators to plugins as *helpers*:

```js
default:
    if (!plugins.dispatch(apiPath, {
        params: params,
        validateUserForDataReadAPI: validateUserForDataReadAPI,   // passed, never called
        ...
    })) { common.returnMessage(params, 400, 'Invalid path, ...'); }
```

`dispatch` returns truthy when a plugin claims the request, and the only fallback is `400 Invalid path` when nothing does. So a plugin registering on `/o` must authorize the request itself. Every other plugin does (`views/api/api.js:879` → `validateRead` at :886; `times-of-day:177` → :185), and this plugin does it in its own siblings (`/o/feedback/data:1258` → :1302, `/o/feedback/widgets:1420` → :1422). The `method === 'star'` branch claimed the request, returned `true`, and never called a validator.

This was an omission rather than a design choice: the only caller is the dashboard Ratings page (`starRatingPlugin.requestPlatformVersion` in `plugins/star-rating/frontend/public/javascripts/countly.models.js`), which sends the session credential and the active app. The plugin's deliberately public endpoints are a separate, clearly named family (`/o/sdk`, `/feedback/widgets`, `/i/feedback/input`).

### Impact

Cross-application disclosure of a specific customer's platform and application-version inventory, to an unauthenticated caller who knows the `app_id`. Rating comments and detailed feedback live in `/o/feedback/data`, which is authorized and unaffected. Medium under [SECURITY.md](https://github.com/Countly/countly-server/blob/master/SECURITY.md).

### Fix

Wrap the branch in `validateRead(params, FEATURE_NAME, …)`, the same check the sibling reads already apply. Authorization runs **before** the `period` parameter is validated, so an unauthorized caller cannot probe the endpoint through its error responses.

`app_id` needs no extra guard: the core `/o` case rejects a request without one, and this branch only concatenates it into a collection name, so there is no `ObjectID()` conversion that could throw for a global admin (whose `validateRead` does not require `app_id`). The diff is mostly re-indentation from introducing the callback.

### Scope of the class

Every plugin `/o` handler in countly-server, countly-platform and countly-enterprise-plugins was swept for a missing validator, and the surviving read candidates were probed unauthenticated. `method=star` is the only unauthenticated tenant-data read. The other credential-less responders return hardcoded lookup tables that ignore `app_id` (`/o/langmap`, `/o/sources`), are widget configuration for end-user SDKs (`/o/feedback/widget`), or are the intentional hooks API-endpoint trigger (`/o/hooks`). `/o/surveys/*`, `/o/calculated_metrics/*` and `/o/journey-engine/folders` all authorize correctly.


plugins/star-rating/api/api.js
```diff
@@ -1577,79 +1577,85 @@ function uploadFile(myfile, id, callback) {
...
+                                doc.meta.platform_version_rate = [];
                             }
-                            if (result[data[0]].indexOf(data[1]) === -1) {
-                                result[data[0]].push(data[1]);
+                            if (doc.meta_v2 && doc.meta_v2.platform_version_rate) {
+                                common.arrayAddUniq(doc.meta.platform_version_rate, Object.keys(doc.meta_v2.platform_version_rate));
                             }
+                            doc.meta.platform_version_rate.forEach(function(item) {
+                                var data = item.split('**');
+                                if (result[data[0]] === undefined) {
```
ar2rsawseen, line 1768: [P2] Use a null-prototype map for stored platform names

`data[0]` comes from the public star-rating event's `platform_version_rate` segmentation and is used as an object key. Values such as `__proto__`, `constructor`, or `toString` resolve inherited members on `{}`, so this condition does not initialize an array and the following `.indexOf()` throws, making the authorized ratings read fail on attacker-seeded data. Build `result` with `Object.create(null)` or use an own-property check and explicitly initialize each key. The equivalent accumulator in the alternate/granular branch needs the same treatment.

ar2rsawseen: Confirmed, and it is a denial rather than a corruption — fixed in 10dcb09d55e.

Ran the loop rather than reasoning about it. Against a plain object, each of these throws:

```
TypeError: result[data[0]].indexOf is not a function
```

for `__proto__` (reads back as `Object.prototype`), `constructor` (the `Object` function), and `toString` / `valueOf` / `hasOwnProperty` (functions). None is `undefined`, so the "not seen yet" branch never runs and the array is never created. One planted row therefore fails the read **for every authorized caller** until it ages out — and the name comes off the public star-rating event's `platform_version_rate` segmentation, so anyone who can write to the app chooses it.

Took the `Object.create(null)` option:

```js
var result = Object.create(null);
```

Nothing downstream changes — the keys are still ordinary strings, and `JSON.stringify` serialises a null-prototype object identically, which is what `returnOutput` does with it. There is a test pinning that.

**And the granular branch too**, as you asked: the platform copy has a second accumulator over `data2.data[z]._id.split('**')` and it gets the same treatment. The server copies only have the one.

On the tests: they lift **both** the accumulator's declaration and the loop out of the real source, rather than the loop alone. That matters here — a test that built its own `result` would pass whichever object the shipping code used. Four of the five fail against the previous code, with the TypeError above.

Cookiezaurs (approved):

PR #7956, opened 2026-08-18
security(star-rating): authorize the /o?method=star ratings read (24.05)

ar2rsawseen:
Backport of #7955 to `release.24.05`.

`/o?method=star` returns the platform and application-version combinations that have received star ratings, and performed **no authorization** — any caller supplying an `app_id` received that application's data with no `api_key`, no `auth_token` and no session, while the sibling `/o/feedback/data` correctly answers `400 Missing parameter "api_key" or "auth_token"` on the identical credential-less request.

Authentication on `/o` is per method, not global: core methods call `validateUserForDataReadAPI` themselves, and the `default:` branch hands the validators to plugins as helpers without calling them, so a plugin that claims a request must authorize it. This branch claimed the request, returned `true`, and never called a validator.

Fix (identical to #7955): wrap the branch in `validateRead(params, FEATURE_NAME, …)`, the same check the sibling reads in this file already apply, running before the `period` parameter is validated so an unauthorized caller cannot probe the endpoint through its error responses. The diff is mostly re-indentation from introducing the callback.

The only caller is the dashboard Ratings page (`starRatingPlugin.requestPlatformVersion`), which sends the session credential and an app the member has access to, so it is unaffected. `eslint` clean.


plugins/star-rating/api/api.js
```diff
@@ -1568,78 +1568,84 @@ function uploadFile(myfile, id, callback) {
...
+                                doc.meta.platform_version_rate = [];
                             }
-                            if (result[data[0]].indexOf(data[1]) === -1) {
-                                result[data[0]].push(data[1]);
+                            if (doc.meta_v2 && doc.meta_v2.platform_version_rate) {
+                                common.arrayAddUniq(doc.meta.platform_version_rate, Object.keys(doc.meta_v2.platform_version_rate));
                             }
+                            doc.meta.platform_version_rate.forEach(function(item) {
+                                var data = item.split('**');
+                                if (result[data[0]] === undefined) {
```
ar2rsawseen, line 1762: [P2] Use a null-prototype map for stored platform names

`data[0]` comes from the public star-rating event's `platform_version_rate` segmentation and is used as an object key. Values such as `__proto__`, `constructor`, or `toString` resolve inherited members on `{}`, so this condition does not initialize an array and the following `.indexOf()` throws, making the authorized ratings read fail on attacker-seeded data. Build `result` with `Object.create(null)` or use an own-property check and explicitly initialize each key. The equivalent accumulator in the alternate/granular branch needs the same treatment.

ar2rsawseen: Confirmed, and it is a denial rather than a corruption — fixed in 2e71724297b.

Ran the loop rather than reasoning about it. Against a plain object, each of these throws:

```
TypeError: result[data[0]].indexOf is not a function
```

for `__proto__` (reads back as `Object.prototype`), `constructor` (the `Object` function), and `toString` / `valueOf` / `hasOwnProperty` (functions). None is `undefined`, so the "not seen yet" branch never runs and the array is never created. One planted row therefore fails the read **for every authorized caller** until it ages out — and the name comes off the public star-rating event's `platform_version_rate` segmentation, so anyone who can write to the app chooses it.

Took the `Object.create(null)` option:

```js
var result = Object.create(null);
```

Nothing downstream changes — the keys are still ordinary strings, and `JSON.stringify` serialises a null-prototype object identically, which is what `returnOutput` does with it. There is a test pinning that.

**And the granular branch too**, as you asked: the platform copy has a second accumulator over `data2.data[z]._id.split('**')` and it gets the same treatment. The server copies only have the one.

On the tests: they lift **both** the accumulator's declaration and the loop out of the real source, rather than the loop alone. That matters here — a test that built its own `result` would pass whichever object the shipping code used. Four of the five fail against the previous code, with the TypeError above.

Cookiezaurs (approved):

PR #7958, opened 2026-08-18
security(core): stop scoped tokens from escalating via token create and login (24.05)

ar2rsawseen:
Backport of #7957 to `release.24.05`.

A token scoped to one app could escalate to the owner's full account: `/i/token/create` isn't gated by the token's app/endpoint restriction (it carries no `app_id`), `validateUser` resolves any bearer token to the full member, the handler saved caller-supplied `app`/`endpoint`/`purpose` with no relationship to the creating credential, and `/login/token/:token` then granted the owner's full session from any token with a login purpose regardless of scope.

Fix (identical to #7957), both ends:
- `/i/token/create` via a token is allowed only if that token has no app and no endpoint restriction; api_key callers and unrestricted tokens (what a dashboard session token is) are unaffected.
- `/login/token/:token` additionally requires the token to be unrestricted; legitimate session tokens always are.

7 regression tests in `test/2.api/16.token.manager.js` (run by `test-api-core`). eslint clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ Fixes:
 Enterprise Fixes:
 - [data-manager] Fixed editing an event whose key contains `&` creating undeletable duplicate rows in the events table
 
+Security Fixes:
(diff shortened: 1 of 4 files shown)


ar2rsawseen:
Closing in favor of a broader token-model rework that addresses this at the root rather than with the two containment checks here. The branch is retained as a fallback and can be reopened if needed.

PR #7961, opened 2026-08-18
security(compliance-hub): HTML-encode the application name in the export-history action cell

ar2rsawseen:
The export/purge history model (`plugins/compliance-hub/frontend/public/javascripts/countly.models.js`) builds each row's action cell as an HTML string, and `exportHistory.html` renders it with `v-html`. Every other value in that string comes from the API, which HTML-encodes its output, but the application name is read from `countlyGlobal`, whose values are raw at runtime. So an application name is placed into the `v-html` string unencoded.

Encode the application name with `countlyCommon.encodeHtml` so it renders as text, matching the encoding of the other fields in the same string.

Only the application name needed it — the remaining fields are already API-encoded and must not be double-encoded; the display is unchanged. This is a distinct sink from #7949 (that fix covered the inline serializer and the active-app jQuery sink); this is downstream HTML construction in the default Compliance Hub plugin.

Scope check: a codebase scan for a raw `countlyGlobal` value reaching an HTML sink found only two sites — the active-app name (fixed in #7949) and this one. The wider `v-html`/`unescapeHtml` source audit is tracked separately.

Reported through the security bug bounty program.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Enterprise Fixes:
 - [data-manager] Fixed editing an event whose key contains `&` creating undeletable duplicate rows in the events table
 
 Security Fixes:
+- [compliance-hub] The export/purge history table now HTML-encodes the application name before it is placed in the action cell, so an application name is shown as text rather than markup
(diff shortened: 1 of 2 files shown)


ar2rsawseen:

PR #7962, opened 2026-08-18
security(compliance-hub): HTML-encode the application name in the export-history action cell (24.05)

ar2rsawseen:
Backport of #7961 to `release.24.05`.

The export/purge history model builds each row's action cell as an HTML string rendered with `v-html`. The other values are API-encoded, but the application name is read from `countlyGlobal`, whose values are raw at runtime, so it reached the `v-html` string unencoded. Encode it with `countlyCommon.encodeHtml` so it renders as text; only the application name needed it and the display is unchanged.

Identical to #7961. eslint clean.


ar2rsawseen:

PR #7964, opened 2026-08-18
security(core): scope auth tokens with explicit CRUD permissions and server-assigned login (24.05)

ar2rsawseen:
### Summary

Backport of #7963 to `release.24.05`.

Replaces the token `app`/`endpoint` scoping model with explicit CRUD permissions, and makes permission to open a dashboard session an explicit, server-assigned property of the token. See #7963 for the full description of the model.

### One addition specific to this branch

This backport also takes `hasAdminAccess` from `5e6014496da` ([SER-2766]), which is on later versions but was never backported here. On this branch it treats an app with **no permission entry for the member** as one they administer, because `isAdmin` starts as `true` and is only cleared when an entry exists and is not `all`.

That has to be corrected for the permission model to mean anything here: an intersected member expresses "no access to app B" as the *absence* of an entry for B, which this branch would otherwise read as administering B. It is also reachable independently of tokens, since `validateUserForWrite` gates solely on `hasAdminAccess`.

Verified on this branch before and after: a member with access to app A only previously returned `true` from `hasAdminAccess(member, B)`, `hasReadRight('core', B, member)` and `hasDeleteRight('core', B, member)`; all now return `false`, while legitimate access to A is unchanged.

### Verification

Ported by cherry-pick; three conflicts resolved by hand (the `rights.js` import list, which lacks `getBaseAppFilter` here, and the CHANGELOG and localization sections, which differ from master).

The same 21 unit tests pass on this branch. The 5 unrelated failures in `api.utils.common.js` and `api.utils.countly-request.js` are pre-existing on `release.24.05` and fail identically without this change.

The API-level suite was exercised on `master` (see #7963: 238 passing / 21 failing, the 21 pre-existing and identical on pristine master, with all 16 token cases green). This branch carries the same two commits, including the fix for a hole those tests found - a scoped credential could mint an unrestricted child by omitting the `permission` parameter, since the subset check then had nothing to compare.

Note the three token-management endpoints now require a full-permission credential, so a legacy app- or endpoint-restricted token is refused at `/i/token/create`, `/o/token/list` and `/i/token/delete`. That is deliberate: `/o/token/list` returns whole token documents and a document's `_id` is the token itself.

### Related

- countly-server `master`: #7963
- countly-platform: Countly/countly-platform#1119
- countly-enterprise-plugins (OIDC mint, 24.05): Countly/countly-enterprise-plugins#3380

Supersedes the earlier containment PR #7958, closed in favour of this model.

[SER-2766]: https://countly.atlassian.net/browse/SER-2766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/utils/authorizer.js
```diff
@@ -249,7 +270,13 @@ var verify_token = function(options, return_owner, return_data) {
...
                     res.endpoint = "";
                 }
-                if (res.endpoint && res.endpoint !== "") {
+                // The endpoint regex is the pre-permission-model scoping mechanism. It only ever
+                // constrained which paths a token could call, never what the resolved member was
+                // allowed to do, so it is not an authorization boundary. Tokens issued under the
+                // permission model carry token_permission instead, which rights.js intersects with
+                // the owner's own permissions on every request. Legacy tokens (no token_permission)
+                // keep being matched exactly as before, so existing integrations are unaffected.
+                if (!res.token_permission && res.endpoint && res.endpoint !== "") {
```
ar2rsawseen, line 279: [P1] Do not bypass endpoint scoping until every verifier applies token permissions

This disables the legacy endpoint check for any token carrying `token_permission`, assuming the request will be intersected in `rights.js`. `/o/actions` is a counterexample: its countly-token branch calls `authorize.verify_return()` directly, receives only the owner, and then invokes `getHeatmap()` without any rights validator. A token scoped to another feature on the same app therefore passes its derived app list and reads heatmap data it was not granted. Return/apply the token document in that branch, or centralize permission enforcement in authorization before skipping endpoint restrictions.

api/utils/requestProcessor.js
```diff
@@ -2615,6 +2626,12 @@ const processRequest = (params) => {
                 */
                 case 'delete':
                     validateUser(() => {
+                        //revoking the owner's other credentials is credential management, not
+                        //something a token narrowed to a subset of the owner's access may do
+                        if (isScopedCredential(params)) {
+                            common.returnMessage(params, 403, "A restricted token cannot delete tokens");
```
ar2rsawseen, line 2712: [P1] Block scoped tokens from account-level 2FA management too

This protects token deletion as credential management, but the same scoped token can call `/i/two-factor-auth?method=disable`, `enable`, or `generate-qr-code`: those routes use only `validateUser`, which applies the app permission intersection but does not reject the request, and `disable` needs no current TOTP code. A token narrowed to one app/feature can therefore weaken or replace its owner's second factor. Apply `isScopedCredential` to these account-level authentication routes (and audit other `validateUser`-only credential mutations), not just the token manager.

api/utils/authorizer.js
ar2rsawseen: Confirmed — `/o/actions` was the one route where `applyTokenScope` never ran. Fixed in 0bc0f171719.

It resolves its own token: `authorize.verify_return` in the `countly-token` branch, then straight to `getHeatmap`. So only the token's **app** restriction was ever checked, and its permissions were not consulted at all — a token scoped to another feature on the *same* app passed, exactly as you describe.

Took your first option, since the second (centralising in authorization) would mean the authorizer deciding feature-level rights, which is what `rights.js` is for:

```js
return_data: true,
callback: function(tokenData, expires_after) {
    …
    params.token_data = tokenData;
    common.db.collection('members').findOne({_id: common.db.ObjectID(tokenData.owner + "")}, function(memberErr, member) {
        var scoped = (memberErr || !member) ? null : applyTokenScope(params, member);
        if (!viewsUtils.ownerCanRead(scoped, app._id + "", FEATURE_NAME)) {
            common.returnMessage(params, 401, 'User does not have view right for this application');
            return false;
        }
        …
```

`applyTokenScope` is exported for it, so this route bounds the member the same way and in the same order `rights.js` does — before anything is authorized.

Two details in the read check, so it does not over-correct:

- **Legacy `user_of` membership is honoured.** `hasReadRight` alone does not cover it, but `validateRead` does, so refusing it here would take away access the same member's `api_key` still has.
- **A locked account is refused**, as `validateRead` refuses it. This branch checked neither before.

Endpoint scoping stays disabled for permission-carrying tokens, which I read as the intent of the change rather than the problem — the problem was that nothing applied the permissions in its place on this route. Now something does.

Unit cases cover `applyTokenScope` directly: unchanged for a token with no permissions, narrowed for one with them (`events` on the app is refused though the *owner* has it), never widened, and a global admin still bounded by their own token's scope.

api/utils/requestProcessor.js
ar2rsawseen: Confirmed on all three methods — fixed in 0bc0f171719.

`disable` is the sharp one: it takes no current TOTP code, so a token narrowed to one feature on one app could simply switch off the factor protecting everything its owner can reach. `enable` replaces the secret and `generate-qr-code` hands out a fresh one to enable with, so all three are credential management in the sense the token manager already uses.

All three now refuse a scoped credential through one helper:

```js
function refuseScopedCredential(params, what) {
    if (isScopedCredential(params)) {
        common.returnMessage(params, 403, "A restricted token cannot " + what);
        return true;
    }
    return false;
}
```

You are right about the mechanism as well as the routes: `validateUser` bounds what a scoped token may touch **per app**, and an account-level route has no app to bound — so bounding is not the same as refusing, and only refusing helps here.

**On the audit you asked for**, the other `validateUser`-only routes in `requestProcessor.js`:

| route | verdict |
|---|---|
| `/i/token/create` | already ceilinged — `params.member` is the *creating credential's* authority, so a scoped token cannot mint a wider child |
| `/o/token/check` | reports on a token supplied as a parameter; the caller must already hold it |
| `/o/countly_version`, `getCurrentUserApps` | read-only, no credential state |
| `/i/users/update`, `admin_check`, `admin_disable` | global admin, not `validateUser` |

So the 2FA trio was the whole set. The 24.05 branch has no `generate-qr-code` method, so it gets the two it has.

Integration tests cover all three refusing a scoped token, **and** an unrestricted `api_key` still getting through — a guard that also blocks the legitimate path would be its own bug.

---

Separately, while adding those: on the **platform** PR both unit suites this branch adds sit in `test/unit-tests/`, which the platform's runner does not glob —

```
"test:unit": "… mocha … 'test/unit/*.unit.js' 'plugins/*/tests/**/*.unit.js' …"
```

— so neither `api.utils.rights.tokenPermissions.js` nor `api.utils.authorizer.tokenFields.js` has ever run in platform CI. Moved to `test/unit/` with the `.unit.js` suffix; 25 passing once they actually run. (The server repo globs `test/unit-tests/`, so they were fine there.)

ar2rsawseen:
### Conflict resolved — `0e790c75a84`

`release.24.05` moved under this branch: **#7873 merged**, which is the `hasAdminAccess` fail-open fix. This branch carried its own copy of the same fix, so the two collided inside that function and git interleaved them into a duplicated body.

Resolved by taking the **release branch's** version of `hasAdminAccess` whole and dropping this branch's duplicate — #7873 is the one that was reviewed and merged, and the two implementations were otherwise identical (`isAdmin = false` plus the `passesAllRules` loop).

Two consequences worth stating:

- **The stacking note on my earlier reply is now moot.** #7873 has merged, so this branch no longer depends on it and the two tests that needed it pass on the release branch as it stands.
- `api/utils/rights.js` keeps every other change this PR makes; only the one conflicting function was replaced.

Verified after the merge: eslint clean, and 31 passing across `api.utils.rights.tokenPermissions.js` + `api.utils.rights.hasAdminAccess.js`, plus 11 passing / 1 pending on the authorizer and two-factor suites (the pending one is `generate-qr-code`, which does not exist on 24.05).

ar2rsawseen:
### Merged the base again — #7936 and #7940 had landed

Three code conflicts, all the same shape: `release.24.05` now carries the merged versions of work this branch also has a copy of.

| file | kept | why |
|---|---|---|
| `api/utils/rights.js` | **base** | #7940's version of `validate_token_if_exists` has the `typeof valid === "object"` narrowing and the fuller note about a single-use token being consumed. This branch had the earlier, plainer one. |
| `plugins/views/api/parts/viewsUtils.js` | **this branch** (doc comment only) | the `ownerCanRead` **body is byte-identical** on both sides — it sits outside the conflict. Only the comment differs, and this branch's says the member must already be token-scoped, which is true here. |
| `plugins/views/api/api.js` | **this branch** | ours applies `applyTokenScope` before the read check; #7936's version checks the owner's own rights alone. Ours is the stronger of the two and is the point of this PR. |

Verified after: eslint clean (with `--rulesdir bin/eslint-rules`), 47 passing / 1 pending across the rights, hasAdminAccess, views token-owner and two-factor suites, and `applyTokenScope` is still applied on the `/o/actions` path.

Also dropped the CHANGELOG entry from this branch — see the separate note.

Cookiezaurs (approved):

PR #7965, opened 2026-08-19
[security][core] escape values at dashboard html sinks the api escaper does not cover (24.05)

ar2rsawseen:
Backport of #7954 to `release.24.05`.

Not a clean cherry-pick: one of the three sites needs a different fix on this branch, because the localized strings differ. Details below.

## Same as master

**compliance-hub export history actions column.** The datatable builds an HTML string in `onReady` and the template renders it. Most of what goes into that string came through `common.returnOutput`, so `escape_html_entities` already escaped it and it is deliberately left alone: escaping twice would surface the entities literally in the UI. One value comes from `countlyGlobal` instead, which `express-expose` serializes for the JavaScript string context only, so the API's HTML escaping never applied to it. It is now escaped where it is interpolated.

`countlyCommon.encodeHtml` and `returnOutput`'s use of `escape_html_entities(k, v, true)` are both identical on this branch, so the reasoning transfers unchanged.

**populator template delete confirmation.** Body converted to text interpolation. `populator.delete-template-description` carries no markup in any of the 26 locale files on this branch.

## Different on this branch

**populator environment delete confirmation.** On master the body converts to text interpolation too. Here it cannot: the string is

```
populator.environment-delete-warning-description = Are you sure you want to delete <b>{0}</b> environment?
```

so converting it would render `<b>` literally. The body keeps `v-html` and the environment name is escaped where it is substituted instead. Same outcome, different mechanism, chosen by what the locale file actually contains rather than by what master did.

The name reaching both dialogs was HTML-decoded on the way in — deliberately, since the dropdown label needs the decoded form — so the API's escaping no longer held at render time. That is what made these reachable.

**plugins dependency confirmation** stays on `v-html`, as on master. `plugins.confirm` carries a `<br/><br/>` in all translated locale files, and the values interpolated into it are plugin titles from package metadata, not anything a dashboard account can write.

## Verification

Ran in a dedicated worktree of `release.24.05`.

- new tests against unpatched code on this branch: **3 failing, 2 passing**. Against patched: **5 passing**. The three failures are exactly the three security assertions.
- full unit suite: **101 passing before, 106 after** (+5, the new tests). Baseline established by re-running with the new file excluded.
- the same **5 pre-existing failures** appear before and after: 3 `Common API utility functions` ObjectID validation cases and 2 network-dependent `Countly Request` cases. Worth flagging that the 3 ObjectID ones are most likely an artifact of running this branch against a newer checkout's `node_modules` rather than anything on the branch itself, since master's suite does not show them. Either way they are identical either side of the change.
- eslint clean on all changed files.

### DIFF
--- a/plugins/compliance-hub/frontend/public/javascripts/countly.models.js
+++ b/plugins/compliance-hub/frontend/public/javascripts/countly.models.js
@@ -194,7 +194,13 @@
                 var ret = "<p>" + ((jQuery.i18n.map["systemlogs.action." + row.a]) ? jQuery.i18n.map["systemlogs.action." + row.a] : row.a) + "</p>";
                 if (typeof row.i === "object") {
                     if (typeof row.i.app_id !== "undefined" && countlyGlobal.apps[row.i.app_id]) {
-                        ret += "<p title='" + row.i.app_id + "'>" + jQuery.i18n.map["systemlogs.for-app"] + ": " + countlyGlobal.apps[row.i.app_id].name + "</p>";
+                        //this string is rendered with v-html, so every interpolated value must already be
+                        //HTML-safe. Everything taken from "row" arrives through common.returnOutput, which
+                        //escape_html_entities has already escaped, so it must NOT be escaped again here or
+                        //the entities would show up literally. The app name is the exception: it comes from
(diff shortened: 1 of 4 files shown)


ar2rsawseen:

PR #7966, opened 2026-08-19
security(core): HTML-encode the application name in the graph note tooltip

ar2rsawseen:
The graph-note hover tooltip builds its content as an HTML string that includes the application name from `countlyGlobal` (raw at runtime) and renders it with `tipsy({html: true})`. Encode the application name with `countlyCommon.encodeHtml` so it renders as text. The other values in the tooltip are API-encoded or i18n; display is unchanged.

Same shape as the active-app name and compliance-hub encodes already in flight.


ar2rsawseen:

PR #7967, opened 2026-08-19
security(core): HTML-encode the application name in the graph note tooltip (24.05)

ar2rsawseen:
Backport of #7966 to `release.24.05`.

The graph-note tooltip builds an HTML string including the application name from `countlyGlobal` (raw at runtime) and renders it with `tipsy({html: true})`. Encode it with `countlyCommon.encodeHtml` so it renders as text; display unchanged.


ar2rsawseen:

PR #7968, opened 2026-08-19
security(push): sanitize message-editor content before rendering into the editor

ar2rsawseen:
The push message editor set the composed message as `innerHTML` on a live `contenteditable` element. This routes it through `countlyCommon.encodeSomeHtml` (the same `filterXSS` sanitizer the note widget uses), with an allowlist restricted to the user-property token `<span>` and the attributes it depends on (`class`, `id`, `contenteditable`, `data-user-property-*`). Any other markup is escaped to inert text.

The message body is user text and the token span is the only legitimate markup, so display is unchanged for normal messages. The token `id` is preserved, so the editor's per-token event wiring (`addEventListeners` → `querySelector("#id-N")`) keeps working. Verified the allowlist deterministically: token attributes and text survive; other tags are escaped.

Note: this touches the message-composer UI, so a quick manual check of composing/editing a notification with a personalization token is worthwhile before merge (it could not be exercised in an automated frontend test here).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Enterprise Fixes:
 - [data-manager] Fixed editing an event whose key contains `&` creating undeletable duplicate rows in the events table
 
 Security Fixes:
+- [push] The message editor now sanitizes message content before rendering it into the editor, allowing only the user-property token element and rendering any other markup as text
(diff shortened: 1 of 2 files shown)


ar2rsawseen:

PR #7969, opened 2026-08-19
security(push): sanitize message-editor content before rendering into the editor

ar2rsawseen:
Backport of #7968 to `release.24.05`.

The push message editor set the composed message as `innerHTML` on a live `contenteditable` element. This routes it through `countlyCommon.encodeSomeHtml` with an allowlist restricted to the user-property token `<span>` and the attributes it depends on (`class`, `id`, `contenteditable`, `data-user-property-*`); any other markup is escaped to inert text.

Display is unchanged for normal messages, and the token `id` is preserved so the editor's per-token event wiring keeps working. Same change as master, applied on identical surrounding code.

Note: a quick manual check of composing/editing a notification with a personalization token is worthwhile before merge.


ar2rsawseen:

PR #7970, opened 2026-08-19
security: encode and sanitize values at frontend HTML sinks

ar2rsawseen:
Consolidates the frontend HTML-sink hardening work into one PR. Each change routes an attacker-influenceable value through the right encoder/sanitizer at the point it is placed into an HTML sink, leaving normal display unchanged. Replaces #7966, #7968, #7961, #7954 and #7949.

- **core / graph note tooltip** — the tooltip HTML-encodes the application name before rendering it (`countly.common.js`).
- **push / message editor** — the composed message is sanitized before it is set as `innerHTML` on the editor's contenteditable, allowing only the user-property token span and its attributes (`countly.views.component.common.js`).
- **compliance-hub / export history** — the export/purge history action cell HTML-encodes the application name; values that already come through the API escaper are left untouched, with a comment and a unit test (`test/unit-tests/plugins.compliance-hub.actions-escaping.js`) guarding both directions.
- **populator / confirm dialogs** — dialog bodies render as text instead of HTML (`environment_detail.html`, `populator.html`).
- **core / res.expose script island** — `<` is escaped when serializing the exposed `countlyGlobal` object into the inline page script, and the active-app name is rendered with `.text()` instead of `.html()` (`express-expose.js`, `countly.template.js`).

Verified: `node --check` and eslint on all changed JS; the compliance-hub escaping unit test passes (5/5).


plugins/push/frontend/public/javascripts/countly.views.component.common.js
```diff
@@ -1,6 +1,15 @@
 /* eslint-disable no-console */
 /*global CV,countlyVue,countlyPushNotification,countlyGlobal,countlyCommon,moment*/
 (function(countlyPushNotificationComponent) {
+    // The message editor is a live contenteditable. Its body is user-authored text; the
+    // only legitimate markup is the user-property token <span>. Allow just that element
+    // (with the attributes the token relies on) and let everything else be escaped to inert
+    // text, so a stored message cannot introduce active markup when the editor is populated.
+    var PUSH_MESSAGE_EDITOR_XSS_OPTIONS = {
+        whiteList: {
+            span: ["class", "id", "contenteditable", "data-user-property-label", "data-user-property-value", "data-user-property-fallback"]
```
ar2rsawseen: [P2] Preserve the token's type attribute

The same component creates each personalization span with `data-user-property-type` and updates it when the user selects a property type, but this allowlist omits that attribute. Every `reset()` therefore strips the type from stored token markup; the next editor change emits `innerHTML` without it and can persist the lossy representation. Include `data-user-property-type` in the allowed span attributes (or remove the attribute from the token contract everywhere if it is truly obsolete).

ar2rsawseen: Confirmed and added in 06a5eda23fe.

The component sets it in two places — on creation (`addEmptyUserProperty`) and on every type change (`setUserPropertyValue`) — and the allowlist named the other three data attributes but not this one, so each sanitize pass dropped it and the following `innerHTML` emit persisted the lossier markup.

Checked the rest of the token contract while I was there: `data-user-property-label`, `-value` and `-fallback`, plus `id`, `class` and `contenteditable`, were all already allowed. Nothing else was missing.

**On the test:** rather than restating the list, it derives the expectation from the component. It reads the allowlist out of the source and the attribute names out of the `setAttribute` calls — in the component *and* in `countly.models.js`, whose `getUserPropertyElement` rebuilds tokens when a saved message is opened and goes through the same sanitizer — then requires every one of them to survive. So the next attribute added to a token has to be allowlisted or the suite says so, which matters because a silently stripped attribute fails quietly. It pins the other direction too: no `on*` handler, nothing that can carry a URL, no element but the span. It fails against the previous state with `data-user-property-type is set on a token but sanitized away`.

**One thing worth knowing, on your "or remove it everywhere if it is truly obsolete".** I checked: it is written and never read. Those two `setAttribute` calls are the *only* occurrences of `data-user-property-type` in countly-server, countly-platform and countly-enterprise-plugins — no `getAttribute`, no CSS selector, no template. The property type is carried in the model (`userPropertyDto[key].t`), and `getUserPropertyElement` does not set the attribute at all, so a message reopened from storage already lacks it.

I still took the first option. Allowlisting a `data-` attribute costs nothing and cannot execute, whereas retiring an attribute from the token contract is a frontend decision that does not belong in an XSS-hardening PR. Flagging it here so it can be retired deliberately if you want it gone.

Cookiezaurs (approved):

